PR #11, opened 2019-03-16
Add a multi-signature smart contract

pmer:
The smart contract uses BLS signatures for its threshold cryptography, so to run properly, it must be inside a chain configured for bls0chain client signatures.

An associated test program can be run with:

```
./docker.local/bin/test.multisigsc.sh
```

This test program tests the basic features and then runs a stress test, performing multi-sig transactions in an infinite loop. It requires that some miners and sharders are running and uses the `docker.local/config/discover_pool.yml` file to discover them.


Sriep:
Closing as over two years old.

PR #59, opened 2021-02-05
Fix round_test.go error

peterlimg:
Change the second parameter of SedRandomSeed and SetRandomSeedForNotarizedBlock from *node.Pool to int,
what it needs is the pool's size.

Initialize the logging on round_test.go so that the calling of Logger would not panic.

### DIFF
--- a/code/go/0chain.net/chaincore/round/entity.go
+++ b/code/go/0chain.net/chaincore/round/entity.go
@@ -206,17 +206,17 @@ func (r *Round) GetRoundNumber() int64 {
 }
 
 //SetRandomSeed - set the random seed of the round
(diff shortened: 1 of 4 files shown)


code/go/0chain.net/chaincore/round/entity.go
```diff
@@ -462,11 +462,15 @@ func SetupRoundSummaryDB() {
 }
 
 /*ComputeMinerRanks - Compute random order of n elements given the random seed of the round */
-func (r *Round) computeMinerRanks(miners *node.Pool) {
-	Logger.Info("waiting to compute miner ranks", zap.Any("num_miners", miners.Size()), zap.Any("round", r.Number))
+func (r *Round) computeMinerRanks(minersNum int) {
```
kirillt, line 465: nice

PR #60, opened 2021-02-07
Fix unit tests in miner_test.go

peterlimg:
## Fixes:
- Fix unit tests in miner_test.go

## Changes:
- Introduce a third-party redis server mocker, as in miner_test.go, the block generation process has to connect a redis server to get transactions. We can either run a mock redis server or do those tests in integration tests, where a real redis server is running. I would chose to run a redis server mocker to minimize the changes.
- Add go mod replace of sharder package in miner's package, as the miner_test.go uses blcokstore, which belongs to sharder.
- Remove the TestBlockVerificationTooFewTransactions cases as insufficient transactions would not be verified in the `VerifyRoundBlock` method, also insufficient transactions would not be returned as an error, and a new block will be created anyway when block generation period is time out.


code/go/0chain.net/miner/miner_test.go
```diff
@@ -147,37 +169,40 @@ func TestBlockVerificationBadHash(t *testing.T) {
...
-		_, err = mc.VerifyRoundBlock(ctx, mr, b)
-	}
-	if err == nil {
-		t.Error("FAIL: Block with too few transactions passed verification")
-	} else {
-		t.Log("SUCCESS: Block with too few transactions failed verifcation")
-	}
-	common.Done()
-}
+//func TestBlockVerificationTooFewTransactions(t *testing.T) {
```
kirillt, line 244: Can we skip this test case instead of commenting it out?
If yes, this is preferred since the code is statically checked and would degrade slower.
Commented out code easily becomes invalid after some time.

code/go/0chain.net/miner/miner_test.go
```diff
@@ -209,29 +235,103 @@ func BenchmarkGenerateAndVerifyALotTransactions(b *testing.B) {
...
+		Dial: func() (redis.Conn, error) {
+			c, err := redis.Dial("tcp", s.Addr())
+			if err != nil {
+				panic(err.Error())
+			}
+			return c, err
+		},
+	})
+
+	//n1 := &node.Node{Type: node.NodeTypeMiner, Host: "", Port: 7071, Status: node.NodeStatusActive}
```
kirillt: Looks like redundant commented out code, if we don't need it better to just remove it for the sake of clarity.

code/go/0chain.net/core/memorystore/connection.go
```diff
@@ -96,8 +96,24 @@ func (s *idStats) CheckExpiredIDs() {
 var pools = make(map[string]*dbpool)
 var idS *idStats
 
-func init() {
-	DefaultPool = NewPool(os.Getenv("REDIS_HOST"), 6379)
+//func init() {
```
kirillt, line 99: Looks like forgotten code, I'll remove it since it is the same as `InitDefaultPool`.

PR #61, opened 2021-02-11
miner unit tests

TheTeaParty:
## Changes
- single go.mod for project to avoid replaces
- unit test coverage for miner
- decouple code using interfaces and DI


peterlimg:
Please update the Dockerfiles in docker.local and docker.aws dirs too, thanks a lot.

kirillt (changes requested):

PR #71, opened 2021-02-18
Feature/sharder code improving

MurashovVen:


code/go/0chain.net/miner/miner_test.go
```diff
@@ -114,6 +119,20 @@ func CreateMockRound(number int64) *MockRound {
 	return mr
 }
 
+func makeTestMinioClient() (blockstore.MinioClient, error) {
+	mConf := blockstore.MinioConfiguration{
+		StorageServiceURL: "play.min.io",
```
kirillt, line 125: We should invent local service since this depends on external configuration, but for now that's fine

PR #72, opened 2021-02-24
[WIP] Add unit tests to sharder.

MurashovVen:


kirillt:
This is good PR. The only reason why I haven't approved it yet, because we need to develop procedure how to work with mocks in our team. Code generation is good but we should avoid committing it since it becomes more difficult to evaluate changes in PRs + we need to update generated mocks when we update code we test. @TheTeaParty uses `mockery` in his PR to generate separate files with mocks — if we adopt this tool, we can add it into `unit_test.sh` script and put generated files into `.gitignore`. This way generated code will be always fresh and invisible for Git.

platsko:
@MurashovVen
1. Please resolve the conflicts.
2. Make sure that the PR changes follow the mock-flow on project.

Sriep:
Replaced with https://github.com/0chain/0chain/pull/236

PR #82, opened 2021-02-26
Speed up docker build

mallochine:
Fork 'docker.local/build.miner.dev/' from 'docker.local/build.miner/'
Edit Dockerfile to do a copy of the miner binary instead of go build.

Benefits: speedup of 36x on the build error path (36 seconds to 1 second). On the build success path, speedup of 7x from 36 sec to 5 sec. 

How to test: build container with `docker.local/bin/build.miners.dev.sh`

Note: building above container also builds the miner binary locally on the filesystem.

# Repro

Steps to repro slow build:
1. Edit "code/go/0chain.net/chaincore/threshold/bls/dkg.go", comment out a field
   in `type DKG struct`
2. Run this command: `time ./docker.local/bin/build.miners.sh`
3. Get this output, showing build time of 36 seconds.

```
The command '/bin/sh -c go build -v -tags "bn256 development" -ldflags "-X 0chain.net/core/build.BuildTag=$GIT_COMMIT"' returned a non-zero code: 2

real    0m36.269s
```

Meanwhile, if I run this command: `cd $TOP/code/go/0chain.net/miner/miner && go build -v -tags "bn256 development"`, it fails in just a second:
```
0chain.net/chaincore/threshold/bls
../chaincore/threshold/bls/dkg.go:68:3: unknown field 'N' in struct literal of type DKG
../chaincore/threshold/bls/dkg.go:90:3: unknown field 'N' in struct literal of type DKG

real    0m0.949s
```

So, this is a real problem.


docker.local/build.miner.dev/Dockerfile.integration_tests
```diff
@@ -0,0 +1,37 @@
+# Compile the miner in an un-tagged image so the final, tagged image can be smaller:
```
kirillt, line 1: Files `Dockerfile.integration_tests`, `b0docker-compose.yml`, `docker-compose.yml`, `nongenesis_docker_compose.yml` are copies of files under another paths, without any changes. Copy-paste is bad practice because when you will need to fix something, you will need to fix it in several places and this is easy to forget.

docker.local/build.miner.dev/Dockerfile
```diff
@@ -0,0 +1,32 @@
...
+
+# Build libzstd:
+# FIXME: Change this after https://github.com/valyala/gozstd/issues/6 is fixed.
+# FIXME: Also, is there a way we can move this to zchain_build_base?
+RUN cd $GOPATH/pkg/mod/github.com/valyala/gozstd* && \
+    chmod -R +w . && \
+    make clean libzstd.a
+
+# Add the source code:
+COPY ./code/go/0chain.net $SRC_DIR/go/0chain.net
```
kirillt, line 26: Instead of copying the whole file and changing small piece of script, it is better to avoid copy-paste and do conditional logic triggered on CLI parameter: https://www.dev-diaries.com/social-posts/conditional-logic-in-dockerfile/

docker.local/bin/build.miners.dev.sh
```diff
@@ -0,0 +1,19 @@
+#!/bin/sh
+set -e
+
+GIT_COMMIT=$(git rev-list -1 HEAD)
+echo $GIT_COMMIT
+
+# Build it locally
+TOP="$PWD"
```
kirillt, line 8: Better this: `TOP=$(git rev-parse --show-toplevel)`, this way you can run your script from any place in the project

docker.local/bin/build.miners.dev.sh
```diff
@@ -0,0 +1,19 @@
...
+cd $TOP/code/go/0chain.net/miner/miner && go build -v -tags "bn256 development" -ldflags "-X 0chain.net/core/build.BuildTag=$GIT_COMMIT"
+cd $TOP
```
kirillt, line 10: Just `cd -`

kirillt (changes requested):

docker.local/bin/build.miners.dev.sh
kirillt: Or `dirname $0` and then relative path from the folder with script, but git way is better IMHO.

docker.local/build.miner.dev/Dockerfile
mallochine: that is conditional RUN. Show me how to do conditional COPY. Already tried googling for 5 min

mallochine: here, best I can get it is I've removed every file except for "Dockerfile.dev", which I put alongside "build.miner/Dockerfile".

So that's only one extra file.

I don't think you're going to easily find a conditional COPY

kirillt (commented):
I agree, conditional COPY in Dockerfile appeared to be not straightforward.
Please, look at this workaround: https://www.taran.space/dev/0chain/dockerfile-workaround.zip

Please, squash commits in the end and we are good.

docker.local/build.miner.dev/Dockerfile
kirillt: I agree, conditional COPY in Dockerfile appeared to be not straightforward.
Please, look at this workaround: https://www.taran.space/dev/0chain/dockerfile-workaround.zip

code/go/0chain.net/chaincore/threshold/bls/dkg.go
```diff
@@ -18,7 +18,7 @@ import (
 /*DKG - to manage DKG process */
 type DKG struct {
 	T  int
-	N  int
+	// N  int
```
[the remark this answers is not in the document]
kirillt: This is the reason, why are you getting this error:
`../chaincore/threshold/bls/dkg.go:68:3: unknown field 'N' in struct literal of type DKG`

There is no such error on `master` after `go build -v -tags "bn256 development"`
from folder `code/go/0chain.net/miner/miner`

kirillt (approved):

Issue #86, opened 2021-03-02
Non converging VRF due to different round timeouts


Sriep:
1. Start 4 miners, 1 sharder.
2. Bring miner2 down. BC should keep moving.
3. 15 seconds later bring miner3 down. BC stops.
4. Bring miner2 up (not miner3). BC gets stuck with no VRF changes with round timeouts. Generally a new round timeout should have new generators assigned with new VRF because you're using the counter increment to generate a new VRF.

Sriep:
On Jun 13, 2020, at 8:32 AM, Saswata Basu <saswata@0chain.net> wrote:
Hey Siva,
I think the round timeout still has an issue in the case where half the miners are on one round, and the other half on another. I think we need to revise the algorithm to have the miners stay in the last or most popular timeout round and not increment it by one, otherwise its a race condition which will never get resolved. Thoughts?
thanks,
Saswata
```go
If (number of vrf tickets > threshold) {
  If (vrfs all same) 
      execute current code
  else sort popular vrf timeouts {
  if popularity rank is same
    select last timeout, add one to sync 
  else 
    select most popular timeout, add one to sync
  }
}
```

Sriep:
Saswata Basu [6:54 PM] Max  [2:12 AM]
Here the file main.go, to demo converging timeout by using my method. 100 nodes start with random timeout [0; 100] and randome delay in the range of 3 sec. This is emulation of restartRound, because for each node it happen for itself timing.
And here Ken’s And Siva’s solution (ken.go). In that case ony 5 nodes converging from 100.
That mean that math method does not work. I suggest that use it in blockchain and build in is useless. And it does not matter how many rounds run. Because the problem with timeout appear only during the restartRound. Moreover, Kens method already has been checked and you have written that there is a problem in converging VRF share.
[Ken's method](https://0chain.slack.com/files/U012NSXAYH3/F018554LMLG/ken.go?origin_team=T9C1E9JEB&origin_channel=D01NYCVANLC)
[Max's main.go](https://0chain.slack.com/files/U012NSXAYH3/F017MF9TD6Y/main.go)
:01
There two implentations that we had regarding the timeout logic - ken’s code and Max’s code (current). I’m not sure if Ken’s code had a problem but I think Max’s code doesnt work either.

Sriep:
[...]

PR #91, opened 2021-03-05
Revert timeout counting argument

Sriep:
The reverts back to a previous algorithm of timeout counting.
It also provides a cap on the size of the timeout counter timeoutCounter.count created by func (tc *timeoutCounter) IncrementTimeoutCount.

It addresses issue https://github.com/0chain/0chain/issues/86 and is an alternative to PR https://github.com/0chain/0chain/pull/90

<details>
  <summary>Model of timeout argument</summary>
The key function is `Incr()`  

  ```go
package main
​
import (
	crand "crypto/rand"
	"fmt"
	"log"
	"math/rand"
	"os"
	"os/signal"
	"sync"
	"time"
)
​
var list []*Node
​
type Node struct {
	id string
​
	sync.Mutex
	count int
	skip  int
	voted map[string]struct{}
	votes map[int]int
}
​
func newNode(i int) (n *Node) {
	n = new(Node)
	n.id = fmt.Sprintf("%0d", i)
	n.resetVotes()
	return
}
​
func (n *Node) resetVotes() {
	n.voted = make(map[string]struct{})
	n.votes = make(map[int]int)
}
​
func (n *Node) isVoted(id string) (ok bool) {
	_, ok = n.voted[id]
	return
}
​
func (n *Node) AddVote(id string, vote int) {
	n.Lock()
	defer n.Unlock()
​
	if n.isVoted(id) {
		return
	}
​
	n.voted[id] = struct{}{}
	n.votes[vote]++
	return
}
​
func (n *Node) Incr() {
	n.Lock()
	defer n.Unlock()
​
	var mostVotes, mostTimeout int = 0, n.count
	for k, v := range n.votes {
		if v > mostVotes || (v == mostVotes && k > mostTimeout) {
			mostVotes = v
			mostTimeout = k
		}
	}
​
	n.resetVotes() // for next voting
​
	if mostTimeout > n.count {
		n.count = mostTimeout + 1 // increase by an external vote
		return
	}
​
	n.count++ // increment own
}
​
func (n *Node) Set(c int) {
	n.Lock()
	defer n.Unlock()
​
	n.count = c
}
​
func (n *Node) Get() int {
	n.Lock()
	defer n.Unlock()
​
	return n.count
}
​
func (n *Node) process() {
	var rnd = getRand()
	n.count = rnd.Intn(100) // start from random in [0; 100) range
	n.skip = n.count        // the same
​
	var delay = time.Millisecond * time.Duration(rnd.Intn(3000))
​
	fmt.Println("=>", n.id, "starts from", n.count, "with delay", delay.String())
​
	time.Sleep(delay) // start with a delay
	for {
		select {
		case <-time.After(3 * time.Second):
			// increment and redo VRFS
			n.Incr()
			for _, x := range list {
				if x.id == n.id {
					continue // skip self
				}
				x.AddVote(n.id, n.count)
			}
			// filter nodes with not the same count
			var got int
			for _, x := range list {
				if x.id == n.id {
					continue // skip self
				}
				if x.Get() != n.Get() {
					continue
				}
				got++
			}
			fmt.Printf("-> %s: got %d/%d (%d)\n", n.id, got+1, len(list), n.Get())
		}
​
	}
}
​
func waitSigInt() os.Signal {
	var sigc = make(chan os.Signal, 2)
	signal.Notify(sigc, os.Interrupt)
	return <-sigc
}
​
func getRand() *rand.Rand {
	var big, err = crand.Prime(crand.Reader, 64)
	if err != nil {
		log.Fatal(err)
	}
	// the seed is OK, but the math/rand gives
	// bad randomization regardless the seed
	var seed = big.Int64()
	return rand.New(rand.NewSource(seed))
}
​
func main() {
​
	const N = 100
	for i := 0; i < N; i++ {
		var nn = newNode(i)
		go nn.process()
		list = append(list, nn)
	}
​
	fmt.Println("got signal", waitSigInt(), "exiting...")
}
  ```

  </details>

### DIFF
--- a/code/go/0chain.net/chaincore/round/entity.go
+++ b/code/go/0chain.net/chaincore/round/entity.go
@@ -3,6 +3,7 @@ package round
 import (
 	"context"
 	"fmt"
+	"github.com/spf13/viper"
 	"math/rand"
 	"os"
 	"runtime/pprof"
@@ -90,7 +91,8 @@ func (tc *timeoutCounter) IncrementTimeoutCount(prrs int64, miners *node.Pool) {
 }
 
 func (tc *timeoutCounter) checkCap() {
-	if tc.count > timeoutCap {
+	timeoutCap := viper.GetInt("server_chain.round_timeouts.timeout_cap")
+	if timeoutCap > 0 && tc.count > timeoutCap {
 		tc.count = timeoutCap
 	}
 }
--- a/code/go/0chain.net/chaincore/round/entity.go
+++ b/code/go/0chain.net/chaincore/round/entity.go
@@ -35,101 +35,61 @@ const timeoutCap = 10 // todo: add to 0chainl.yaml later
 
 type timeoutCounter struct {
 	mutex sync.RWMutex // async safe
+	count int          // current round timeout
 
-	prrs int64    // previous round random seed
-	perm []string // miners of this (not previous) round sorted by the seed
-
-	count int // current round timeout
-
-	votes map[string]int // voted miner_id -> timeout
-}
-
-// The rankTimeoutCounters computes ranks of miners to choose timeout counter.
-// Should be called under lock.
-func (tc *timeoutCounter) rankTimeoutCounters(prrs int64, miners *node.Pool) {
-
-	var nodes = miners.CopyNodes()
-
-	sort.Slice(nodes, func(i, j int) bool {
-		return nodes[i].ID < nodes[j].ID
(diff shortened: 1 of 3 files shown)


kirillt (approved):

kirillt:
Please, squash commits before merging

kirillt:
Should we close this PR since we have merged alternative one?

PR #93, opened 2021-03-06
[WIP] Challenge should time out if its not received from blobber

### LINKED ISSUE #81: Challenges should fail after some time if they have not been processed

shravanshetty1:
closes #81

### DIFF
--- a/code/go/0chain.net/smartcontract/storagesc/challenge.go
+++ b/code/go/0chain.net/smartcontract/storagesc/challenge.go
@@ -216,7 +216,7 @@ func (ssc *StorageSmartContract) saveStakePools(validators []datastore.Key,
 }
 
 // move tokens from challenge pool back to write pool
-func (sc *StorageSmartContract) blobberPenalty(t *transaction.Transaction,
+func (sc *StorageSmartContract) blobberPenalty(creationDate common.Timestamp,
 	alloc *StorageAllocation, prev common.Timestamp, bc *BlobberChallenge,
 	details *BlobberAllocation, validators []string,
 	balances c_state.StateContextI) (err error) {
@@ -302,7 +302,7 @@ func (sc *StorageSmartContract) blobberPenalty(t *transaction.Transaction,
 		// make sure all mints not payed yet be payed before the stake
 		// pools will be slashed
 		var info *stakePoolUpdateInfo
-		info, err = sp.update(conf, sc.ID, t.CreationDate, balances)
+		info, err = sp.update(conf, sc.ID, creationDate, balances)
 		if err != nil {
 			return fmt.Errorf("updating stake pool: %v", err)
 		}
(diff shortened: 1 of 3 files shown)


code/go/0chain.net/smartcontract/storagesc/challenge.go
```diff
@@ -340,9 +350,90 @@ func (sc *StorageSmartContract) blobberPenalty(t *transaction.Transaction,
...
+
+			cct := toSeconds(details.Terms.ChallengeCompletionTime)
+			fresh := challenge.Created+cct >= t.CreationDate
+
+			if !fresh {
+				// time of previous complete challenge (not the current one)
+				// or allocation start time if no challenges
+				var prev = alloc.StartTime
+				if last := blobberChallengeObj.LatestCompletedChallenge; last != nil {
+					prev = last.Created
```
shravanshetty1, line 393: @guruhubb I am not sure we can handle it like this, seems like processing of challenges need to happen in a specific order - to properly calculate the reward/penalty amount. I cant find any documentation on this except for comments. This seems to be an architecture issue. Since these challenges would have been roughly created at the same time the penalty will be very low for timeouts, not sure this is intended.

guruhubb: Have you looked at https://github.com/0chain/0chain/tree/master/code/go/0chain.net/smartcontract/storagesc regarding payments. The blobber gets paid if they pass challenges, otherwise, after paying the validator, it goes back to client along with slash of their stake. When the chain generates a challenge, the SC assigns random validators. The blobbers periodically check to see if there are challenges and then they send their response to the assigned validators.

andrenerd (approved):
Ok.
Was asked to review the PR. Frankly the updated logic is far beyond my current understanding of the project.  So just read the code.

shravanshetty1:
Test doesnt work - will have to read documentation so i can create proper tests. Can continue on this issue - but i think it may take a long time for me to write tests, due to lack of knowledge

code/go/0chain.net/smartcontract/storagesc/challenge.go
guruhubb: @Sriep can you look into this PR since you're very familiar with challenge rewards, etc

Sriep:
Unclear whether filtering timed out challenges should be done in the blobber or 0chain. Payments are currently handled in [finalizeAllocation](https://github.com/0chain/0chain/blob/master/code/go/0chain.net/smartcontract/storagesc/allocation.go#L1253) so it is not necessary for them to be moved here.

I'll leave this PR open for the moment, but it might become redundant by the time issue https://github.com/0chain/0chain/issues/81 is closed.

Sriep:
Closing. Replacing with 0chain issue https://github.com/0chain/0chain/issues/198 and blobber issue https://github.com/0chain/blobber/issues/93. 
Fixing existing code, `finalizeAllocation` and `cacnelAllocationRequest` rather than writing new code.

PR #116, opened 2021-03-16
state_context.go fix for smartcontgract/storatec

andriykutsevol:
This fixes  an incompatibility between the GetKey functions in the file:
https://github.com/0chain/0chain/blob/master/code/go/0chain.net/smartcontract/storagesc/models.go

and the IsHash function:
https://github.com/0chain/0chain/blob/378cc81404ca6e4dfab81eab2a432d174aaaf074/code/go/0chain.net/core/encryption/hash.go#L22

IsHash wants HASH_LENGTH = 32
But GetKey functions add an arbitrary number of bytes to the HASH_LENGTH.

```
func (sn *StorageAllocation) GetKey(globalKey string) datastore.Key {
	return datastore.Key(globalKey + sn.ID)
}
```

Regression tests were made:
create allocation
upload file to allocation.
./zbox listallocations

### DIFF
--- a/code/go/0chain.net/chaincore/chain/state/state_context.go
+++ b/code/go/0chain.net/chaincore/chain/state/state_context.go
@@ -227,25 +227,34 @@ func (sc *StateContext) GetSignatureScheme() encryption.SignatureScheme {
 }
 
 func (sc *StateContext) GetTrieNode(key datastore.Key) (util.Serializable, error) {
-	if encryption.IsHash(key) {
+
+	key_hash := encryption.Hash(key)
+
+	if !encryption.IsHash(key_hash) {
 		return nil, common.NewError("failed to get trie node", "key is too short")
 	}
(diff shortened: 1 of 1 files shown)


code/go/0chain.net/chaincore/chain/state/state_context.go
```diff
@@ -227,25 +227,34 @@ func (sc *StateContext) GetSignatureScheme() encryption.SignatureScheme {
 }
 
 func (sc *StateContext) GetTrieNode(key datastore.Key) (util.Serializable, error) {
-	if encryption.IsHash(key) {
-		return nil, common.NewError("failed to get trie node", "key is too short")
+
+	key_hash := encryption.Hash(key)
+
+	if !encryption.IsHash(key_hash) {
```
Sriep, line 233: Can this test ever fail? You have just created `key_hash`, so isn't it a hash? Same for insert and delete.

PR #127, opened 2021-03-26
Merging general and SC unit tests scripts

kirillt:
We have separate tests script for `smartcontract` module which wasn't called at CI.
This PR merges this script into "general" `unit_test.sh` and also contains some refactoring.

New **CI output** (now calls `smartcontract` submodules):
https://www.0chain-ci.net/root/0chain/-/jobs/3191/raw

Now, for **local usage** instead of this
```
./docker.local/bin/sc_unit_test.sh <package>
```
you need to run this:
```
./docker.local/bin/unit_test.sh --sc <package>
```


docker.local/bin/unit_test.sh
```diff
@@ -4,27 +4,70 @@ set -e
...
 PACKAGE=""
+IMAGE=""
 
-if [[ "$1" == "--ci" ]]
+
+GENERAL_IMAGE="zchain_unit_test"
+SC_IMAGE="zchain_sc_unit_test"
+
+if [[ "$@" == *"--ci"* ]]
 then
```
Sriep, line 14: I don think `[[` is valid in sh, use `[`.

PR #117, opened 2021-03-17
Fixing rewards and fees + fees_test, lots of refactoring and other small fixes

kirillt:
Files `fees.go` and `fees_test.go` reworked.
Interests and their test case in `fees_test.go` wasn't touched and need to be overhauled later.

`MinerNode` renamed to `ConsensusNode` for the sake of clarity (since before, sharders were also `MinerNode`).
Small fixes like configuration messages/logs typos, etc.

PR #136, opened 2021-03-31
Minersc payFes: fix sharder delegate reward and view change interest

Sriep:
Fix a couple of issues in fees.go.

* The service change for sharder delegates was wrong. The service charge was not being removed. This was also fixed in PR https://github.com/0chain/0chain/pull/117 but that came with extensive refactoring. This aims to change as little as possible.
* Interest payments not being payed when `0chian.view_change = true`. Now interest payments are paid every multiple of `sc.minersc.reward_round_frequency`, this is the same irrespective of the `view_change setting`.


Sriep:
Replaced with https://github.com/0chain/0chain/pull/142

PR #112, opened 2021-03-14
Feature/unit test interestpoolsc

gasparyanyur:

PR #144, opened 2021-04-03
unit test for interstpoolsc

Sriep:
Tests `InterestPoolSmartContract.lock` and `InterestPoolSmartContract.unlock`.

These are the smart contract methods that support the `zwallet lock` and `zwallet unlock cli commands.

There is a spelling mistake in [sc.go](https://github.com/0chain/0chain/blob/master/code/go/0chain.net/smartcontract/interestpoolsc/sc.go#L60) `insufficent`-> `insufficient`


gasparyanyur:
there is a PR related to interstpoolsc.
https://github.com/0chain/0chain/pull/112

So we need to merge PR112 first and there is already implemented mock service for balances!

Sriep:
There is another mock service for balances but that is overly complicated for this test which only needs a few of the `state.StateContextI` methods implemented.

Sriep:
replaced with https://github.com/0chain/0chain/pull/171

Issue #149, opened 2021-04-05
build.miners.sh: --dev flag leads to standard_init_linux.go:219: exec user process caused "no such file or directory"


bbist:
While running the image build using build.miners.sh with --dev flag enabled, on ubuntu 20.10 container crashes with following error.
```
standard_init_linux.go:219: exec user process caused "no such file or directory"
```
The build script seems to build the image on local system and copy it to alpine based image.

This issue could be addressed by moving the go build step inside docker build stage.

mallochine:
in codereview

mallochine:
merged.

Issue #126, opened 2021-03-25
This doesn't actually copy from dev environment `./docker.local/bin/build.miners.sh --dev`


mallochine:
@kirillt Sorry not sure what happened but doesn't look like this is working.

Steps to repro:
1. git reset --hard 6412185c3ca1981367aaa82ca8dd722da19fc3e0
2. ./docker.local/bin/build.miners.sh --dev
3. The docker script starts compiling the golang code instead of copying from the host filesystem.

You know the script starts building when you see this. What's happening here is the script is compiling the code in the Docker container instead of copying it from the dev environment.

As an additional check, `Artifacts will be copied from local machine` doesn't get echoed to terminal

```
Step 17/21 : RUN if [ "$DEV" = "yes" ] ;    then echo "Artifacts will be copied from local machine";    else go build -v -tags "bn256 development" -ldflags "-X 0chain.net/core/build.BuildTag=$GIT_COMMIT" ;fi
 ---> Running in 03a588d22e9f
golang.org/x/sys/unix
github.com/fsnotify/fsnotify
github.com/hashicorp/hcl/hcl/strconv
github.com/hashicorp/hcl/hcl/token
github.com/hashicorp/hcl/hcl/ast
github.com/hashicorp/hcl/hcl/scanner
github.com/hashicorp/hcl/hcl/parser
github.com/hashicorp/hcl/json/token
github.com/hashicorp/hcl/json/scanner
github.com/hashicorp/hcl/json/parser
github.com/hashicorp/hcl
...
```

kirillt:
`6412185` is "Using the same script for both dev and prod build" commit which isn't last commit of that PR (#82)
https://github.com/0chain/0chain/pull/82/commits

`git checkout feat/faster-docker-miner-build` is better.
Also, in order to test properly we should take Docker caches into account.

kirillt:
Indeed, we are lacking `ARG DEV` before usage of `DEV` in `Dockerfile.template`

mallochine:
```
[kirill@lenovo docker]$ cat Dockerfile 
FROM alpine
ARG DEV
RUN echo "$DEV"
RUN if [ "$DEV" = "yes" ] ;\
    then echo "yes" ;\
    else echo "no" ;\
fi
[kirill@lenovo docker]$ docker image prune -f && docker build --build-arg DEV=no .
52091d08b082ad6e3127b0c4fc20fa0b1fd2b4a54b2494ac861f114b76878ed1
[...]

PR #150, opened 2021-04-05
Miner sharder dev build flag #149

bbist:
This change references following issue: https://github.com/0chain/0chain/issues/149

As specified in above issue, dev build logic for miner has been moved inside container. This is done by vendoring dependencies and building inside the container using -vendor flag.

Note that, the .dockerignore file had **/pkg entry. One of the vendored dependencies (minio) needed the pkg directory that included credentials. Hence, **/pkg entry needed to be removed from .dockerignore. Ignoring such directory was probably done because of some necessity. Therefore, it needs some review.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -2,4 +2,3 @@
 awsnet
 docker.local/miner*
 docker.local/sharder*
-**/pkg
--- a/docker.local/bin/build.miners.sh
+++ b/docker.local/bin/build.miners.sh
@@ -9,32 +9,37 @@ DOCKERDIR="$ROOT/docker.local/build.miner"
 DOCKERFILE="$DOCKERDIR/Dockerfile"
 DOCKERCOMPOSE="$DOCKERDIR/docker-compose.yml"
 
+APP_DIR="$ROOT/code/go/0chain.net/miner/miner"
+
 if [[ "$@" == *"--dev"* ]]
 then
-    echo -e "\nDevelopment mode: building miner locally\n"
+    cd $APP_DIR
+    echo "Building: --dev mode: vendoring dependencies"
+    rm -rf vendor
+    go mod vendor
+    # libzstd: start: to rebuild inside container
+    dstdir="vendor/github.com/valyala"
+    rm -r $dstdir/*
+    srcdir="$GOPATH/pkg/mod/github.com/valyala"
+    cp -r "$srcdir/$(ls $srcdir | tail -n1)" $dstdir/gozstd
+    chmod -R +w vendor
+    # libzstd: end
+    cd $ROOT
+fi
 
-    cd "$ROOT/code/go/0chain.net/miner/miner"
-    go build -v -tags "bn256 development" \
-        -ldflags "-X 0chain.net/core/build.BuildTag=$GIT_COMMIT"
+docker build --build-arg GIT_COMMIT=$GIT_COMMIT -f $DOCKERFILE -t miner .
 
-    sed 's,%COPY%,COPY ./code,g' "$DOCKERFILE.template" > "$DOCKERFILE"
-
-    cd "$ROOT"
-    docker build --build-arg GIT_COMMIT=$GIT_COMMIT \
-        -f "$DOCKERFILE" . -t miner --build-arg DEV=yes
-else
-    echo -e "\nProduction mode: building miner in Docker\n"
-
-    sed 's,%COPY%,COPY --from=miner_build $APP_DIR,g' "$DOCKERFILE.template" > "$DOCKERFILE"
-
-    cd "$ROOT"
-    docker build --build-arg GIT_COMMIT=$GIT_COMMIT \
-        -f "$DOCKERFILE" . -t miner --build-arg DEV=no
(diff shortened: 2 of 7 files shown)


mallochine:
no wait, the point of the --dev is to build it on the local system.

mallochine:
if I purposely introduce a compile error in golang, how many seconds does it take for building with --dev to fail?

What are the steps you took to test that --dev work?

bbist:
> no wait, the point of the --dev is to build it on the local system.
> 
> What are the full steps you take to repro #149

Happened a couple of times. Didn't put a lot of effort into reproducing, as I was primarily working on something else. Made above mentioned changes and got it working.

mallochine:
So, wait, you actually use the --dev flag? How many do you think uses --dev flag?

bbist:
> if I purposely introduce a compile error in golang, how many seconds does it take for building with --dev to fail?
> 
> What are the steps you took to test that --dev work?

I assumed the reason for --dev build was to make it build faster. The build fails in case there's any compile error in go code after the valyala/gozstd compilation step inside container. It's not immediate, but it's also not that slow. And as the build actually takes place inside the container based on alpine, it seemed like a proper approach.
Let me know if I missed something.

bbist:
> So, wait, you actually use the --dev flag? How many do you think uses --dev flag?

Yes, I do. Until I have whole dev environment setup on my local system without running sharders/miners/blobbers/0dns/etc.. on docker, I'll likely use it.
I'm not sure if someone else is using it.

mallochine:
Yes the point of --dev is to speed up build.

Normally, a bad build fails in under a second with --dev

how many seconds does it take to fail with your stuff here?

mallochine:
What are the steps to test your stuff after I checkout your branch? 1, 2, 3 sorta instructions

bbist:
> What are the steps to test your stuff after I checkout your branch? 1, 2, 3 sorta instructions

The branch only includes that specific change. You can run following command to build miner image.
```
docker.local/bin/build.miners.sh --dev
```

You can also take a look into the modified scripts and dockerfiles if you like.
Because of the valyala/gozstd it takes about 20 seconds for the go build to start. But that's kind of the step used to avoid the panic specified in the title.
Anyway, I hoped you'd review before deciding whether to accept the changes.

mallochine:
anyway how do you know the `miner` is getting built on your local system? I don't think it is.

See #126 where the script never actually ends up copying from system.

Wanna fix the bug?

You need this:

```
[kirill@lenovo docker]$ cat Dockerfile 
FROM alpine
ARG DEV
RUN echo "$DEV"
RUN if [ "$DEV" = "yes" ] ;\
    then echo "yes" ;\
    else echo "no" ;\
fi
[kirill@lenovo docker]$ docker image prune -f && docker build --build-arg DEV=no .
52091d08b082ad6e3127b0c4fc20fa0b1fd2b4a54b2494ac861f114b76878ed1
STEP 1: FROM alpine
STEP 2: ARG DEV
--> Using cache 367354a9d1e67788bfe14bc6414382ffb851b0b2e21d135c8fbf6600cdb091fa
STEP 3: RUN echo "$DEV"
--> Using cache c67915c230bf3d9685fe105b1759769034dc246ebe66be8fd42922f5a4bfa096
STEP 4: RUN if [ "$DEV" = "yes" ] ;    then echo "yes" ;    else echo "no" ;fi
no
STEP 5: COMMIT
--> f9c34ec80fa
f9c34ec80fa0c13bc057446197e3f26bf816fbed8bd36b8f41f2310dca228f07
[kirill@lenovo docker]$ docker image prune -f && docker build --build-arg DEV=yes .
f9c34ec80fa0c13bc057446197e3f26bf816fbed8bd36b8f41f2310dca228f07
STEP 1: FROM alpine
STEP 2: ARG DEV
--> Using cache 367354a9d1e67788bfe14bc6414382ffb851b0b2e21d135c8fbf6600cdb091fa
STEP 3: RUN echo "$DEV"
yes
--> 9013e6cbff8
STEP 4: RUN if [ "$DEV" = "yes" ] ;    then echo "yes" ;    else echo "no" ;fi
yes
STEP 5: COMMIT
--> 6e377a1b52d
6e377a1b52dfeab65dbf5ab0239b2fac09200fea2f3ff3cf7c208af8e87715c5
```

bbist:
Wasn't aware of older issues. In that case, you can ignore the pull request for now. It's not that big of a deal for me anyway. If someone else faces similar issue, they'll have this thread to refer to.

mallochine:
If you have an easy way to save 20 seconds every time you build, that's something to consider

mallochine:
hey @bbist  I think I fixed this in master with my commit 6ed936135f8d6b752486901665f5adbd5dea75c0, could you test whether that works for you, and if it does, close out this pull request?

bbist:
> hey @bbist I think I fixed this in master with my commit [6ed9361](https://github.com/0chain/0chain/commit/6ed936135f8d6b752486901665f5adbd5dea75c0), could you test whether that works for you, and if it does, close out this pull request?

Thanks for the help. This can be closed.

PR #153, opened 2021-04-06
Revert "[WIP] Fix BC stuck on VC:ture"

Sriep:
Reverts 0chain/0chain#152

This Pr failed the unit test check and should not have been merged. As a result syntax errors have entered the master branch and everyone else's Pr checks are failing. 

The problem appears to be adding a new parameter to `MerklePatriciaTrie.SaveChanges` without updating the unit tests with the new parameter.

To avoid this happening again I would advise only allowing merging PRs that pass all checks.


peterlimg:
Don't revert PR #152, I'll fix the unit tests, was busy in fixing the VC:true issue.

PR #158, opened 2021-04-07
Handle miracl pubkey

mallochine:
There is only one commit in this PR, it's mine.

How it was tested:
* @anish-squareops  tested it on five network. Zero issue
* I wrote a small unit test sanity checking it works.

Why it's safe:
* it's 10 lines of code....takes a string and returns a string. No side effects. Totally safe.


Sriep:
This pull request needs to be resubmitted with only the changes relevant to this fix.

PR #177, opened 2021-04-12
Implement persistencestore interface types: SessionI, QueryI, BatchI, IteratorI

MurashovVen:


Sriep:
The [unit test checks](https://github.com/0chain/0chain/pull/177/checks?check_run_id=2325512068) are failing. I tested master https://github.com/0chain/0chain/pull/181 and there does not seem a problem.

If you don't think the unit test failure is due to this PR, could you force the cheks to rerun; you do this by making a dummy commit. I can squash it out later.

I can't approve a PR unless the unit-test checks pass.

Issue #180, opened 2021-04-13
Blober challange,  validators not getting there service charge


Sriep:
[blobberReward in chaallenge.go](https://github.com/0chain/0chain/blob/master/code/go/0chain.net/smartcontract/storagesc/challenge.go#L76) is not paying the service charge to validators. [moveToValidator](https://github.com/0chain/0chain/blob/master/code/go/0chain.net/smartcontract/storagesc/challengepool.go#L192) only pays out to delegates.

Sriep:
Fixed in https://github.com/0chain/0chain/pull/182

PR #182, opened 2021-04-13
Fix Validators not getting service charge for blobber challenge

Sriep:
Fixes Issue https://github.com/0chain/0chain/issues/180. Required for PRs https://github.com/0chain/0chain/pull/186 and https://github.com/0chain/0chain/pull/122

Merges the service charge calculations for blobber and validators. `moveToBlobber` and `moveToValidator` merged into `moveReward`. The only difference between the two, apart from the bug being fixed, is whether to increment `sp.Rewards.Blobber` or `sp.Rewards.Validator`, This is moved this outside the function.

Interest was being paid only for the case blobber delegates and challenge failing, which was a little odd. This removes all interest payments during the challenge. Also, the interest calculation was refactored to reduce rounding errors.

This change broke a couple of unit tests, `allocation_test.go` and `blobber_test.go`, which relied on using no validator delegates. However, validator delegates are compulsory and the new code gives an error if they are missing. I did the minimum changes to get the tests working, there is a separate PR https://github.com/0chain/0chain/pull/122 which is updating `blobber_test.go`, but is waiting for this PR so that it can merge in changes.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,5 +1,6 @@
 # TestNet Setup with Docker Containers
 
+
 ## Table of Contents
 
 - [Initial Setup](#initial-setup) 
--- a/README.md
+++ b/README.md
@@ -1,6 +1,5 @@
 # TestNet Setup with Docker Containers
 
-
 ## Table of Contents
 
 - [Initial Setup](#initial-setup) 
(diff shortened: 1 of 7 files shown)


bbist (approved):
All changes look good. Unit tests need to be revisited once these changes are merged.

Issue #130, opened 2021-03-27
Failing unit tests.


Sriep:
The following unit tests fail and need fixing. They were deactivated in PR https://github.com/0chain/0chain/pull/124 so that the master branch would pass unit test checks.
```
code/go/0chain.net/chaincore/chain/protocol_round_test.go
code/go/0chain.net/chaincore/client/client_test.go
code/go/0chain.net/chaincore/wallet/wallet_test.go
code/go/0chain.net/core/persistencestore/persistencestore_test.go
code/go/0chain.net/core/persistencestore/persistencestore_test.go
code/go/0chain.net/miner/miner_test.go
code/go/0chain.net/sharder/blockstore/blockdb_store_test.go
code/go/0chain.net/smartcontract/minersc/fees_test.go
```

The following unit tests use `fmt.print` to display output. This is bad style, slows down scripts and makes it difficult to follow the screen and help with [unit test checks](https://github.com/0chain/0chain/pull/124) . These print statements should probably be removed; should not take long. The worst offenders like `merkle_patricia_tire_test.go` and `merkle_patricia_tire2_test.go` should certainly be cleaned up, as they dump numerous MPT databases contents to output.
```
pool_test.go 
config_test.go 
blobber_test.go 
blockdb_test.go 
dkg_test.go 
round_test.go 
node_test.go   
mpt_nodedb_test.go 
mpt_nodedb_test.go     
merkle_patricia_tire_test.go     
merkle_patricia_tire2_test.go
ed25519_test.go 
encryption_test.go 
ed25519_test.go  
bls0chain_test.go 
bls0chain_splittable_signature_test.go  
bls0chain_aggregate_test.go 
codex_test.go 
wallet_test.go
```

Sriep:
Three ways to [prevent race conditions.](https://www.sohamkamani.com/golang/data-races/)
1. Use mutexes to prevent multiple access to objects. Risk of deadlock conditions.
2. Use channels to force all access down a single thread.
3. Use waitgroups to force goroutines to run one at a time.

Sriep:
`home/piers/GolandProjects/0chain/code/go/0chain.net/sharder/blockstore/fs_store_test.go` sometimes fails with concurrent writes, indicating race issues. Run with test with -race to see data race issues.

Sriep:
code/go/0chain.net/chaincore/chain/protocol_round_test.go fixed in PR https://github.com/0chain/0chain/pull/185

Sriep:
```
[...]

PR #185, opened 2021-04-14
Fix unit test: synchronise tests with adding offset

MurashovVen:


Sriep:
This test appers linked to issue 130 https://github.com/0chain/0chain/issues/130

PR #187, opened 2021-04-15
refactored by moving locks to appropriate level and fixed race condit…

bbist:
…ions in MemoryNodeDB methods: Iterate, Validate, ComputeRoot

As few exported methods in MemoryNodeDB didn't have locks while accessing map, it may lead to concurrent modification panics. Added appropriate locks at appropriate levels.

### DIFF
--- a/code/go/0chain.net/core/util/mpt_nodedb.go
+++ b/code/go/0chain.net/core/util/mpt_nodedb.go
@@ -137,7 +137,6 @@ func (mndb *MemoryNodeDB) DeleteNode(key Key) (err error) {
 func (mndb *MemoryNodeDB) MultiGetNode(keys []Key) (nodes []Node, err error) {
 	mndb.mutex.RLock()
 	defer mndb.mutex.RUnlock()
-	nodes = make([]Node, 0, len(keys))
 	for _, key := range keys {
 		node, nerr := mndb.getNode(key)
 		if nerr != nil {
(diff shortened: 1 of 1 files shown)


Sriep:
Please ensure unit test check pass.  Your changes seem to have broken TestMemoryNodeDB_Full.

bbist:
> Please ensure unit test check passes. Your changes seem to have broken TestMemoryNodeDB_Full.

I've fixed the portions causing tests to fail. Now all 0chain.net/core/util unit tests pass.

peterlimg (changes requested):
Please avoid using named return values when possible, it should only be used when necessary, such as when using a defer to catch an error on function exit, or there are lot of return values that would confuse people.

Sriep (approved):
I second peterlimg's dislike of named return values, unfortunately, they were used liberally when 0chain was first written; which sometimes creates a conundrum, but not here.

bbist:
> Please avoid using named return values when possible, it should only be used when necessary, such as when using a defer to catch an error on function exit, or there are lot of return values that would confuse people.

Minimized use of named return values.
https://github.com/0chain/0chain/commit/2b8fd6c5f6a282df710cf1437ae7ac5ef877ad93

code/go/0chain.net/core/util/mpt_nodedb.go
```diff
@@ -470,33 +519,33 @@ func (lndb *LevelNodeDB) Size(ctx context.Context) int64 {
 
 // PruneBelowVersion - implement interface.
 func (lndb *LevelNodeDB) PruneBelowVersion(ctx context.Context, version Sequence) error {
-	// TODO
+	// TODO: Why not implemented?
 	return nil
```
Sriep, line 523: I assume it is not implemented because it's not used anywhere. 

However, I don't like it silently failing; Either delete it or add in a panic. Unless there is a clear intention to use it later my preference is to delete it.

Issue #190, opened 2021-04-16
View Change process gets stuck at Share phase and restarts whenever one of the MPK miners goes down (or intentionally rejects share) during Share phase


bbist:
In the DKGProcess() method in 0chain.net/miner/protocol_view_change.go, 
a new phase event is allowed to execute on following conditions:
1. It's a `share` phase
2. It's a `start` phase
3. It's one phase after current phase

But this logic breaks the view-change process because in the current implementation when a `share` phase function keeps returning error, the current phase doesn't move. Re-execution of `share` phase is okay because periodic phase event would force miner to re-share it's shares to other MPK miners. But when the phase completion time (in terms of rounds) expires, the MinerSC smart contract moves to `publish` phase, and a `publish` phase event is received by all DKG miners. But the miners that were stuck in re-executing `share` phase because they were not able to get verification from at least one of the DKG miners and weren't able to move phase from `contribute` to `share`, would not be able to proceed because of the condition 3 mentioned above. The current phase is at `contribute`, but the phase event received is `publish` (`contribute` + 2). This shows up as "dkg process -- jumping over a phase" debug log in 0chain.log.

Solution for this would be to allow `share` phase to move current phase from `contribute` to `share`, even if the `share` phase function fails to send shares to and receive verification from few of the MPK miners. And allow the `share` phase to be retried every "repeat" interval (until next phase is started), so that it might succeed this time if there were some network issues before. This will ensure the current phase has moved from `contribute` to `share` during `share` phase even if few MPK miners failed to verify shares (for whatever reason). And the new `publish` phase event would execute as expected in condition 3.

Terminologies:
DKG Miners: All miners selected for DKG process
MPK Miners: All DKG miners who submitted their MPK to the BC

peterlimg:
> But when the phase completion time (in terms of rounds) expires, the MinerSC smart contract moves to publish phase
May I know where's the code that related to this, thank you.

bbist:
The logs that helped identify this were from https://two.devnet-0chain.net/miner*. Following sequence of logs seems to repeat continuously.
```
2021-04-14T16:37:36.285Z        INFO    miner/protocol_view_change.go:160       dkg process start       {"next_phase": {"phase":0,"start_round":158701,"cu
rrent_round":158702,"restarts":726}, "phase": "unknown", "sc funcs": 5}
[...]

PR #191, opened 2021-04-17
allow share phase to partially fail and still move to publish phase #190

bbist:
Hi @peterlimg and @Kenwes13 ,

Can you please review these changes regarding https://github.com/0chain/0chain/issues/190?


code/go/0chain.net/smartcontract/minersc/dkg.go
```diff
@@ -703,6 +696,13 @@ func (msc *MinerSmartContract) shareSignsOrShares(t *transaction.Transaction,
 			"invalid state: decoding miners MPK: %v", err)
 	}
 
+	// We may use != instead of < here? @kenwes13
+	if len(sos.ShareOrSigns) < len(mpks.Mpks)-1 {
```
peterlimg, line 700: Use `if len(sos.ShareOrSigns) < dmn.K-1`.

peterlimg (changes requested):

code/go/0chain.net/smartcontract/minersc/dkg.go
peterlimg: Never mind, I'll update it in fix/vc-true-stuck branch. Let's see how would this go with fix/vc-true-struck branch.

Issue #189, opened 2021-04-16
Avoid lock copy in 0chain.net/chaincore/node.SetInfo/GetInfo


bbist:
There's lock copy in SetInfo/GetInfo that's been there for a while.
Using a reference to the Info object and calling it's Copy() method wherever necessary will allow us to avoid it.
```

Line 637:
func (n *Node) SetInfo(info *Info) {
	n.mutex.Lock()
	defer n.mutex.Unlock()
	n.Info = info.Copy()
}

// GetInfo returns copy Info.
func (n *Node) GetInfo() Info {
	n.mutex.RLock()
	defer n.mutex.RUnlock()
	return n.Info.Copy()
}

Line 153:
			info := Info{}
			if err := common.FromJSON(resp.Body, &info); err == nil {
				info.AsOf = time.Now()
				nd.SetInfo(&info)
			}
```

Sriep:
Fixed by 193

PR #193, opened 2021-04-19
use copy method instead of direct assignment to avoid lock copy #189

bbist:
Issue: https://github.com/0chain/0chain/issues/189


bbist:
I'll fix the failing unit tests from other packages and let you know.

bbist:
> I'll fix the failing unit tests from other packages and let you know.

Test is passing locally.
```bash
$ go test -timeout 30s -run "^TestChain_GetBlockBySummary$" 0chain.net/sharder
ok      0chain.net/sharder      0.078s
```

bbist:
Actually, using lock in Info struct doesn't seem to make sense anyway. Maybe we should completely do away with the mutex @peterlimg ?

Sriep (approved):

Issue #176, opened 2021-04-12
unit_test.sh usng [[ in sh


Sriep:
unit_test.sh uses sh `#!/bin/sh` but uses the `[[` construct. When run only showing errors you get
```sh
docker.local/bin/unit_test.sh: 13: [[: not found
docker.local/bin/unit_test.sh: 24: [[: not found
docker.local/bin/unit_test.sh: 38: [[: not found
```
[[ is only valid in bash not sh. Either use `#!/bin/bash` or change `[[` to `[`. 
Also check `start.b0sharder.sh` and `start.b0miner.sh`. 
`build.miner.sh` should be ok as it uses `bash`.

Sriep:
https://github.com/0chain/0chain/pull/194

PR #194, opened 2021-04-19
Fix bash in scripts

Sriep:
Changed shell scripts that make use of `[[ ]]` to use `bash` as they don't work in `sh`. 

Modified scripts as suggested by [SC2199](https://github.com/koalaman/shellcheck/wiki/SC2199) to not use `$@` in `[[`.

Fixed `unit_test.sh`, which was not working as intended.

Resolves issue https://github.com/0chain/0chain/issues/176


bbist (approved):
All changes look good.
One thing to note about docker.local/bin/unit_test.sh is that the expected order of flags and package name has changed. Previously, if a package and --sc flag were provided --sc flag was expected to come after the package name. Now --sc flag (if provided) should come before package name (if provided). I think the new one is better as it brings consistency to the usage of unit_test.sh.
If someone is using unit_test.sh locally, they may want to change the order of their flags and arguments.

Issue #92, opened 2021-03-06
Smart contract return 400 for every possible error


shravanshetty1:
Should return appropriate status code

service-0chain:
Manual system tests [failure] with the following config
| config  | value |
| ------------- | ------------- |
| tests skipped  | false  |
| system-tests  |   |
| gosdk  | NOT OVERRIDDEN  |
| zboxcli  |   |
| zwalletcli  |   |
| 0chain  |   |
| blobber  |   |
| authorizer  |   |
| 0box  |   |
| 0dns  |   |

PR #162, opened 2021-04-07
Single `go.mod` For Project 

TheTeaParty:
Closes #141

PR #196, opened 2021-04-20
Implement valid response status code

MurashovVen:
- Implementation of a common error interface for smart contracts.
- Changing the error handling process and writing the appropriate status code in http response.
- Unit tests.


bbist (commented):
There seems to be a problem with unit test. It's currently using Go 1.12 as base image, but the chaincore package requires 1.13, and the function errors.Is that you've added was only included in errors package in Go 1.13. Therefore, we need to update the build-base image to at least Go 1.13.
The unit tests pass with Go 1.13 in base image. You can replace,
`FROM golang:1.12.4-alpine3.9`
in file 0chain/0chain/docker.local/build.base/Dockerfile.build_base with
`FROM golang:1.13-alpine3.12`
and run the unit-tests locally,
`docker.local/bin/unit_test_build.sh `

I'll review the code as soon as I finish up few of my other tasks.

code/go/0chain.net/chaincore/chain/state_handler.go
```diff
@@ -31,6 +34,46 @@ func SetupStateHandlers() {
...
+			w.WriteHeader(http.StatusInternalServerError)
+		case errors.Is(err, sc.NewErrNoResource()):
+			w.WriteHeader(http.StatusNotFound)
+		default:
+			w.WriteHeader(http.StatusBadRequest)
+		}
+
+		buf := bytes.NewBuffer(nil)
+		json.NewEncoder(buf).Encode(data)
+		buf.WriteTo(w)
```
bbist, line 59: Suggestion:
json.NewEncoder(w).Encode(data)

code/go/0chain.net/chaincore/chain/state_handler.go
```diff
@@ -31,6 +34,46 @@ func SetupStateHandlers() {
...
+		return
+	}
+	if data != nil {
+		w.Header().Set("Content-Type", "application/json")
+		if !strings.Contains(r.Header.Get("Accept-Encoding"), "gzip") {
```
bbist, line 64: Would it be better if compressions are handled by an explicit middleware?

code/go/0chain.net/chaincore/chain/state_handler.go
```diff
@@ -42,7 +85,12 @@ func (c *Chain) HandleSCRest(w http.ResponseWriter, r *http.Request) {
 		scRestRE = regexp.MustCompile(`/v1/screst/(.*)?/(.*)`)
 		pathParams = scRestRE.FindStringSubmatch(r.URL.Path)
 		if len(pathParams) == 3 {
-			common.ToJSONResponse(c.GetSCRestOutput)(w, r)
+			if !common.CheckCrossOrigin(w, r) {
```
bbist, line 88: Would it be better if crossorigin checks are handled by explicit middlewares?

code/go/0chain.net/smartcontract/errors.go
```diff
@@ -0,0 +1,136 @@
...
+	FailRetrievingBlobbersListErr   = errors.New("can't get blobbers list")
+	FailRetrievingBlobberErr        = errors.New("can't get blobber")
+	FailRetrievingReadPoolErr       = errors.New("can't get read pool")
+	FailRetrievingWritePoolErr      = errors.New("can't get write pool")
+	FailRetrievingStakePool         = errors.New("can't get related stake pool")
+	FailRetrievingUserStakePoolErr  = errors.New("can't get user stake pools")
+	FailRetrievingChallengePoolErr  = errors.New("can't get challenge pool")
+	FailRetrievingPoolErr           = errors.New("can't get pool")
+	FailGetOrCreateClientPoolsErr   = errors.New("can't get or create client pools")
+)
```
bbist, line 77: Maybe we can define smartcontract specific errors inside corresponding smartcontract package?

code/go/0chain.net/smartcontract/errors.go
```diff
@@ -0,0 +1,136 @@
+package smartcontract
```
bbist, line 1: Even though the issue title suggested response status specific to smart contracts, this way of API error reporting looks good and is something we could share across different parts of our application. It would make our error reporting more consistent. Just a thought. Maybe we can do that later?

code/go/0chain.net/smartcontract/errors.go
```diff
@@ -0,0 +1,136 @@
...
+var (
+	DecodingErr                 = errors.New("can't decode resource")
+	FailDecodingMinerErr        = errors.New("can't decode miner from passed params")
+	FailDecodingMpksBytesErr    = errors.New("can't decode mpks bytes")
+	FailDecodingGroupErr        = errors.New("can't decode group")
+	FailDecodingMagicBlockErr   = errors.New("can't decode magic block")
+	FailDecodingReadMarker      = errors.New("can't decode read marker")
+	FailDecodingAllocationErr   = errors.New("can't decode allocation")
+	BlobberChallengeDecodingErr = errors.New("fail decoding blobber challenge")
+)
```
bbist, line 112: Should we define smartcontract specific errors inside corresponding smartcontract package?

code/go/0chain.net/smartcontract/faucetsc/handler.go
```diff
@@ -52,7 +55,8 @@ func (fc *FaucetSmartContract) globalPerodicLimit(ctx context.Context, params ur
 func (fc *FaucetSmartContract) pourAmount(ctx context.Context, params url.Values, balances c_state.StateContextI) (interface{}, error) {
 	gn, err := fc.getGlobalNode(balances)
 	if err != nil {
-		return nil, common.NewError("failed to get limits", "global node does not exist")
+		err := smartcontract.NewError(smartcontract.FailRetrievingLimitsErr, "global node does not exist")
```
bbist, line 58: We could use a different error "more related to pourAmount" instead of FailRetrievingLimitsErr.

code/go/0chain.net/smartcontract/minersc/handler.go
```diff
@@ -120,32 +130,37 @@ func (msc *MinerSmartContract) GetGroupShareOrSignsHandler(ctx context.Context,
...
 	groupBytes, err := balances.GetTrieNode(GroupShareOrSignsKey)
 	if err != nil {
-		return nil, err
+		err := smartcontract.NewError(smartcontract.FailRetrievingGroupErr, err)
+		return nil, smartcontract.WrapErrNoResource(err)
 	}
 	err = gsos.Decode(groupBytes.Encode())
 	if err != nil {
-		return "", err
+		err := smartcontract.NewError(smartcontract.FailDecodingGroupErr, err)
```
bbist, line 138: Error could be more granular. Something like "can't decode group shares or signs".

code/go/0chain.net/smartcontract/minersc/handler.go
```diff
@@ -120,32 +130,37 @@ func (msc *MinerSmartContract) GetGroupShareOrSignsHandler(ctx context.Context,
 	gsos := block.NewGroupSharesOrSigns()
 	groupBytes, err := balances.GetTrieNode(GroupShareOrSignsKey)
 	if err != nil {
-		return nil, err
+		err := smartcontract.NewError(smartcontract.FailRetrievingGroupErr, err)
```
bbist, line 133: Error could be more granular. Something like "can't get group shares or signs".

code/go/0chain.net/smartcontract/minersc/sharder.go
```diff
@@ -149,7 +150,7 @@ func (msc *MinerSmartContract) getShardersList(balances cstate.StateContextI,
 	}
 	err = allMinersList.Decode(allMinersBytes.Encode())
 	if err != nil {
-		return nil, err
+		return nil, smartcontract.NewError(smartcontract.DecodingErr, err)
```
bbist, line 153: The decoding error message seems too generic. Maybe we could add more context into the message as it's being used in API.

code/go/0chain.net/smartcontract/storagesc/challengepool.go
```diff
@@ -364,15 +365,24 @@ func (ssc *StorageSmartContract) getChallengePoolStatHandler(
...
 	if alloc, err = ssc.getAllocation(allocationID, balances); err != nil {
-		return
+		intErr := smartcontract.NewError(smartcontract.FailRetrievingAllocationErr, err)
+
+		switch {
+		case errors.Is(err, util.ErrValueNotPresent):
+			return nil, smartcontract.WrapErrNoResource(intErr)
+		default:
+			return nil, smartcontract.WrapErrInternal(intErr)
+		}
```
bbist, line 380: if errors.Is(err, util.ErrValueNotPresent) {
    return nil, smartcontract.WrapErrNoResource(intErr)
}
return nil, smartcontract.WrapErrInternal(intErr)

Just for consistency across different parts of the application.

code/go/0chain.net/smartcontract/storagesc/handler.go
```diff
@@ -21,10 +22,16 @@ func (ssc *StorageSmartContract) GetBlobberHandler(ctx context.Context,
...
 	if blobberID == "" {
-		return nil, errors.New("missing 'blobber_id' URL query parameter")
+		return nil, smartcontract.WrapErrInvalidRequest(errors.New("missing 'blobber_id' URL query parameter"))
 	}
 
-	return ssc.getBlobber(blobberID, balances)
+	bl, err := ssc.getBlobber(blobberID, balances)
+	if err != nil {
+		err := smartcontract.NewError(smartcontract.FailRetrievingBlobberErr, err)
+		return nil, smartcontract.WrapErrNoResource(err)
```
bbist, line 31: ErrNodeNotFound, ErrValueNotPresent could be NoResource errors. Other errors could be reported as internal error?

bbist: If checking this at every error becomes too repetitive, we can use some helper function to do it.

bbist (changes requested):
Most of the changes look good. But there are places I think we could improve.
1. We could make errors more granular to include more context. More generic errors like decoding error could be improved. This can be done for other generic errors.
2. We could use helper functions to reduce code verbosity. Like following pattern seems very repetitive. We could update Wrap method to do this in a single line.
```
err := smartcontract.NewError(smartcontract.FailRetrievingConfigErr, err.Error())
return nil, smartcontract.WrapErrNoResource(err)
```
3. Smart contract specific errors could be defined inside corresponding smartcontract packages.
4. General HTTP error handling looks good, so it could be moved to core package so that we could use similar reporting pattern across other parts of the application. We can later add more HTTP error statuses as necessary and improve on current implementation.

code/go/0chain.net/smartcontract/errors.go
```diff
@@ -0,0 +1,136 @@
...
+
+func WrapErrInternal(err error) error {
+	return fmt.Errorf("%w: %s", NewErrInternal(), err)
+}
+
+var internal = &internalErr{msg: internalErrMsg}
+
+func NewErrInternal() error {
+	return internal
+}
```
vudn95, line 101: Should we define `internalErr` as `*common.Error` and `internalErrMsg` could be present as `Code` format?
> `internalErrMsg` could be `internal_error`

code/go/0chain.net/smartcontract/errors.go
```diff
@@ -0,0 +1,136 @@
...
+
+func WrapErrInvalidRequest(err error) error {
+	return fmt.Errorf("%w: %s", NewErrInvalidRequest(), err)
+}
+
+var invalidRequest = &invalidRequestErr{msg: invalidRequestErrMsg}
+
+func NewErrInvalidRequest() error {
+	return invalidRequest
+}
```
vudn95, line 136: Should we define `invalidRequestErr` as `*common.Error` and `invalidRequestErrMsg` could be present as `Code` format?
> `invalidRequestErrMsg` could be `invalid_request`

vudn95 (commented):
Most of the updates look good to me. But I think we should only use only one format structure of error to handle elsewhere and also responding to the client.
What I expected is all the main errors should implement as `*common.Error` struct.

code/go/0chain.net/chaincore/chain/state_handler.go
MurashovVen: Good suggestion, thank you.
I've made the appropriate changes.

code/go/0chain.net/chaincore/chain/state_handler.go
MurashovVen: I've moved this method from the `state_handler` to the `common` package, as it was before.

code/go/0chain.net/chaincore/chain/state_handler.go
MurashovVen: I've moved this method from the `state_handler `to the `common `package, as it was before.

code/go/0chain.net/smartcontract/errors.go
MurashovVen: Sure, I've replaced it with string constants.

code/go/0chain.net/smartcontract/errors.go
MurashovVen: Sure, I've moved the main errors, corresponds to http statuses, to the `common` package.

code/go/0chain.net/smartcontract/errors.go
MurashovVen: Sure, I've replaced it with string constants.

code/go/0chain.net/smartcontract/faucetsc/handler.go
MurashovVen: Done.

code/go/0chain.net/smartcontract/minersc/handler.go
MurashovVen: Done.

code/go/0chain.net/smartcontract/minersc/handler.go
MurashovVen: Done.

code/go/0chain.net/smartcontract/minersc/sharder.go
MurashovVen: Done.

code/go/0chain.net/smartcontract/storagesc/challengepool.go
MurashovVen: I added the `NewErrNoResourceOrErrInternal` method solving the boilerplate code problem.

code/go/0chain.net/smartcontract/storagesc/handler.go
MurashovVen: I've added `NewErrNoResourceOrErrInternal` method.

code/go/0chain.net/chaincore/chain/state_handler.go
```diff
@@ -31,6 +34,46 @@ func SetupStateHandlers() {
...
+		switch {
+		case errors.Is(err, sc.NewErrInvalidRequest()):
+			w.WriteHeader(http.StatusBadRequest)
+		case errors.Is(err, sc.NewErrInternal()):
+			w.WriteHeader(http.StatusInternalServerError)
+		case errors.Is(err, sc.NewErrNoResource()):
+			w.WriteHeader(http.StatusNotFound)
+		default:
+			w.WriteHeader(http.StatusBadRequest)
+		}
```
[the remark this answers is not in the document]
MurashovVen: Yes, it would be better. I've implemented `Is(target error)` method for the `Error` struct.

code/go/0chain.net/smartcontract/errors.go
```diff
@@ -0,0 +1,136 @@
...
+
+func WrapErrNoResource(err error) error {
+	return fmt.Errorf("%w: %s", NewErrNoResource(), err)
+}
+
+var noResource = &noResourceErr{msg: noResourceErrMsg}
+
+func NewErrNoResource() error {
+	return noResource
+}
```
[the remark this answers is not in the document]
MurashovVen: Yes, it solves by implementing `Is(target error)` method for the `Error` struct.

code/go/0chain.net/smartcontract/errors.go
MurashovVen: Done.

code/go/0chain.net/smartcontract/errors.go
MurashovVen: Done.

MurashovVen:
Unit test validation failed, but on my local machine, all tests pass.
I can make an empty commit to rerun the tests.

Sriep:
There are merge conflicts with PR https://github.com/0chain/0chain/pull/162, this PR converts 0cahin to use a single go.mod. 

Unit tests need to pass. If there is a problem here we need to solve it.

Any review posts that have been resolved, please set to resolved with the `resolve conversation` button. Still, open issues imply there is still a problem.

code/go/0chain.net/smartcontract/errors.go
TheTeaParty: @MurashovVen 
Maybe lets try avoiding `common` and `util` package nameing, as at some point those packages become harder to identify their purpose. Better have domain specific error separation.
https://blog.golang.org/package-names

> Avoid meaningless package names. Packages named util, common, or misc provide clients with no sense of what the package contains. This makes it harder for clients to use the package and makes it harder for maintainers to keep the package focused. Over time, they accumulate dependencies that can make compilation significantly and unnecessarily slower, especially in large programs. And since such package names are generic, they are more likely to collide with other packages imported by client code, forcing clients to invent names to distinguish them.

code/go/0chain.net/smartcontract/faucetsc/handler.go
```diff
@@ -1,24 +1,30 @@
...
 
 	c_state "0chain.net/chaincore/chain/state"
-	"0chain.net/core/common"
+)
+
+const (
+	noLimitsMsg     = "can't get limits"
+	noGlobalNodeMsg = "can't get global node"
+	noClient        = "can't get client"
 )
```
TheTeaParty, line 18: what you think isn't it better to replace those with `errors`. So it will be easier to mentain those errors not only http handlers wide but also system wide like 
```
ErrNoLimitsMsg = errors.New("can't get limits")
```

this make it way easy to handle specific error using standard tools
also we are loosing original cause of error, maybe wrapping will be better

code/go/0chain.net/smartcontract/errors.go
TheTeaParty: @MurashovVen but why have you replaced it with string constants? moving those exact errors into corresponding smartcontract package will look perfect.

code/go/0chain.net/smartcontract/faucetsc/handler.go
Sriep: This Pr does not seem to be creating any new packages, so don't understand why package naming is an issue here. 
Current error handling across `0chain` products is crude, with inline string constants being the norm. Might be an idea to create an issue to discuss error handling.

TheTeaParty: @Sriep 
Sorry maybe I was not cleat, but I mean as @bbist mentioned and I totally agree with this. Ad this relates to any other application domain object, each error need to be defined in corresponding domain object or level of it's usage.

> Should we define smartcontract specific errors inside corresponding smartcontract package?

And for some reason some errors have have been moved to `common` package, we should avoid any new code in common directory.

Sriep:
This Pr fixes https://github.com/0chain/0chain/issues/92 which is all it was intended to do. Move discussion of the perfect error reporting framework to https://github.com/0chain/0chain/issues/235.

PR #212, opened 2021-04-29
Mock Submodule

### LINKED ISSUE #170: Unit test mocks
https://github.com/0chain/0chain/pull/115, https://github.com/0chain/0chain/pull/87, https://github.com/0chain/0chain/pull/72, https://github.com/0chain/0chain/pull/61 and possibly https://github.com/0chain/0chain/pull/112 make use of unit test mocks. 

Currently, the code is mixed in with unit test in the above PRs, difficult to find and undocumented.

This issue is to discuss this potential new feature.

### LINKED ISSUE #210: Mock Submodule
`mocks` are useful for users of your library as well, so we need to create `mocks`, sub package in `mocks` directory, which will contain both generated and hand made mocks.

TheTeaParty:
Closes #210 and Closes #170


platsko:
In my opinion, combining commits in this way is completely inappropriate.
I propose to follow the principle that PR should solve only one issue and the rule to create a PR with one commit (or a group of commits) logical related to an issue. So, please split this PR into at least four separated ones.

TheTeaParty:
> In my opinion, combining commits in this way is completely inappropriate.
> I propose to follow the principle that PR should solve only one issue and the rule to create a PR with one commit (or a group of commits) logical related to an issue. So, please split this PR into at least four separated ones.

This commit, only includes mocks, and nothing else, the only related change for mocks is to remove dot import because, there is an issue with  identifying types when generating mocks + it is not recommended use dot imports by community anyway.

PR #214, opened 2021-04-29
Protect entityMetadataMap from concurrent access #197

### LINKED ISSUE #197: Protect entityMetadataMap from concurrent access
[entityMetadataMap](https://github.com/0chain/0chain/blob/master/code/go/0chain.net/core/datastore/entity_metadata.go#L5) is a global map accessed from numerous files across the project. It should be protected from concurrent access.

Adding a RWMutex to `RegisterEntityMetadata` and `GetEntityMetadata`  should be sufficient.

This causes a race issue when running `go test -race` on the  `/0chain/code/go/0chain.net/sharder/blockstore/` directory.

platsko:
Closes #197


code/go/0chain.net/go.mod
```diff
@@ -1,9 +1,10 @@
 module 0chain.net
 
-go 1.15
+go 1.16
 
 require (
 	github.com/0chain/gorocksdb v0.0.0-20181010114359-8752a9433481
+	github.com/alicebob/miniredis/v2 v2.14.3
 	github.com/asaskevich/govalidator v0.0.0-20190424111038-f61b66f89f4a
```
Sriep, line 8: Why are changes to go.mod needed? If this is not part of the issue being fixed these changes should be removed.

code/go/0chain.net/go.sum
```diff
@@ -4,212 +4,381 @@ cloud.google.com/go v0.38.0/go.mod h1:990N+gfupTy94rShfmMCWGDn0LpTmnzTp2qbd1dvSR
 cloud.google.com/go v0.44.1/go.mod h1:iSa0KzasP4Uvy3f1mN/7PiObzGgflwredwwASm/v6AU=
 cloud.google.com/go v0.44.2/go.mod h1:60680Gw3Yr4ikxnPRS/oxxkBccT6SA1yMk63TGekxKY=
 cloud.google.com/go v0.45.1/go.mod h1:RpBamKRgapWJb87xiFSdk4g1CME7QZg3uwTez+TSTjc=
+cloud.google.com/go v0.46.3 h1:AVXDdKsrtX33oR9fbCMu/+c1o8Ofjq6Ku/MInaLVg5Y=
 cloud.google.com/go v0.46.3/go.mod h1:a6bKKbmY7er1mI7TEI4lsAkts/mkhTSZK8w33B4RAg0=
```
Sriep, line 8: Why are changes to go.sum needed? If this is not part of the issue being fixed these changes should be removed.

Sriep (changes requested):
Recommend removing changes to go.mod and go.sum

code/go/0chain.net/go.mod
platsko: Pls see https://golang.org/ref/mod#go-mod-tidy

platsko: I suppose it will be good practice to execute the `go mod tidy` command before push a commit to the remote repository.

Sriep: I am sure `go mod tidy` is working fine. ​But this is unrelated to your changes, so the `go.mod`changes are due to other files in your local source. Even if your source files are not dirty, there is no way to confirm this.
Try to avoid unrelated refactoring, there are too many ways this can go wrong. 
If you are worried get a clean copy of master down and run `go mod tidy` on that. You can raise an issue and create a Pr if changes are suggested.

PR #87, opened 2021-03-02
[WIP] Adding core unit tests

MurashovVen:
1. Adding unit tests to core module.
2. Improving code: 
- Minor fixes.
- Implementing mockery mock generator.
- Adding Session, Query, Batch and Iterator interfaces in `persistencestore` package.

Issue #200, opened 2021-04-25
Investigate ToMsgpack for race issues.


Sriep:
Someone write an ineffective unit test, [codec_test.go](https://github.com/0chain/0chain/blob/master/code/go/0chain.net/core/common/codec_test.go#L10), to test race issues in [ToMaspack](https://github.com/0chain/0chain/blob/master/code/go/0chain.net/core/common/codec.go#L33). Investigate where they missed any race issues.

ToMsgpack is designed to receive objects that can have an interface that locks and unlocks a mutex. Confirm any objects that might have race issues on using ToMsgpack, have the [ReadLockable interface](https://github.com/0chain/0chain/blob/master/code/go/0chain.net/core/common/readlockable.go).

Remove confounding race issues from `codec_test.go`.

PR #218, opened 2021-04-30
Fixing chaincore and core unit tests

MurashovVen:
Fixed tests:

1) Removing logs from `client_test`
2) Removing logs from `node_test`
3) Removing logs from `round_test` and adding error checking
4) Removing logs from `wallet_test` and minor code refactoring 
5) Removing logs from `dkg_test` and minor code refactoring 
6) Removing logs from `codec_test` 
7) Tests of `util` package (`merkle_patricia_trie_test`, `merkle_patricia_trie2_test`, `merkle_patricia_trie3_test`, `merkle_trie_test`, `mpt_nodedb_test`):
removing logs, code refactoring .
8) Removing old tests from `memorystore` and adding new
9) Removing old tests from `persistencestore` and adding new
10) Test of `encryption` package (`bls0chain_test`, `bls0chain_aggregate_test`, `bls0chain_splittable_test`, `ed25519_test`, `encryption_test`):
removing logs

Also added `mocks` package which includes mocks, used in added tests.


Sriep (commented):
All new unit tests should be checked with the `-race` option to  make sure they are not contributing race issues.

 On running the tests with the `-race` options two tests modified here produced errors.

* code/go/0chain.net/core/common/codec_test.go Preexisting code. Probably can be handled in https://github.com/0chain/0chain/issues/200, as the whole test needs a rethink.
* code/go/0chain.net/core/memorystore/store_test.go This is a new test and should not produce confounding race errors. If the race errors here are due to the unit test, they should be fixed. If they are due to the production code; Well done! raise an issue to fix them.

MurashovVen:
These errors occur because `memorystore.DefaultPool` isn't protected from concurrent access. We can resolve these race errors by adding global mutex that protects `DefaultPool`. 
Also, `trackCollection` method runs  `CollectionTrimmer` as a goroutine that needs access to `DefaultPool`. It makes race test errors and I think we can resolve it by locking mutex when CollectionTrimmer calls `GetEntityConnection` func.

Sriep:
With this, can we close PR https://github.com/0chain/0chain/pull/87 Adding core unit tests?

Sriep (approved):

MurashovVen:
> Added issue #242, to fix memorystore.DefaultPool.
> 
> With this, can we close PR #87 Adding core unit tests?

Only partially. There are also tests that I would like to decompose into packages and make appropriate PRs:
- `cache`
- `common`
- `datastore`
- `ememorystore`
- `encryption`
- `logging`
- `metric`
- `util`

PR #246, opened 2021-05-06
﻿ISSUE#195 github.com/spf13/viper is not thread safe

platsko:
There is a suggestion to solve the problem with the Viper package using wrapper functions. Let's discuss this.


Sriep (changes requested):
Looks promising. We need unit tests for `core/viper`. What about reading in various 0chain config files, writing them out then rereading them?

platsko:
@Sriep
During testing, significant shortcomings in the implementation were revealed, many methods cannot be thread-safe - I think it would be better to get rid of them altogether. Please see the last commit.

Sriep:
Do we use any of the non-thread-safe methods? 

Eye on the goal: We don't want race errors. So either avoid using unsafe viper methods or make sure they are only accessed from one thread.

Sriep:
I think we just have to let the non-thread-safe methods stand. Assume they are always called safely.

If necessary we can then raise a separate issue to check our usage of non-thread-safe viper methods and, if necessary, fix that in a separate PR.

Sriep (approved):

Issue #249, opened 2021-05-07
Race error: Protect blockdb.mapIndex.


Sriep:
[blockdb.mapIndex](https://github.com/0chain/0chain/blob/master/code/go/0chain.net/sharder/blockdb/index.go#L21) generates race errors.

PR https://github.com/0chain/0chain/pull/250 will generate race errors in `code/go/0chain.net/sharder/blockdb/blockdb_test.go` which this issue should fix.

 <details>
    <summary>Race errors from running unit test</summary>

```
=== RUN   TestDBWrite
==================
WARNING: DATA RACE
Write at 0x00c0001b23f0 by goroutine 9:
  runtime.mapassign_faststr()
      /usr/local/go/src/runtime/map_faststr.go:202 +0x0
  0chain.net/sharder/blockdb.(*mapIndex).SetOffset()
      /home/piers/GolandProjects/0chain/code/go/0chain.net/sharder/blockdb/index.go:29 +0x6f
  0chain.net/sharder/blockdb.(*BlockDB).WriteData()
      /home/piers/GolandProjects/0chain/code/go/0chain.net/sharder/blockdb/db.go:150 +0x12c
  0chain.net/sharder/blockdb.TestDBWrite.func1()
      /home/piers/GolandProjects/0chain/code/go/0chain.net/sharder/blockdb/blockdb_test.go:72 +0xb9

Previous write at 0x00c0001b23f0 by goroutine 8:
  runtime.mapassign_faststr()
      /usr/local/go/src/runtime/map_faststr.go:202 +0x0
  0chain.net/sharder/blockdb.(*mapIndex).SetOffset()
      /home/piers/GolandProjects/0chain/code/go/0chain.net/sharder/blockdb/index.go:29 +0x6f
  0chain.net/sharder/blockdb.(*BlockDB).WriteData()
      /home/piers/GolandProjects/0chain/code/go/0chain.net/sharder/blockdb/db.go:150 +0x12c
  0chain.net/sharder/blockdb.TestDBWrite.func1()
      /home/piers/GolandProjects/0chain/code/go/0chain.net/sharder/blockdb/blockdb_test.go:72 +0xb9

Goroutine 9 (running) created at:
  0chain.net/sharder/blockdb.TestDBWrite()
      /home/piers/GolandProjects/0chain/code/go/0chain.net/sharder/blockdb/blockdb_test.go:70 +0x64f
  testing.tRunner()
      /usr/local/go/src/testing/testing.go:1194 +0x202

[...]

PR #250, opened 2021-05-07
Expose race errors in blockIndex.mapIndex

Sriep:
Modify blockdb_test.go to WriteData in separate goroutines. This generates race errors from `blockdb.mapIndex`.

These errors should be fixed in issue https://github.com/0chain/0chain/issues/249.


code/go/0chain.net/sharder/blockdb/blockdb_test.go
```diff
@@ -63,12 +63,20 @@ func TestDBWrite(t *testing.T) {
...
 	students[2] = &Student{Name: "Apache - the first open source web server", ID: "1995"}
+
+	var wg sync.WaitGroup
 	for _, s := range students {
-		err = db.WriteData(s)
-		if err != nil {
-			panic(err)
-		}
+		wg.Add(1)
+		go func(s *Student, wg *sync.WaitGroup) {
```
platsko, line 70: `wg` won't race, so you don't need to pass it as an argument

platsko (approved):

PR #262, opened 2021-05-11
Feature: add datastore and ememorystore unit tests

MurashovVen:
- Adding datastore unit tests.
- Adding ememorystore unit tests.
- Adding mocks generated by `mockery`:
datastore/Store;
datastore/Entity;
datastore/EntityMetadata;
- Minor code fixes:
fixing errors handling;


code/go/0chain.net/core/ememorystore/connection.go
```diff
@@ -4,9 +4,10 @@ import (
 	"context"
 	"fmt"
 
+	"github.com/0chain/gorocksdb"
+
 	"0chain.net/core/common"
 	"0chain.net/core/datastore"
-	"github.com/0chain/gorocksdb"
```
Sriep, line 9: Does not seem like a meaningful change. Suggest reverting this.

Sriep (changes requested):

code/go/0chain.net/core/ememorystore/connection.go
MurashovVen: I've sorted imports because it increases readability.
The first group for core go libs, the second for external imports, and the third for internal imports.
I can revert this, but I think sorting is good.

platsko: I completely agree with Benjamin - strongly recommended sort imports automatically by `goimports`

Sriep: Doing unrelated refactoring in files you are not working on is not recommended. 

No problem with `goimports`, I recommend using it systematically, maybe raise an issue about it.

Sriep: Doing unrelated refactoring in files you are not working on is not to be recommended. 

No problem with `goimports`. Maybe we can add it to CI.

Sriep (approved):

PR #272, opened 2021-05-12
check for duplicate blobbers

andrenerd:
Prevent "double" registration for blobbers, checking Id/Url.


Sriep:
Can we have a unit test to prove the changes? Also add a description, even if it just links to the attached issues.

Sriep (changes requested):
Can we have a unit test to prove the changes? Also add a description, even if it just links to the attached issues.

Sriep (approved):

PR #282, opened 2021-05-15
Add mutex to mapIndex

Sriep:
Issue https://github.com/0chain/0chain/issues/249.

Also removed confounding race error from code/go/0chain.net/sharder/blockdb/blockdb_test.go.


bbist (commented):
Looks good.

PR #286, opened 2021-05-17
Fix vc true stuck partially

peterlimg:
DO NOT `squash and merge`
Changes:

- Return a clone block of latest finalized magic block when it will be accessed concurrently
- Update to send entity requests concurrently instead of one by one. When there are lots of miners and shaders, sending request one by one would always time out.
- on `getMinerNode` function, check the node from state db first, if failed, check from `allMinersList`
- Update dkg move functions to return error instead of bool to see errors easier.
- Fix nodes status monitor issues.
- Fix `unknown batch error`
- Fix panic errors.


peterlimg:
DO NOT `Squash and merge`

bbist:
@peterlimg,
Could you resolve the unit test failures?

PR #294, opened 2021-05-19
WIP Features/clientpool

andrenerd:


andrenerd:
It's not fully completed yet.
Just a prototype to discuss and integrate with other features, like freeContracts (@bbist) and etc.

code/go/0chain.net/smartcontract/storagesc/readpool.go
```diff
@@ -140,14 +140,6 @@ type readPoolRedeem struct {
 	Balance state.Balance `json:"balance"` // balance reduction
 }
 
-func toJson(val interface{}) string {
```
bbist, line 143: Why was toJson method removed?

bbist (commented):
Looks fine.
Could you please add unit tests for clientpool.go? There are a lot of methods with "calculation" dependent logic. It'd be much easier to work with later on if we have full test coverage since the beginning.

code/go/0chain.net/smartcontract/storagesc/readpool.go
andrenerd: it's a minor utility function (not a  struct method to be precise), that was used poorly. Particularly, it was called from the "sub" packages that are candidates for removal (read/writepool). Also it has no "intrinsic" value, from my point of view.

Sriep:
clientpool.go needs a unit test. Also, complete all `todo`s and remove commented out code.

code/go/0chain.net/smartcontract/storagesc/balances.go
```diff
@@ -0,0 +1,37 @@
+// Experimental
+// todo: naming? balance(s)?
```
Sriep, line 2: Yes its a silly name, but everyone does it.

code/go/0chain.net/smartcontract/storagesc/clientpool.go
```diff
@@ -0,0 +1,722 @@
...
+// todo: add description, rename it
+func (cPool *clientPool) blobberCut(aPoolID, bPoolID string, eTimestamp common.Timestamp,
+) []*allocationPool {
+	return cPool.Allocations.blobberCut(aPoolID, bPoolID, eTimestamp)
+}
+
+// Former removeEmpty
+// todo: add description
+// todo: clarify and rename - to remove allocation or allocations?
+func (cPool *clientPool) removeEmptyAllocation(aPoolID string, aPools []*allocationPool) {
```
Sriep, line 63: `remove empty allocation` seems self explanatry.

code/go/0chain.net/smartcontract/storagesc/clientpool.go
```diff
@@ -0,0 +1,722 @@
...
+	cPool.Allocations.removeEmpty(aPoolID, aPools)
+}
+
+// For read operations only
+// todo: better rename it, adding "Read" in
+// todo: something probably wrong with it.
+// todo: clientPool entity should be used insted of some params to be passed?
+func (cPool *clientPool) moveBlobberCharge(ssContractKey string, sPool *stakePool,
```
Sriep, line 71: Might be best to use [function.go.](https://github.com/0chain/0chain/blob/master/code/go/0chain.net/smartcontract/storagesc/functions.go#L36) athough this is being redone in PR https://github.com/0chain/0chain/pull/317 where it has will be changed to [transferReward](https://github.com/0chain/0chain/blob/966d7479999ac15ce1642e029154e73c95405ba3/code/go/0chain.net/smartcontract/storagesc/functions.go#L16)

code/go/0chain.net/smartcontract/storagesc/clientpool.go
```diff
@@ -0,0 +1,722 @@
...
+	}
+
+	// blobber service charge
+	sPool.Rewards.Charge += value
+	return
+}
+
+// For read operations only
+// todo: better rename it, adding "Read" in
+func (cPool *clientPool) movePartToBlobber(ssContractKey string, aPool *allocationPool,
```
Sriep, line 98: Might be best to use [function.go.](https://github.com/0chain/0chain/blob/master/code/go/0chain.net/smartcontract/storagesc/functions.go#L36) athough this is being redone in PR https://github.com/0chain/0chain/pull/317 where it has will be changed to [transferReward](https://github.com/0chain/0chain/blob/966d7479999ac15ce1642e029154e73c95405ba3/code/go/0chain.net/smartcontract/storagesc/functions.go#L16)

code/go/0chain.net/smartcontract/storagesc/clientpool.go
```diff
@@ -0,0 +1,722 @@
...
+
+// Save pool in tree
+func (cPool *clientPool) save(ssContractKey, clientID string, balances cstate.StateContextI,
+) (err error) {
+	_, err = balances.InsertTrieNode(clientPoolKey(ssContractKey, clientID), cPool)
+	return
+}
+
+// todo: add description, rename it
+func (cPool *clientPool) blobberCut(aPoolID, bPoolID string, eTimestamp common.Timestamp,
```
Sriep, line 55: Its just a wrapper around a blobberCut method so no reason to name it anything else, unless you rename the whole stack.

code/go/0chain.net/smartcontract/storagesc/clientpool.go
```diff
@@ -0,0 +1,722 @@
...
+		dp.Rewards += move         // add to stake_pool_holder rewards
+		sPool.Rewards.Blobber += move // add to total blobber rewards
+	}
+
+	return
+}
+
+// For write operations only
+// todo: better rename it, adding "Write" in
+func (cPool *clientPool) moveToChallenge(allocID, blobID string,
```
Sriep, line 145: This is moving tokens to the clinet pool, I don't see any writing.

code/go/0chain.net/smartcontract/storagesc/clientpool.go
```diff
@@ -0,0 +1,722 @@
...
+
+	// remove empty allocation pool(s?)
+	cPool.removeEmptyAllocation(allocID, torm)
+	return
+}
+
+// The clientPoolReadRedeem represents part of response of read markers redeeming.
+// A Blobber uses this response for internal client pools cache.
+// Former readPoolRedeem
+// todo: should it be reafactored?
```
Sriep, line 202: How?

code/go/0chain.net/smartcontract/storagesc/clientpool.go
```diff
@@ -0,0 +1,722 @@
...
+type clientPoolReadRedeem struct {
+	PoolID  string        `json:"pool_id"` // client pool ID
+	Balance state.Balance `json:"balance"` // balance reduction
+}
+
+// For read operations only
+// todo: better rename it, adding "Read" in
+func (cPool *clientPool) moveToBlobber(ssContractKey, allocID, blobID string,
```
Sriep, line 210: I belive this is moving tokens, nothing is being read.

code/go/0chain.net/smartcontract/storagesc/clientpool.go
```diff
@@ -0,0 +1,722 @@
+package storagesc
```
Sriep, line 1: Rather than block copy code from somewhere. see if you can come up with a way of using common code wherever possable. So bugs will only have to be fixed in one place. For example consider [function.go](https://github.com/0chain/0chain/blob/966d7479999ac15ce1642e029154e73c95405ba3/code/go/0chain.net/smartcontract/storagesc/functions.go).

Sriep (changes requested):
There should be a unit test for clientpool.go. Also resolve all todos before merging.
Try and avoid block copying code if you can make common code instead.

Sriep (changes requested):
Having thought about it, I think what we need is a `Pool`  'base' object, probably best to make it an interface.

Anyway, we can then define
```go
type writePool struct {
	Pool
}
```
and 
```go
type clientPool struct {
	Pool
}
```
etc for other ...Pool types.

See what method are common to `writepool`, `stakepool`, `readpool`, ``challengepool`, clientpool` and `allocation_pool`. If the base object is an interface we don't need to worry about the implementations being different.

Issue #260, opened 2021-05-11
Race issue: Smart contract id.


Sriep:
The [SmartContract](https://github.com/0chain/0chain/blob/master/code/go/0chain.net/chaincore/smartcontractinterface/interface.go#L16) needs protection, in particular the ID field.

The [NewSC](https://github.com/0chain/0chain/blob/master/code/go/0chain.net/chaincore/smartcontractinterface/interface.go#L22) results in race issues in https://github.com/0chain/0chain/pull/258, `code/go/0chain.net/chaincore/chain/state_handler_test.go`.

<details>
  <summary>Example race errors</summary>

```
==================
WARNING: DATA RACE
Read at 0x00c0003e4a80 by goroutine 96:
  0chain.net/smartcontract/storagesc.(*StorageSmartContract).getWritePoolBytes()
      /home/piers/GolandProjects/0chain/code/go/0chain.net/smartcontract/storagesc/writepool.go:279 +0x64
  0chain.net/smartcontract/storagesc.(*StorageSmartContract).getWritePool()
      /home/piers/GolandProjects/0chain/code/go/0chain.net/smartcontract/storagesc/writepool.go:291 +0x89
  0chain.net/smartcontract/storagesc.(*StorageSmartContract).getWritePoolStatHandler()
      /home/piers/GolandProjects/0chain/code/go/0chain.net/smartcontract/storagesc/writepool.go:580 +0x8e
  0chain.net/smartcontract/storagesc.(*StorageSmartContract).getWritePoolStatHandler-fm()
      /home/piers/GolandProjects/0chain/code/go/0chain.net/smartcontract/storagesc/writepool.go:571 +0x97
  0chain.net/chaincore/smartcontract.ExecuteRestAPI()
      /home/piers/GolandProjects/0chain/code/go/0chain.net/chaincore/smartcontract/handler.go:36 +0x21c
  0chain.net/chaincore/chain.(*Chain).GetSCRestOutput()
      /home/piers/GolandProjects/0chain/code/go/0chain.net/chaincore/chain/state_handler.go:68 +0x9f3
  0chain.net/chaincore/chain.(*Chain).GetSCRestOutput-fm()
      /home/piers/GolandProjects/0chain/code/go/0chain.net/chaincore/chain/state_handler.go:50 +0x77
  0chain.net/core/common.ToJSONResponse.func1()
      /home/piers/GolandProjects/0chain/code/go/0chain.net/core/common/handler.go:133 +0xd6
[...]

PR #295, opened 2021-05-19
New smart contract factroy

Sriep:
This resolves issue https://github.com/0chain/0chain/issues/260

Currently, there is only one smart contract object for each of the smart contracts. Each time a new smart contract is requested, its contents are copied onto itself. This copying is not properly protected against concurrent access.

This PR creates a different instance of a smart contract each time one is called. A new smart contract factory object is built to create them. This extra effort of a factory interface is needed to prevent against import cycles.


bbist (commented):
Looks good.

Issue #287, opened 2021-05-17
Add 0chain developer guide


cnlangzi:
Please answer these questions before submitting your issue.

- Why do you submit this issue?
- [ ] Question or discussion
- [ ] Bug
- [ ] Requirement
- [x] Feature or performance improvement

___
### Question

___
### Bug

___
### Requirement or improvement
For now, the documentation and scripts are scattered in different repos. It is difficult to setup a local 0chain development environment for testing and debugging. 
It should be better to add a full 0chain developer guide.
- how to clone all repos
- how to build latest docker images
- how to run docker container with `x` sharders `y` miners and `z` blobbers
- how to debug sharder/miner/blobber/zboxcli/zwallet. eg.  check logs in docker container or start dlv debugger in vscode
- how to update docker images, and restart docker containers
- FAQ

cnlangzi:
There are many useful scripts on https://github.com/0chain/0chain/tree/master/quickstart/1s_2m_4b

ma2b0043:
@cnlangzi when can we expect to close this issue?

PR #296, opened 2021-05-20
Fix/developer guide

cnlangzi:
it is link to issue https://github.com/0chain/0chain/issues/287

- Local dev env will be setup in only one folder : ~/0chain/quickstart/dev.local/data .It will be easier to check all things
- added lable "zchain" to docker image and container. it let us easy to start/stop/rm them
```
docker rm `docker ps -aq -f "label=zchain"
```


Sriep (changes requested):
There might be some useful scripts here, but it has the `just another` feel to it. 

Does it provide any extra benefit over [0chain main readme](https://github.com/0chain/0chain/blob/master/README.md)?

First time I saw 0chain, I follwed the directions in the [main project readme.md](https://github.com/0chain/0chain/blob/master/README.md) and got a chain running inside half an hour. While so far this PR's readme has me just confused.

If the idea is to have a develoer guide, then there should be a lot more text, with pictures . Something a developer can refer to if they get lost.

If we want to add a few useful scripts, fine add a section to the main 0chain readme to explain their use.

cnlangzi:
@Sriep  thanks. it is [WIP], just leave it with me for now.

Issue #104, opened 2021-03-12
Optimise storagesc smart contract.


Sriep:
# storagesc

Make sure interest rate, and earnings are correct for a delegate and stakers based on write and read price, slash rate, and they get rewarded min_demand even if client does not put any storage. Also try to optimize the smart contract code for faster execution, especially when there is invalid data such as bad ID (allocation, wallet, etc)

1. Does blobber delegate receive all the storage earnings and interest based on their period. When do they receive it? Can they unstake, and if so, how much?
2. Is the storage earnings and interest received correct? How do you verify?
3. Does blobber delegate receive service charge even if he does not stake? Is it just based on storage earnings?
4. Does a client who stakes on a blobber receive storage earnings minus slashes and minus the service charge? Is the interest affected by slashes or service charge?
5. Lets say we have 2 clients, and they have unequal stake, do they receive proportionate storage earnings minus slashes and minus service charge. What about interest?
6. Are the storage earnings correct based on write and read price and slash rate? Do the blobbers receive min_demand even if client does not put any storage? 
7. When allocation expires does all tokens go to the blobber?
8.  When client cancels allocation due to failed challenges, do all the tokens go back to the client?
9. Is the staking separate for validator or they go to the blobber delegate wallet and clients who stake for blobbers.

Sriep:
Benchmark tests with 1000 miners and 1000 sharders done by a previous developer - in GitHub (find link)

#### Notes from the developer: 
Just tested how fast the transactions for 1000 blobbers. It was 141ms. The databases were empty. If we load them with something transactions will take even more time. And those tests were on a synthetic system, not the real one so the real one could even longer on empty blobbers. With loaded blobbers, it could take approximately 300-500ms.

We tested challenges. Here is the result
generateChallenges 344.41 ms/op
Did that
using: challenge_rate_per_mb_min: 1
max_challenges_per_generation: 100

For challenge_response it took 20ms and in this case, it does not matter how many allocaitons/blobbers/validators we used. Test parameters were data_shards=10 parity_shards=10


<img width="911" alt="Screen Shot 2020-12-30 at 1 21 55 PM" src="https://user-images.githubusercontent.com/6240686/110920565-724af900-8315-11eb-902a-23c4638821a2.png">

Sriep:
Unit test for payments/fees. 

[...]

PR #313, opened 2021-05-22
Immutable SmartContract objects

Sriep:
Currently, a new [SmartContract](https://github.com/0chain/0chain/blob/master/code/go/0chain.net/chaincore/smartcontractinterface/interface.go#L16) object is made for each time a smart contract is called. This is unnecessary as all the SmartContract does is hold a list of handlers and endpoints for smart contract calls. This PR makes seven immutable SmartContract objects, one for each type of smart contract.
* [SmartContractInterface](https://github.com/0chain/0chain/blob/master/code/go/0chain.net/chaincore/smartcontractinterface/interface.go#L33): Remove methods that change or access the underlying SmartContract object. It needs to immutable.
* The only reason the underlying SmartContract object was needed was to access HandlerStats or ExecutionStats, so add accessors for this to the `SmartContractInterface`.
* Modify `smartcontracts` to implement the new `SmartContractInterface` methods and remove or make local mutatling methods removed from `SmartContractInterface`.
* Change existing code to use the new `SmartContractInterface` methods rather than `SmartContract` objects.
* Remove factory.
* Update setupsc.

This PR addresses issues raised by https://github.com/0chain/0chain/issues/104. Also related to issue https://github.com/0chain/0chain/issues/260.


bbist (commented):
Looks much better.

PR #315, opened 2021-05-24
Select diverse blobbers by geo location

andrenerd:
- add a new boolean to StorageAllocation.
- add new code to select blobber so as to maximise geo-diversity.
- modify unit tests to confirm changes.

ps. the diversifying logic was added not to filterBlobbers, as proposed, but another newly created method. as the "filter" couldn't fit it well.


code/go/0chain.net/smartcontract/storagesc/models.go
```diff
@@ -649,6 +660,93 @@ List:
 	return list[:i]
 }
 
+func (sa *StorageAllocation) diversifyBlobbers(list []*StorageNode, size int) (diversified []*StorageNode) {
```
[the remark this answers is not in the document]
andrenerd: Ok.

It looks for the most "distanced" subset of blobbers. Where the distance for each group is  defined as a sum of distances between all of its "pairs"). The biggest part of the code is just a couple of subfunctions to generate an array of all possible subsets of blobbers and an array of pairs for each one of them. And to calc geo-distance for a pair of locations then.

code/go/0chain.net/smartcontract/storagesc/models_test.go
```diff
@@ -121,3 +121,86 @@ func TestStorageAllocation_filterBlobbers(t *testing.T) {
...
+		&StorageNode{Geolocation:
+			StorageNodeGeolocation{Latitude: 22.633333, Longitude: 120.266670}}, // Kaohsiung City
+	}
+
+	tests := []struct {
+		name string
+		args args
+		want want
+	}{
+		{
```
Sriep, line 151: What happens if the StorageNodes have no Latitude or Longitude information. Maybe that information is optional for blobbers. Anyway I think we should add a unit test for that.

Sriep (changes requested):
What should we do if there is no latitude or longitude information? Probably randomise them.

code/go/0chain.net/smartcontract/storagesc/models_test.go
andrenerd: If not set, there will be zeros.
I don't think "randomizing" make sense. It's hard to define what and when to randomize. For example, iff 3 of 5 geolcations are "zeros"? What should we do?

Frankly, i wouldn't overcomplicated it with randomizing. And it's not about geo (!) diversification, anyway.

Sriep: So what do we do? At the moment the code tries to diversify a set of blobbers with zero longitudes and latitude, which seems a somewhat inefficient and uncontrolled way of randomising them.

code/go/0chain.net/smartcontract/storagesc/allocation.go
```diff
@@ -304,6 +304,11 @@ func (sc *StorageSmartContract) newAllocationRequest(t *transaction.Transaction,
 		blobberNodes = randomizeNodes(list, blobberNodes, size, seed)
 	}
 
+	// diversify blobber nodes
+	if len(blobberNodes) > size {
+		blobberNodes = sa.diversifyBlobbers(blobberNodes, size)
+	}
+
```
Sriep, line 311: If  I understand the logic here we only ever call `diversifyBlobbers` if the number of preferred blobbers exceeds the number of required bobbers. A strange edge case.

I think diversifyBlobbers should be done in place of randomizeNodes depending on a configuration setting. Probably from `0chain.yaml` but confirm this. so
```go
if 0chain.yanl.deversify {
   use diversifyBlbbers //  new code
} else {
   use randomizeNodes // old code
}
```
In which case, in diversifyBlobbers we are diversifying `size` blobbers from `list` not `blobberNodes`. 

However, we should probably respect the preferred blobbers, so we should consider `blobberNodes` as containing pre-picked blobbers irrespective of diversity.

Sriep (changes requested):
We should have a config entry to toggle the diversity functionality. We can probably assume the longitude and latitude value makes sense if this is toggled on.

The confusion in this function has convinced me we need a unit test for `newAllocationRequest` that focuses on blobber selection.

code/go/0chain.net/smartcontract/storagesc/allocation.go
andrenerd: Preferred blobbers and other params are respected in the code preceding the "diversification" call. That part of the code is beyond the scope of the task. And that code hasn't been change.

The "randomization" idea is probably nice, but it's also beyond the scope. If we need to add "randomization", let's merge the "diversification" and create new task for randomization. As randomization is not part of the diversification per se, it's a part of the higher level flow, from where "diver"/"random" stuff is called.

If there is no changes to the diversification itself, let's merge and continue with further improvements.

code/go/0chain.net/smartcontract/storagesc/models_test.go
andrenerd: Imho, a test with all the zeros makes no sense, as the result is basically defined by order of the blobs in the array in such case. The test wouldn't prove/disprove the calcs.

code/go/0chain.net/smartcontract/storagesc/allocation.go
Sriep: I am not suggesting a `randomisation` idea. I am saying that currently the blobbers are randomised and that your diversifyBlobber routine is never used, except in a rare edge case. 

Step through the code in a unit test and see what happens. 
```go
	// randomize blobber nodes
	if len(blobberNodes) < size {
		var seed int64
		if seed, err = strconv.ParseInt(t.Hash[0:8], 16, 64); err != nil {
			return "", common.NewError("allocation_creation_failed",
				"Failed to create seed for randomizeNodes")
		}
		blobberNodes = randomizeNodes(list, blobberNodes, size, seed)
	}

	// diversify blobber nodes
	if len(blobberNodes) > size {
		blobberNodes = sa.diversifyBlobbers(blobberNodes, size)
	}
```
usually, len(blobberNodes) starts around zero. So first check `len*blobberNodes<size` true so `blobberNodes = randomizeNodes(list, blobberNodes, size, seed)` sets the number of blobberNodes to size. Next when `len(blobberNodes) > size ` is done, the len(blobberNodes) is usually size so the check fails and  diversifyBlobbers does not get called.

Really it's an `either or`. Either use the **existing** `randomiseNodes` or the new `diversifyBlobbers`. With this PR it still uses the old `randomixeNodes` I am suggesting a config entry to choose between them.

andrenerd: @Sriep 
let's start from the basics (to clarify that "strange"):
"
If I understand the logic here we only ever call diversifyBlobbers if the number of preferred blobbers exceeds the number of required bobbers. A strange edge case.
"

- - -

Let's assume, we have N items.
And we need to select subset of S items based on a CRITERIA, ranging subsets of the SAME power (in our case it's geo loc max distance, but doesn't matter for now).
And also the subset should include not less than M items (derived from another conditions for blobbers, see "size" param).

- - -

When we choose subset of S items from N item universe by the criteria, we need to know the both values. Otherwise the task makes no sense!! Because the criteria ranges subsets of the SAME power only (see above).
As we have S and N (where S <=N by definition), also we can compare the values with M prior running the cals:
a) N < M: no need to calculate we have not enough values anyway. so all items to be included to the final set (=subset). 
b) N = M: no need to calculate also. as all items to be included to the final set (=subset).
c) N > M: as we have more value than we need, lets select subset of S items by the criteria. Where M <= S <= N by definition.

That's how we come the that non-strange "if" statement, combining pre-conditions (a) and (b):
```
	if len(blobberNodes) > size {
		blobberNodes = sa.diversifyBlobbers(blobberNodes, size)
	}
```

PS.
Next, i can comment why this statement is place here, not in other part of the func and so on.

PPS.
The algo hss nothing with "randomization". It fits the requirements of the issue, which don't say a word about randomization there.

Sriep: I can't approve this PR because it does not change anything.

The new diversifyBlobbers function is never called, except in the edge case of more than size preferred blobbers.
```
	if len(blobberNodes) > size {
		blobberNodes = sa.diversifyBlobbers(blobberNodes, size)
	}
```
`len(blobberNodes)` will almost always equal `size` so `diversifyBlobbers` is never called.

Use a unit test to step through the code.

guruhubb: Yes perhaps we can deprecate randomization in favor of geodiversity based randomization, because I don't see the value of plain randomization.

Sriep: The advantage of randomisation over a deterministic algorithm is that we avoid the same blobbers being picked all the time. 

Also is there an issue with blobber latitude and lonitude being set so that particular blobber always get picked. We should also [validate longitude and latitude somewhere.](https://stackoverflow.com/questions/15965166/what-are-the-lengths-of-location-coordinates-latitude-and-longitude)

guruhubb: Yes agree. @andrenerd pls add validation in the blobber.

andrenerd: @Sriep @guruhubb 
validation: will be added.

PS. @Sriep 
- "I can't approve this PR because it does not change anything.": sorry, what?
- "Try to get the test to call diversifyBlobbers.": check the test, there are cases when N > M. and the diversifier does it job in such cases.
- "never called, except in the edge case of more than size preferred blobbers.": please, read my reply above again, it has all the answers. it's not "edge case" or "strange case", as you like to call it. it's the case when diversification make sense to apply. in all the other case the diversifier should return the same set as it receives (because M >= N). so no reason to call it.

PPS.
"The advantage of randomisation over a deterministic algorithm is that we avoid the same blobbers being picked all the time.": yep, if we apply geo diversifier to the same set of blobbers (and M is the same), we always pickup the same subset of blobbers. was it unclear at the begging? and why geo diversification was proposed?

PPPS. 
potentially, i can add "randomizing" factors to the geo diversifier. currently it select the most distanced subset of blobbers. it could be two things added:
- replace S = M with S = rand[M...N-1]
- replace  _the most distanced subset_ with _a random subset of the SS% more distanced subsets_, where SS is discussable.

code/go/0chain.net/smartcontract/storagesc/models_test.go
Sriep: If someone just uses this change without their blobbers then all zeros is what we will be getting. An all-zero test makes sure the method does not error out but does something sensible.

Sriep:
Created gosdk https://github.com/0chain/gosdk/pull/100 change https://github.com/0chain/gosdk/pull/100 to activate diversify blobbers in 0cahin.

Sriep:
<details>
  <summary>Fixed lint issues in alloaction_test.go</summary>

```
smartcontract/storagesc/allocation_test.go:1072:6: `addBloberChallenges` is unused (deadcode)
func addBloberChallenges(t *testing.T, sscID string, alloc *StorageAllocation,
     ^
smartcontract/storagesc/allocation_test.go:585:2: ineffectual assignment to resp (ineffassign)
	resp, err = ssc.newAllocationRequest(&tx, mustEncode(t, &nar), balances)
	^
smartcontract/storagesc/allocation_test.go:1206:2: ineffectual assignment to wp (ineffassign)
	wp, err = ssc.getWritePool(client.id, balances)
	^
smartcontract/storagesc/allocation_test.go:1210:2: ineffectual assignment to cp (ineffassign)
	cp, err = ssc.getChallengePool(allocID, balances)
	^     
smartcontract/storagesc/allocation_test.go:226:5: SA4006: this value of `alloc` is never used (staticcheck)
	if alloc, err = ssc.getAllocation(allocID, balances); err == nil {
	   ^
```

</details>

Sriep (approved):

Issue #263, opened 2021-05-11
Race issue: Protect core.memorystore.pools


Sriep:
Package wide global map [pools](https://github.com/0chain/0chain/blob/master/code/go/0chain.net/core/memorystore/connection.go#L54) needs protection.

Race error detected in PR https://github.com/0chain/0chain/pull/258 `code/go/0chain.net/core/memorystore/connection_test.go`

<details>
  <summary>TestAddPool</summary>

```
==================
WARNING: DATA RACE
Write at 0x00c0001c4660 by goroutine 62:
  runtime.mapassign_faststr()
      /usr/local/go/src/runtime/map_faststr.go:202 +0x0
  0chain.net/core/memorystore.AddPool()
      /home/piers/GolandProjects/0chain/code/go/0chain.net/core/memorystore/connection.go:71 +0x168
  0chain.net/core/memorystore.TestAddPool.func3()
      /home/piers/GolandProjects/0chain/code/go/0chain.net/core/memorystore/connection_test.go:233 +0x68
  testing.tRunner()
      /usr/local/go/src/testing/testing.go:1194 +0x202

Previous write at 0x00c0001c4660 by goroutine 63:
  runtime.mapassign_faststr()
      /usr/local/go/src/runtime/map_faststr.go:202 +0x0
  0chain.net/core/memorystore.AddPool()
      /home/piers/GolandProjects/0chain/code/go/0chain.net/core/memorystore/connection.go:71 +0x168
  0chain.net/core/memorystore.TestAddPool.func3()
      /home/piers/GolandProjects/0chain/code/go/0chain.net/core/memorystore/connection_test.go:233 +0x68
  testing.tRunner()
      /usr/local/go/src/testing/testing.go:1194 +0x202

Goroutine 62 (running) created at:
  testing.(*T).Run()
      /usr/local/go/src/testing/testing.go:1239 +0x5d7
  0chain.net/core/memorystore.TestAddPool()
      /home/piers/GolandProjects/0chain/code/go/0chain.net/core/memorystore/connection_test.go:231 +0x294
  testing.tRunner()
      /usr/local/go/src/testing/testing.go:1194 +0x202

Goroutine 63 (running) created at:
  testing.(*T).Run()
      /usr/local/go/src/testing/testing.go:1239 +0x5d7
  0chain.net/core/memorystore.TestAddPool()
      /home/piers/GolandProjects/0chain/code/go/0chain.net/core/memorystore/connection_test.go:231 +0x294
[...]

Issue #265, opened 2021-05-11
Race Issue: Protect package global map connections


Sriep:
Protect the package global map [connections](https://github.com/0chain/0chain/blob/master/code/go/0chain.net/core/memorystore/connection.go#L143) from concurrent access. Update `code/go/0chain.net/core/memorystore/connection_test.go`.

Generated a race issue in PR https://github.com/0chain/0chain/pull/258 `code/go/0chain.net/core/memorystore/connection_test.go`

<details>
  <summary>TestClose/Test_Close_OK2</summary>

```
==================
WARNING: DATA RACE
Read at 0x00c0005ac308 by goroutine 102:
  github.com/gomodule/redigo/redis.(*activeConn).Close()
      /home/piers/GolandProjects/0chain/code/go/pkg/mod/github.com/gomodule/redigo@v2.0.0+incompatible/redis/pool.go:395 +0x64
  0chain.net/core/memorystore.Close()
      /home/piers/GolandProjects/0chain/code/go/0chain.net/core/memorystore/connection.go:244 +0x14c
  0chain.net/core/memorystore.TestClose.func1()
      /home/piers/GolandProjects/0chain/code/go/0chain.net/core/memorystore/connection_test.go:849 +0x58
  testing.tRunner()
      /usr/local/go/src/testing/testing.go:1194 +0x202

Previous write at 0x00c0005ac308 by goroutine 174:
  github.com/gomodule/redigo/redis.(*activeConn).Close()
      /home/piers/GolandProjects/0chain/code/go/pkg/mod/github.com/gomodule/redigo@v2.0.0+incompatible/redis/pool.go:399 +0x8f
  0chain.net/core/memorystore.Close()
      /home/piers/GolandProjects/0chain/code/go/0chain.net/core/memorystore/connection.go:244 +0x14c
  0chain.net/core/memorystore.TestClose.func1()
      /home/piers/GolandProjects/0chain/code/go/0chain.net/core/memorystore/connection_test.go:849 +0x58
  testing.tRunner()
      /usr/local/go/src/testing/testing.go:1194 +0x202

Goroutine 102 (running) created at:
  testing.(*T).Run()
      /usr/local/go/src/testing/testing.go:1239 +0x5d7
  0chain.net/core/memorystore.TestClose()
      /home/piers/GolandProjects/0chain/code/go/0chain.net/core/memorystore/connection_test.go:847 +0xa2e
  testing.tRunner()
[...]

PR #316, opened 2021-05-25
Race issue: Protect core.memorystore.pools #263

platsko:


Sriep (commented):
I suggest only make changes need for the `pools` object, and leave any chances to the connection code for issue https://github.com/0chain/0chain/issues/265. 

In fact, it might be clearer to make a separate file for the `pool` object.

Sriep:
* Split `connections.go` up. Created `pool.go` for the poolList type. Split up unit test.
* Created poolList methods for the functions that needed poolList mutex.
* Global functions changed to wrap poolList methods
* Removed getConnectionCount which was not being used.
* Some CamelCase refactoring.

Sriep (approved):

PR #325, opened 2021-05-31
Unit test doc

Sriep:
Document to help standardise 0chian unit tests.  Please suggest input.

### DIFF
--- a/docs/unit-tests.md
+++ b/docs/unit-tests.md
@@ -387,12 +387,17 @@ so that we can mock them.
 Having done the preliminary steps, we have a pile object definitions, 
 and a list of test we wish to run. However, everything we need should be
 already be there. We need to shuffle everything around so all our tests
-work.
+work. 
 
 Remember that care needs to be taken to avoid sharing memory between tests.
 Copy by value not pointer. 
 
+If you feel tempted to set up a working environment for your test
+function to work in, then don't. Really don't. Strive to minimise the
+amount of setup you need.
+
 Idiot check the function's purpose. Yuo should know by now, but just make sure.
+Do your tests align with these purposes.
 
 ## Mocks
 
--- a/docs/unit-tests.md
+++ b/docs/unit-tests.md
@@ -1,9 +1,18 @@
 # Unit tests
 
+## Table of Contents
+
+- [Intorduction](#introduction)
+- [0chain style](#0chain-style)
+- [Mocks](mocks)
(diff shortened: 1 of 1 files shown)


docs/unit-tests.md
```diff
@@ -0,0 +1,445 @@
...
+### Vektra mockery
+
+To help with our unit tests, in `0chain`, we autogenerate mock object for all our
+interface objects using the
+[vektra/mockery](https://github.com/vektra/mockery) package. 
+
+As we autogenerate these mock object, we avoid keeping them under
+version control. Instead, we generate them when we run unit tests during 
+continuous integration. Developers can generate them using the 
+[0chain\generate_mcoks.sh](https://github.com/0chain/0chain/blob/master/generate_mocks.sh) script.
```
shravanshetty1, line 289: typo

shravanshetty1 (changes requested):

Dmdv:
@Sriep Do we use Ginkgo and gomock or it is not encouraged?

Dmdv:
@Sriep   
I found no mentions about a convention where to keep all test factory code.  
Imagine not interfaces but types which are not generated by **mockery**.  
Some use-test cases are repetitive throughout the same category tests and factory for said types are definitely worth keeping in the separate directory rather than being kept in a single file.

Keeping all test types factories make the code clumpy

Sriep:
```
Do we use Ginkgo and gomock or it is not encouraged?
```
Don't use Ginko or gomock unless we decide to use gomock or Ginkgo.

Sriep:
```
I found no mentions about a convention where to keep all test factory code.
```
How do you mean `test factory code?` We want to keep tests decoupled where possible. 

Some common functions maybe, but we want to avoid a large amount of common setup code 
that is not necessary for any particular test.

Dmdv:
> ```
> I found no mentions about a convention where to keep all test factory code.
> ```
> 
> How do you mean `test factory code?` Keep tests decoupled where possible. Also, try to avoid a large amount of common setup code that is not necessary for any particular test.

@Sriep 
Do you mean keeping all tests outside of the package folder in the test folder or keeping all tests alongside the tested files?

As for test factory code, I mean some common types created in different tests with copy-pasting. 
Do we place them in a separate folder?

Sriep:
For common methods and objects just be sensible. A common help _test file is fine.

PR #339, opened 2021-06-02
Protect memroystore connection obejct with a mutex

Sriep:
Wrapped the [memorystore.connection](https://github.com/0chain/0chain/blob/c244ed39a07ac582afff9e7ae8356e9a93e395a8/code/go/0chain.net/core/memorystore/connection.go#L78) map in an object and protected get and set methods with a mutex.
Addresses issue https://github.com/0chain/0chain/issues/265.

### DIFF
--- a/code/go/0chain.net/core/memorystore/connection.go
+++ b/code/go/0chain.net/core/memorystore/connection.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"regexp"
+	"sync"
 	"time"
 
 	"0chain.net/core/common"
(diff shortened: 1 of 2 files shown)


platsko (approved):
Not sure it is good practice to use a pointer to mutex: https://github.com/uber-go/guide/blob/master/style.md#zero-value-mutexes-are-valid

Sriep:
The problem is lint raises an object to copying mutex.
```
core/memorystore/connection.go:120:45: copylocks: call of context.WithValue copies lock value: 0chain.net/core/memorystore.connections contains sync.RWMutex (govet)
		return context.WithValue(ctx, CONNECTION, cMap)
		                                          ^
core/memorystore/connection.go:169:45: copylocks: call of context.WithValue copies lock value: 0chain.net/core/memorystore.connections contains sync.RWMutex (govet)
		return context.WithValue(ctx, CONNECTION, cMap)
		                                          ^
core/memorystore/connection.go:211:10: copylocks: assignment copies lock value to cMap: 0chain.net/core/memorystore.connections contains sync.RWMutex (govet)
	cMap := c.(connections)
	        ^
```

PR #344, opened 2021-06-03
Miner unit tests

### LINKED ISSUE #343: Miner unit tests

TheTeaParty:
Closes #343


code/go/0chain.net/miner/block_notarization_entity_test.go
```diff
@@ -0,0 +1,56 @@
...
+	"0chain.net/core/mocks"
+	"github.com/stretchr/testify/suite"
+	"testing"
+)
+
+type BlockNotarizationEntityTestSuite struct {
+	suite.Suite
+}
+
+func TestBlockNotarizationEntityTestSuiteSuite(t *testing.T) {
```
Sriep, line 14: You don't seem to actually be testing anything in these tests.

Proably a key point here is that `Notarization` satisfies both the `Entity` and `ReadLockable` interfaces.

You can test DoReadLock and DoReadUnlock by calling various read and write calls on the `Entity` interface in Parallel. Then run `go test -race` to show there are no concurrency errors.

code/go/0chain.net/miner/chain_test.go
```diff
@@ -0,0 +1,390 @@
...
+	"github.com/rafaeljusto/redigomock"
+	"github.com/stretchr/testify/mock"
+	"github.com/stretchr/testify/suite"
+	"go.uber.org/zap"
+	"net/http"
+	"testing"
+)
+
+type ChainTestSuite struct {
+	suite.Suite
```
Sriep, line 28: Don't use test Suite unless you need them. Use usally don't.

code/go/0chain.net/miner/chain_test.go
```diff
@@ -0,0 +1,390 @@
...
+	suite.Run(t, &ChainTestSuite{})
+}
+
+func (s *ChainTestSuite) TestChainStarted() {
+	c := chain.NewChainFromConfig()
+	SetupMinerChain(c)
+
+	mc := GetMinerChain()
+
+	mc.ChainStarted(context.Background())
```
Sriep, line 45: So ChainStarted runs in a loop unitil some vales are set suitably.

I woud suggest running tests by
1. call ChainStarted.
2. Modify the chain object occording to test paramters.
3. Confirm ChainStarted exits with the response the test wants.

code/go/0chain.net/miner/chain_test.go
```diff
@@ -0,0 +1,390 @@
...
+
+}
+
+func (s *ChainTestSuite) TestCreateRound() {
```
Sriep, line 49: You can skip doing tests for simple functions like accessors or object creation.

code/go/0chain.net/miner/chain_test.go
```diff
@@ -0,0 +1,390 @@
+package miner
```
Sriep, line 1: chain.go looks like a difficult to write a unit test for. 

Pick a method and focus on testing only the input and outputs of the method. Remeber that chain satisfies the [Entity](https://github.com/0chain/0chain/blob/master/code/go/0chain.net/core/datastore/entity.go#L19), [Stoppable](https://pkg.go.dev/github.com/rcrowley/go-metrics#Stoppable), [CreateTrackable](https://github.com/0chain/0chain/blob/master/code/go/0chain.net/core/datastore/creation_date.go#L9) and [Chainer](https://github.com/0chain/0chain/blob/master/code/go/0chain.net/chaincore/chain/block_fetcher.go#L316) interfaces, so if a test fucntion calls any of these interface methods you can mock the response.

Sriep (changes requested):
You can check out my [unit test document](https://github.com/0chain/0chain/pull/325), in particular the writing a unit test section.

Sriep:
This Pr does not seem to be going anywhere. Closing.

PR #349, opened 2021-06-07
Improve unit_test.sh and unit_test_build.sh

### LINKED ISSUE #348: Improve unit_test.sh and unit_test_build.sh
Do not rebuild whole container on source code change.

TheTeaParty:
Closes #348


code/go/0chain.net/mocks/state-context-i.go
```diff
@@ -0,0 +1,334 @@
+// Code generated by mockery v2.8.0. DO NOT EDIT.
```
Sriep, line 1: Why is this file here?

generate_mocks.sh
```diff
@@ -4,25 +4,25 @@ BASEDIR=$(pwd)
...
+cd $BASEDIR/0chain.net/sharder || exit
 mockery --output=../sharder/mocks --all
 
-cd $BASEDIR/code/go/0chain.net/smartcontract || exit
+cd $BASEDIR/0chain.net/smartcontract || exit
 mockery --output=../smartcontract/mocks --all
 
-cd $BASEDIR/code/go/0chain.net/chaincore/chain/state || exit
+cd $BASEDIR/0chain.net/chaincore/chain/state || exit
 mockery --name=StateContextI --output=../../../mocks --filename=state-context-i.go
```
Sriep, line 26: Why are we putting a second copy of  StateContextI.go here. I know its popular, but having two copies of the same file is usally bad.

Sriep (changes requested):
Confused by state-context-i.go changes. Does not seem to have anything to do with issue https://github.com/0chain/0chain/issues/348.

Why do we need empty log directories?

code/go/0chain.net/mocks/state-context-i.go
TheTeaParty: This is not my change tbh, someone added it and not .gitignored it, will add to gitignore

generate_mocks.sh
TheTeaParty: this is not my change, it is better to ask original creator of it @MurashovVen .

Sriep:
Please remove all changes that are not part of this PR = `Improve unit_test.sh and unit_test_build.sh`.

Sriep:
Take into account PR https://github.com/0chain/0chain/pull/353. We are starting to run [shellcheck](https://github.com/koalaman/shellcheck) on shell scripts. So going forward new shell scripts should be checked with shellcheck.

TheTeaParty:
> Changes that are not part of this PR = `Improve unit_test.sh and unit_test_build.sh` should be removed. E.g.,
> 
> 1. .gitkeep files in log directoires
> 2. changes to .gitignore
> 3. 0chian.net.co/mocks/state-context-i.go
> 4. changes to generate_mocks.sh not needed for unit test dockerfile changes.

1. logs are added to .gitignore to keep directories used in runtime, as volumes are used, and it is to avoid permission issues
2. same as 1
3. Not part of my change but will add it to gitignore
4. not quite undeerstood

Thanks.

Sriep:
Mock directories are made with root privilege, this causes a lot of problems for user development tools.

Sriep:
I think this will cause more problems than it solves, so closing.

PR #351, opened 2021-06-09
Added M1 support to docker files

Dmdv:
## Fixing docker builds of AMD/Intel on Apple M1 chips
You can use buildkit, and then:
```docker buildx build --platform linux/amd64 .```
or change Dockerfile definition with
```FROM --platform=linux/amd64 golang:1.12.4-alpine3.9```

### Testing

Following scripts support flags: m1, -m1, --m1

```
./docker.local/bin/build.base.sh -m1
./docker.local/bin/build.miners.sh -m1
./docker.local/bin/build.sharders.sh  -m1
./docker.local/bin/build.magic_block.sh -m1
```

### Prerequisites:

https://docs.docker.com/docker-for-mac/apple-silicon

### Other links:  

https://docs.docker.com/buildx/working-with-buildx
https://docs.docker.com/engine/reference/commandline/buildx

### CODE REVIEW

File: docker.local/bin/build.miners.apple.sh [hunk 1]
```diff
@@ -0,0 +1,40 @@
...
+    echo -e "\nDevelopment mode: building miner locally\n"
+
+    cd "$ROOT/code/go/0chain.net/miner/miner"
+    go build -v -tags "bn256 development" \
+        -ldflags "-X 0chain.net/core/build.BuildTag=$GIT_COMMIT"
+
+    sed 's,%COPY%,COPY ./code,g' "$DOCKERFILE.template" > "$DOCKERFILE"
+
+    cd "$ROOT"
+    docker buildx build --platform linux/amd64 --build-arg GIT_COMMIT=$GIT_COMMIT \
```
mallochine, line 23: is `linux/amd64` the only different with the other build.miners.sh? Since this is a shell script, they have easy support for flags. So we should really make that a flag. (Dockerfiles do NOT have good support for flags)

docker.local/bin/build.sharders.apple.sh
```diff
@@ -0,0 +1,14 @@
+#!/bin/sh
+set -e
+
+GIT_COMMIT=$(git rev-list -1 HEAD)
+echo $GIT_COMMIT
+
+docker buildx build --platform linux/amd64 --build-arg GIT_COMMIT=$GIT_COMMIT -f docker.local/build.sharder/Dockerfile . -t sharder
```
mallochine, line 7: this should be a flag if the change is that trivial.

like if --m1 flag is set then linux/amd64

BTW should not be --apple, there are still many apple computers with intel chip

mallochine (commented):
you should also make a note in top-level README.md regarding how to use your script if dev needs to compile on m1 chip.

mallochine:
ok tested, all the build commands work on my machine

docker.local/bin/build.miners.apple.sh
Dmdv: same for other scripts

Dmdv:
@mallochine Could you pls check again after the fix

docker.local/bin/build.sharders.apple.sh
Dmdv: Fixed

docker.local/bin/build.miners.apple.sh
Dmdv: Fixed

Dmdv:
@mallochine ready for review

mallochine (approved):

Dmdv:
@mallochine Anything else from my side to push this PR?

PR #353, opened 2021-06-14
Shell Checks

NoSkillGuy:
## Proposed changes

Since we deal with a lot of shell scripts, ensuring best practices is the way going forward.  I have used [this](https://github.com/koalaman/shellcheck) static code analyzer to detect the issues and fix them, 
it's also available as a plugin among various code editors and has integrations with multiple CI/CD platforms. 

Few corrections that I have solved
```
Double quote to prevent globbing and word splitting.
See SC2086(https://github.com/koalaman/shellcheck/wiki/SC2086).

Quote this to prevent word splitting.
See SC2046(https://github.com/koalaman/shellcheck/wiki/SC2046).

Use $(...) notation instead of legacy backticked `...`.
See SC2006(https://github.com/koalaman/shellcheck/wiki/SC2006).

Useless cat. Consider 'cmd < file | ..' or 'cmd file | ..' instead.
See SC2002 (https://github.com/koalaman/shellcheck/wiki/SC2002).

fgrep is non-standard and deprecated. Use grep -F instead.
See SC2197 (https://github.com/koalaman/shellcheck/wiki/SC2197). 
```

## Further comments

[_CI Suggestion_]: We can add a CI Pipeline to check if all the best practices are followed while adding/editing more shell files.

### DIFF
--- a/docker.local/bin/build.miners-integration-tests.sh
+++ b/docker.local/bin/build.miners-integration-tests.sh
@@ -2,7 +2,7 @@
 set -e
 
 GIT_COMMIT=$(git rev-list -1 HEAD)
-echo $GIT_COMMIT
+echo "$GIT_COMMIT"
 
 docker build --build-arg GIT_COMMIT=$GIT_COMMIT -f docker.local/build.miner/Dockerfile.integration_tests . -t miner
 
--- a/docker.local/bin/build.miners-integration-tests.sh
+++ b/docker.local/bin/build.miners-integration-tests.sh
@@ -4,11 +4,11 @@ set -e
 GIT_COMMIT=$(git rev-list -1 HEAD)
 echo "$GIT_COMMIT"
 
-docker build --build-arg GIT_COMMIT=$GIT_COMMIT -f docker.local/build.miner/Dockerfile.integration_tests . -t miner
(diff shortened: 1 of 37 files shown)


stewartie4:
Agreed, let's add this to our CI and make it fail the build if possible. Now we've fixed all the issues, lets make sure none of them re-occur in future

stewartie4 (approved):

Sriep (approved):
Looks ok, except the merge conflicts need to be resolved. Added issue https://github.com/0chain/0chain/issues/356 for CI change.

Sriep:
You have only changed shell scripts in `docker.local/bin`. Is there a reason for this restriction?

There are `.sh` files elsewhere, for example, the root folder has `generate.mocks.sh` and `build_docker.sh` both of which showed issues when I run `shellcheck`.

NoSkillGuy:
@Sriep Thanks for reviewing the PR. I will fix those merge conflicts. I will do it for all the shell files. I just did it for `docker.local/bin` as a pilot. I will extend it to all the other shell files

NoSkillGuy:
@Sriep I have updated the PR resolving the conflicts, I think PR is good for a merge now. 
I will raise a follow-up PR that would solve the shell checks for the entire project

PR #357, opened 2021-06-17
Fixing Authorizers issues, adding more tests for zcnsc

Dmdv:
This PR will add:
1. Unit tests for zcnsc smart contract
2. Authorizer issues: LockPool and registering authorizer node to the chain
3. Minor refactoring
4. AuthorizerNodes and AuthorizerNode serialization didn't work. I've implemented serialization
5. TokenPool didn't serialize along with the authorizer node. Adding authorizer wouldn't work
6. Fixed authorizers nodes initialization
7. Fixed token pool initialization


code/go/0chain.net/smartcontract/zcnsc/authorizer_test.go
```diff
@@ -0,0 +1,532 @@
...
+func init() {
+	rand.Seed(time.Now().UnixNano())
+	chain.ServerChain = new(chain.Chain)
+	chain.ServerChain.Config = new(chain.Config)
+	chain.ServerChain.ClientSignatureScheme = "bls0chain"
+
+	logging.Logger = zap.NewNop()
+}
+
+func Test_TransferStateAfterAddingAuthorizer(t *testing.T) {
```
Sriep, line 33: Here and everywhere else: I think it is better style to have one test for each function and use subtests for different cases. 
`TestAddAuthorizer` with subtest "transfer state after adding authorizer".

This looks like you are testing 30 functions reather than two.

code/go/0chain.net/smartcontract/zcnsc/test_factory_helpers.go
```diff
@@ -0,0 +1,130 @@
...
+}
+
+func createUserNode(id string, nonce int64) *userNode {
+	return &userNode{
+		ID:    id,
+		Nonce: nonce,
+	}
+}
+
+func addAuthorizer(t *testing.T, contract *ZCNSmartContract, ctx state.StateContextI, clientId, pk string) {
```
Sriep, line 120: This might be better as a local function in `TestAddAuthorizer`.  Is it likely to be used elsewhere?

Sriep (changes requested):
This adds a new smartcontract, `zcnsc`. Is there a link to an issue or document giving background for this new smart contract?

code/go/0chain.net/smartcontract/zcnsc/test_factory_helpers.go
Dmdv: @Sriep Actually, yes, this is used in minting tests

code/go/0chain.net/smartcontract/zcnsc/authorizer_test.go
Dmdv: @Sriep 
It would be, but to fit all tests into a single test, the **data** should be:
1. Homogeneous
2. Preconditions should be the same

**The logic flow** should be the same for all use cases.

I reviewed my tests again and saw that they cannot be converged into a single-use case, and the input data cannot be made homogeneous. Some tests look similar.

Dmdv:
> This adds a new smartcontract, `zcnsc`. Is there a link to an issue or document giving background for this new smart contract?

Yes, we have a document for this new SC, in PDF.
As for the diagrams, I'll get them prepared soon

Dmdv:
New PR instead of this includes the latest fixes to master

PR #359, opened 2021-06-17
[WIP] added shell linter to github workflows

NoSkillGuy:


NoSkillGuy:
@guruhubb can you approve the new workflow

NoSkillGuy:
@Sriep can you review this PR and approve the new workflows

Sriep:
The master branch **has** to be able to pass checks. So this can't be merged until the lint-sh check passes.

So either
1. The sh errors that are preventing the shellcheck passing need to be fixed.
2. Some directories are exempted from the shellcheck, if there is a good reason to do this.
3. If fixing all the errors is too much work at the moment, could do something like [lint errors in linter.sh](https://github.com/0chain/0chain/blob/master/docker.local/bin/linter.sh#L9), where we check the number of errors does not increase.

Sriep:
Has been abandoned. Close for the present.

PR #367, opened 2021-06-21
Return empty list, if there is no challenges

andrenerd:
Blobbers start to try to fetch open challenges prior any challenge (incl. challenge pool) is created.
Sharders should return empty list, not error message, if there are no challenges/challenge pool yet.

Screenshot 2021-06-21 at 06 20 53

PS. It could be extended later to return error for non-registered blobbers. When blobber is unknown. Even though it's not critical.


Sriep:
unit test `TestChain_HandleSCRest_Status` needs fixing.

PR #386, opened 2021-07-06
Update commands

hm90121:
get rid of the $ so that the copy is exact command


Sriep:
build base check should not be failing. Please merge in the latest master changes and fix any conflcts.

PR #389, opened 2021-07-08
Allocation payments, spread cost across all write pools

Sriep:
Currently, only the owner of an allocation can pay for upgrades and write markers. This PR changes it so that funds can be sourced from all allocation pools linked to an allocation. 

1. New WritePoolOwners field added to `StorageAllocation` to keep users with linked allocation pools. Modifications to various methods to populate this field when a new allocation pool is applied as in writePoolLock. 
2. New `allocationWritePools` object. This temporary object holds the write pools and allocation pools linked to an allocation. It is used in place of a `wrtiePool` in the endpoints `commitBlobberConnection` and `updateAllocationRequest`.
3. New method `getAllocationPools` added to `StorageAllocation` to return the allocation's `allocationWritePools` object.
4. writePools methods modified to be used by allocationWrtePools, and various updates to apply these changes
   * `moveToChallange` changed to be a`allocationPools` method so allocationWritePools.allocationPools can use it.
   * `fill` method changed to return an allocation pool rather than add to a `writePool`. The new function renamed to `newAllocatinoPool`  to better suit its purpose. The add to write pool bit being moved outside the method.
5. Couple of unused methods removed from `allocationPools`.
6. New unit test for `extendAllocation`.
7. Fixes to various unit tests that were broken by the above changes.

### DIFF
--- a/code/go/0chain.net/smartcontract/storagesc/allocation.go
+++ b/code/go/0chain.net/smartcontract/storagesc/allocation.go
@@ -609,9 +609,14 @@ func weightedAverage(prev, next *Terms, tx, pexp, expDiff common.Timestamp,
 
 // The adjustChallengePool moves more or moves some tokens back from or to
 // challenge pool during allocation extending or reducing.
-func (sc *StorageSmartContract) adjustChallengePool(alloc *StorageAllocation,
-	wp *writePool, odr, ndr common.Timestamp, oterms []Terms,
-	now common.Timestamp, balances chainstate.StateContextI) (err error) {
+func (sc *StorageSmartContract) adjustChallengePool(
+	alloc *StorageAllocation,
+	awp *allocationWritePools,
+	odr, ndr common.Timestamp,
+	oterms []Terms,
+	now common.Timestamp,
+	balances chainstate.StateContextI,
+) (err error) {
 
 	var (
 		changes = alloc.challengePoolChanges(odr, ndr, oterms)
@@ -628,15 +633,18 @@ func (sc *StorageSmartContract) adjustChallengePool(alloc *StorageAllocation,
 		var blobID = alloc.BlobberDetails[i].BlobberID
 		switch {
 		case ch > 0:
-			err = wp.moveToChallenge(alloc.ID, blobID, cp, now, ch)
+			err = awp.moveToChallenge(alloc.ID, blobID, cp, now, ch)
 			changed = true
 		case ch < 0:
 			// only if the challenge pool has the tokens; all the tokens
 			// can be moved back already, or moved to a blobber due to
 			// challenge process
 			if cp.Balance >= -ch {
-				err = cp.moveToWritePool(alloc.ID, blobID, alloc.Until(), wp,
-					-ch)
+				wp, err := awp.getOwnerWP()
+				if err != nil {
+					return fmt.Errorf("adjust_challenge_pool: %v", err)
+				}
+				err = cp.moveToWritePool(alloc, blobID, alloc.Until(), wp, -ch)
 				changed = true
 			}
 		default:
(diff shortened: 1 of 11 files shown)


code/go/0chain.net/smartcontract/storagesc/allocation.go
```diff
@@ -609,9 +610,14 @@ func weightedAverage(prev, next *Terms, tx, pexp, expDiff common.Timestamp,
 
 // The adjustChallengePool moves more or moves some tokens back from or to
```
cnlangzi, line 611: Could we write comment follow https://blog.golang.org/godoc ? it is helpful for us to generate godoc in the future.

code/go/0chain.net/smartcontract/storagesc/stakepool.go
```diff
@@ -564,11 +567,12 @@ func (sp *stakePool) slash(allocID, blobID string, until common.Timestamp,
 	// related stake holders, that can loose some tokens due to
 	// division error;
 
-	var ap = wp.allocPool(allocID, until)
+	var ap = wp.allocPool(alloc.ID, until)
```
cnlangzi, line 570: Could we check `alloc` before access `alloc.id` ?.  App will crash when it is nil. It might be an unpredictable crash on production.

code/go/0chain.net/smartcontract/storagesc/models.go
```diff
@@ -610,6 +614,56 @@ func (sa *StorageAllocation) restMinLockDemand() (rest state.Balance) {
 	return
 }
 
+func (sa *StorageAllocation) addWritePoolOwner(userId string) {
+	for _, id := range sa.WritePoolOwners {
```
cnlangzi, line 618: It might have `nil pointer dereference` crash too.

code/go/0chain.net/smartcontract/storagesc/allocation_pool.go
```diff
@@ -142,6 +144,47 @@ type allocationPool struct {
...
+			Amount:     state.Balance(t.Value),
+		})
+	} else {
+		if err = balances.AddTransfer(transfer); err != nil {
+			return nil, fmt.Errorf("adding transfer to write pool: %v", err)
+		}
+	}
+
+	// set fields
+	ap.AllocationID = alloc.ID
```
cnlangzi, line 179: It might have `nil pointer dereference` crash too.

code/go/0chain.net/smartcontract/storagesc/allocation_pool.go
```diff
@@ -142,6 +144,47 @@ type allocationPool struct {
...
+		}
+	}
+
+	var ap allocationPool
+	var transfer *state.Transfer
+	if transfer, _, err = ap.DigPool(t.Hash, t); err != nil {
+		return nil, fmt.Errorf("digging write pool: %v", err)
+	}
+	if mintNewTokens {
+		balances.AddMint(&state.Mint{
```
cnlangzi, line 167: should we check the result of `balances.AddMint` ?

code/go/0chain.net/smartcontract/storagesc/allocation_write_pools.go
```diff
@@ -0,0 +1,67 @@
...
+type allocationWritePools struct {
+	// The indices for ids and writePools match
+	ownerId         int
+	ids             []string
+	writePools      []*writePool
+	allocationPools allocationPools
+}
+
+func (awp *allocationWritePools) getOwnerWP() (*writePool, error) {
+	if len(awp.writePools) == 0 {
```
cnlangzi, line 21: It might have `nil pointer dereference` crash too.  do you think it is better to add 
```
if awp == nil {
return nil, errors.new("no write pool instance")
}
```

cnlangzi:
@Sriep  do you think it is better to add `nil pointer dereference` checking in code?

cnlangzi (commented):
@Sriep  do you think it is better to add `nil pointer dereference` checking in code?

code/go/0chain.net/smartcontract/storagesc/allocation.go
Sriep: Adding comments for pre-existing code is out of the scope. Plus this is an internal function so should not need comments for godoc, but only to help confused developers, I would expect only the endpoint methods linked in the `sc.go` files would need a godoc explanation.

code/go/0chain.net/smartcontract/storagesc/stakepool.go
Sriep: It is impossible for alloc to nil. I could put a `panic` in, but if alloc is `nil` there is an earlier coding error that `must` be fixed. It should `never` crash in production.

code/go/0chain.net/smartcontract/storagesc/models.go
Sriep: As before, if `sa` is nil then there is a serious coding error elsewhere and the chain is going to crash in the next few nanoseconds anyway.

code/go/0chain.net/smartcontract/storagesc/allocation_pool.go
Sriep: As before, if `sa` is nil then there is a serious coding error elsewhere and the chain is going to crash in the next few nanoseconds anyway.

code/go/0chain.net/smartcontract/storagesc/allocation_pool.go
Sriep: tks

code/go/0chain.net/smartcontract/storagesc/allocation_write_pools.go
Sriep: This is always called after `awp, err := getAllocationPools` has not returned an error. This can never be nil in production. If it is ever nil there is a coding error that `must` be fixed.

moldis (approved):

PR #366, opened 2021-06-20
Add curator and transfer allocatin ownership

Sriep:
Part of issue https://github.com/0chain/0chain/issues/333

Adds a curator to a list of allocation's curators. A curator can transfer allocations owners.
Adds transaction for a curator to trasnfer ownership.

Requires changes to gosdk  https://github.com/0chain/gosdk/pull/121 and zboxcli https://github.com/0chain/zboxcli/pull/47.

PR #402, opened 2021-07-14
Fix curator transfer allocation not changing client allocation list

Sriep:
https://github.com/0chain/0chain/pull/366 failed to alter the clientAllocation lists for the new and old owner. This PR removes the id from the old owner's list and adds it to the new owner's list.


moldis (approved):
Looks good!

PR #403, opened 2021-07-14
Fund allocation pools for someone else

Sriep:
Add ability to fund allocation pools for someone else, and add to their read or write pool.
1. New fundedPools object to hold the list of allocation pools a user has funded. Plus unit test.
2. Modified writePoolLock and readPoolLock to remember the funder of the new allocation pool.
3. Check that the client funded the pool being unlocked in writePoolUnlock and readPoolUnlock.
4. Fix existing unit tests to work with this new feature.


moldis:
Looks cool. With this PR I will be able to lock tokens from my 0box server for other wallets? As of now I have a roadblock - chain not let me lock them as 0box wallet is not the owner of allocation

PR #407, opened 2021-07-17
Fix the issue of usage different BLS & MCL libraries among 0Chain repos

platsko:
Bug description:
```
  `wrapReadRandCgo':
  /go/pkg/mod/github.com/herumi/bls-go-binary/bls/callback.go:7: multiple definition of `wrapReadRandCgo';
  /go/pkg/mod/github.com/herumi/bls/ffi/go/bls/callback.go:7: first defined here

```
- change imports to use "github.com/herumi/bls-go-binary/bls"
- sync Dockerfiles
- update build md file


mallochine:
Can you run this unit test: `TestMiraclToHerumiPK`

Make sure the output of `pk1` is sane. Do `fmt.Println("pk1", pk1)`. Paste the output here.

I'm asking because the last time someone changed the bls library version, it broke this unit test.

platsko:
We need another way to solve this problem

PR #411, opened 2021-07-19
Upgrade versions of 0Chain external modules

platsko:
- upgrade Golang and Alpine versions to golang:1.16-alpine3.12
- upgrade Rocks DB version 6.15.5 - latest compatible with package github.com/0chain/gorocksdb
- upgrade MCL & BLS go modules to latest versions
- update builders Docker and info Markdown files
- minor improvement for Makefile

PS the reason described in [this PR](https://github.com/0chain/0chain/pull/407)


platsko:
Run this unit test: TestMiraclToHerumiPK:
```
GOROOT=null #gosetup
GOPATH=/home/plats/go-path #gosetup
/usr/local/go/bin/go test -c -tags bn256 -o /mnt/c/Users/plats/AppData/Local/Temp/intellij-target.2/TestMiraclToHerumiPK_in_0chain_net_core_encryption_linux.test 0chain.net/core/encryption #gosetup
/usr/local/go/bin/go tool test2json -t /mnt/c/Users/plats/AppData/Local/Temp/intellij-target.2/TestMiraclToHerumiPK_in_0chain_net_core_encryption_linux.test -test.v -test.paniconexit0 -test.run ^\QTestMiraclToHerumiPK\E$ #gosetup
=== RUN   TestMiraclToHerumiPK
    bls0chain_test.go:19: miracl pk: 0418a02c6bd223ae0dfda1d2f9a3c81726ab436ce5e9d17c531ff0a385a13a0b491bdfed3a85690775ee35c61678957aaba7b1a1899438829f1dc94248d87ed36817f6dfafec19bfa87bf791a4d694f43fec227ae6f5a867490e30328cac05eaff039ac7dfc3364e851ebd2631ea6f1685609fc66d50223cc696cb59ff2fee47ac
    bls0chain_test.go:20: herumi pk: 68d37ed84842c91d9f82389489a1b1a7ab7a957816c635ee750769853aeddf1b490b3aa185a3f01f537cd1e9e56c43ab2617c8a3f9d2a1fd0dae23d26b2ca018
--- PASS: TestMiraclToHerumiPK (0.00s)
PASS

Process finished with the exit code 0
```

mallochine:
ok nice that unit test output looks correct.

Can you now just add assert to TestMiraclToHerumiPK that the "pk1" return value is equal to those values you printed out.

The reason why I'm asking you add assert is the last time someone changed the BLS library version, that BLS method was silently failing. That is, the function wouldn't return or throw error, but the output was still all 0s "000000..." instead of the hex string you've shown there.

By adding it as an assert, we make sure there's no regression there.

platsko:
> ok nice that unit test output looks correct.
> 
> Can you now just add assert to TestMiraclToHerumiPK that the "pk1" return value is equal to those values you printed out.
> 
> The reason why I'm asking you add assert is the last time someone changed the BLS library version, that BLS method was silently failing. That is, the function wouldn't return or throw error, but the output was still all 0s "000000..." instead of the hex string you've shown there.
> 
> By adding it as an assert, we make sure there's no regression there.

This is the test you are talking about it works well - the result in my comment above.

```
func TestMiraclToHerumiPK(t *testing.T) {
	miraclpk1 := `0418a02c6bd223ae0dfda1d2f9a3c81726ab436ce5e9d17c531ff0a385a13a0b491bdfed3a85690775ee35c61678957aaba7b1a1899438829f1dc94248d87ed36817f6dfafec19bfa87bf791a4d694f43fec227ae6f5a867490e30328cac05eaff039ac7dfc3364e851ebd2631ea6f1685609fc66d50223cc696cb59ff2fee47ac`
	pk1 := MiraclToHerumiPK(miraclpk1)

	require.EqualValues(t, pk1, "68d37ed84842c91d9f82389489a1b1a7ab7a957816c635ee750769853aeddf1b490b3aa185a3f01f537cd1e9e56c43ab2617c8a3f9d2a1fd0dae23d26b2ca018")

	// Assert DeserializeHexStr works on the output of MiraclToHerumiPK
	var pk bls.PublicKey
	err := pk.DeserializeHexStr(pk1)
	require.NoError(t, err)
}
```

mallochine:
ok LGTM then

Sriep (commented):
A welcome upgrade.

The explanation of setting up a build environment in https://github.com/0chain/0chain/blob/master/local/build_environment.md will now be wrong. It might be an idea to update `build_environment.md` it or tag it as outdated so new developers don't get confused.

platsko:
> A welcome upgrade.
> 
> The explanation of setting up a build environment in https://github.com/0chain/0chain/blob/master/local/build_environment.md will now be wrong. It might be an idea to update `build_environment.md` it or tag it as outdated so new developers don't get confused.

https://github.com/0chain/0chain/pull/411/files#diff-cee5ad2effc3fc5d9e62a54e0f7b3beece4d7b54dfb6afbab3cd73b8bde015c7 - do you mean this one?

Sriep (approved):
Ok looks great.

PR #420, opened 2021-07-23
Split Free storage payment between read and write pool.

Sriep:
Creates a new sc.yaml entry `read_pool_fraction`
On free storage allocation split the free funds between an allocation write pool and read pool in the ratio of `read_pool_fraction: 1-read_pool_fraction`.
Update unit test.


docker.local/config/0chain.yaml
```diff
@@ -2,7 +2,7 @@ version: 1.0
 
 logging:
   level: "debug"
-  console: false # printing log to console is only supported in development mode
+  console: true # printing log to console is only supported in development mode
```
moldis, line 5: I guess it's accidentally here

docker.local/config/0chain.yaml
```diff
@@ -117,10 +117,10 @@ server_chain:
     max_simultaneous_from_sharders: 30
 
 network:
-  magic_block_file: config/b0magicBlock_4_miners_2_sharders.json
+  magic_block_file: config/b0magicBlock_4_miners_1_sharder_piers.json
```
moldis, line 120: +1 accidentally added here

docker.local/config/0chain.yaml
```diff
@@ -117,10 +117,10 @@ server_chain:
...
   initial_states: config/initial_state.yaml
   genesis_dkg: 0
-  dns_url: "" # http://198.18.0.98:9091
+  dns_url: "" # http://198.18.0.98:9091 http://192.168.1.100:9091
```
moldis, line 123: +1 accidentally added here

docker.local/config/sc.yaml
```diff
@@ -151,10 +154,10 @@ smart_contracts:
     block_reward:
       block_reward: 1
       qualifying_stake: 1
-      sharder_ratio: 80
-      miner_ratio: 20
-      blobber_capacity_ratio: 20
-      blobber_usage_ratio: 80
+      sharder_ratio: 10
```
moldis, line 157: +1 accidentally added here

moldis (changes requested):
Minor cleanup

docker.local/config/sc.yaml
Sriep: No, actually that one is actually part of the PR. Change the block reward rations.

docker.local/config/0chain.yaml
```diff
@@ -117,7 +117,7 @@ server_chain:
     max_simultaneous_from_sharders: 30
 
 network:
-  magic_block_file: config/b0magicBlock_4_miners_2_sharders.json
+  magic_block_file: config/b0magicBlock_4_miners_2_sharder_piers.json
```
moldis, line 120: I think this still wrong as pointing to ...piers.json

PR #433, opened 2021-07-27
log Origin update in UpdateChanges()

rrrooommmaaa:
This PR writes warning to log when suspicious Origin update takes place

### DIFF
--- a/code/go/0chain.net/core/util/mpt_node_change.go
+++ b/code/go/0chain.net/core/util/mpt_node_change.go
@@ -6,6 +6,9 @@ import (
 	"fmt"
 	"io"
 	"sync"
+
+	"0chain.net/core/logging"
+	"go.uber.org/zap"
 )
 
 /*NodeChange - track a change to the node */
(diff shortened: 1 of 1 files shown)


code/go/0chain.net/core/util/mpt_node_change.go
```diff
@@ -122,8 +125,21 @@ func (cc *ChangeCollector) UpdateChanges(ndb NodeDB, origin Sequence, includeDel
 	nodes := make([]Node, len(cc.Changes))
 	idx := 0
 	for _, c := range cc.Changes {
-		c.New.SetOrigin(origin)
-		keys[idx] = c.New.GetHashBytes()
+		if _, ok := c.New.(*LeafNode); ok && origin != c.New.GetOrigin() {
```
peterlimg, line 128: Updating the origin would not break the references, so this logging code might not be needed.

rrrooommmaaa: Origin of leaf nodes is included in hash bytes -- why wouldn't it break references?

PR #436, opened 2021-07-29
WIP Update block rewards

Sriep:
This updates the block rewards smart contract to be more efficient; reduce the blockchain data access during the block rewards smart contract to a minimum.

It also adds a new endpoint \getBlobberStakeTotals to return a list of blobber stake totals. gosdk https://github.com/0chain/gosdk/pull/161 and zbox  https://github.com/0chain/zboxcli/pull/73 need this PR.

Fixes various broken unit tests, mainly due to changes to `stakepool.go`.

It creates two new objects to store on the blockchain:

`blobberStakeTotals` holds a list of blobbers with their total stakes. This is useful both in block reward calculations but also in the GUI for displaying blobber information. Hence also adding the \getBlobberStakeTotals endpoint.
```go
type blobberStakeTotals struct {
	Totals map[string]state.Balance `json:"totals"`
}

```

`blockRewardMints` holds pending mints for each blobber stake pool. Plus fields to allow a maximum check.
```go
type blockRewardMints struct {
	MintedRewards  float64 `json:"minted_rewards"`
	MaxMintRewards float64 `json:"max_mint_rewards"`
	// the max mint check is made before adding to UnMintedBlockRewards map
	UnProcessedMints map[string]float64 `json:"un_processed_mints"`
}
```

## Block rewards smart contract

image

The checks against MaxMint and MaxMintRewards are done in the block rewards smart contract.
The ratio of stakes in a blobber can't change unless `getStakePool` is called, at which point pending mints for that blobber are processed before the shareholder's ratios can be changed.

## Changes to `getStakePool`

image

A local copy of the stake total on loading is kept, to compare against on saving,
Only the mints for this particular blobber are processed.

## Changes to `stakePool.save`

image

Updates the total stakes for a particular blobber if that total has changed.


Sriep:
We are going to use a different method as this one tasks too long, given 1,000,000 blobbers.

PR #438, opened 2021-07-31
better thread safety

rrrooommmaaa:


Sriep (approved):

Sriep (changes requested):
When I tested this I got a stuck or slowly moving chain.

Sriep (approved):

PR #439, opened 2021-08-01
Test correct handling of MultiPutNode during pruning

rrrooommmaaa:
This failing test shows that error in UpdateVersion() may lead to deletion of nodes


Sriep:
When I tested locally the chain kept getting stuck before round 20. Suggest should be manually integration tested again.

Sriep (changes requested):
When I tested locally the chain kept getting stuck before round 20.

rrrooommmaaa:
I have merged changes from `master`, now the chain is moving fast.
Please merge.

Sriep (approved):

PR #455, opened 2021-08-06
Update magic block files

Sriep:
Some magic block files are out of date.


guruhubb:
You can also use 2 miners, but need to change min_n to 2 in sc.yaml and change n, but keep t, k the same for the magic block file.

cnlangzi (approved):
LGTM

anish-squareops (approved):

PR #458, opened 2021-08-08
Fix sharder panic on start

peterlimg:
Fix issue reported in https://0chain.slack.com/archives/C01V27E2Z9N/p1627278798076400.

Sharders would panic on start if they try to load Latest finalized block from others before running the `initN2NHandlers()`.


cnlangzi (approved):
LGTM

PR #470, opened 2021-08-10
Remove curator

Sriep:
Add smart contract endpoint to remove a curator from an allocation.
Created a new curator.go file and moved addCurator into it, plus moved unit test into no curator_test.go file.

Needed for https://github.com/0chain/zboxcli/pull/83 and https://github.com/0chain/gosdk/pull/201.


cnlangzi (approved):
LGTM

PR #492, opened 2021-08-18
Added MPT tests for multiple concurrent inserts and concurrent merges

rrrooommmaaa:


code/go/0chain.net/chaincore/block/entity.go
```diff
@@ -910,7 +909,7 @@ func (b *Block) ApplyBlockStateChange(bsc *StateChange, c Chainer) error {
 	//c.stateMutex.Lock()
 	//defer c.stateMutex.Unlock()
 
-	err := b.ClientState.MergeDB(bsc.GetNodeDB(), bsc.GetRoot().GetHashBytes())
+	err := b.ClientState.MergeChanges(bsc.GetRoot().GetHashBytes(), bsc.GetChanges(), nil, bsc.StartRoot)
```
peterlimg, line 912: This might cause problem by replacing the `MergeDB` with `MergeChanges`, as the `MergeDB` would only insert new nodes into db, no nodes would be deleted. The `MergeChanges` would delete the old nodes  in state changes. We will have prune goroutine to delete those old nodes, before that we'd better just keep it. 

Another problem is the `bsc.StartRoot`, usually a block's state start root is it's previous block's state root hash, and if the `bsc.StartRoot` is the state root of previous block, then we will have problem here. Check the code above 
```go
	if b.ClientState == nil {
		b.CreateState(c.GetStateDB(), root.GetHashBytes())
	}
```
If the block's state is nil, it will try to create a state with code above, and you will see the block's state root will be set to the root of the block's own state (not previous blocks). Apparently, the bsc.StartRoot is the previous block's state root, they are different, MergeChanges() will return error then.

What the `ApplyBlockStateChange` function do is get state changes from network, create the block's client state with the pnodeDB and the root calcuated from the state changes(acquired from network), then insert the changes to db directly. We can not do the startRoot check as the root could be either the previous block's state root and the block's own state root.

rrrooommmaaa: > The MergeChanges would delete the old nodes in state changes. 

`MergeChanges() deletes only nodes specified in the 3rd argument `deletes []Node` which is nil, So no deletes will take place, they are not even transmitted over the network.

> Apparently, the bsc.StartRoot is the previous block's state root, they are different, MergeChanges() will return error then.

Ok, we need to properly handle this error somehow. My point is -- we can't just apply node changes (at least in the form of []NodeChange to an arbitrary MPT -- this causes broken references (node not found during iteration) -- proven by tests). Why do we need this "block's state root will be set to the root of the block's own state (not previous blocks)"?
Didn't you propose to load all missing previous blocks and process them from oldest to newest?
Why is state change tranferred between the nodes in the form of []NodeChange? Why not pass an array of (path, value) so we can simply insert them into our MPT?

rrrooommmaaa: I suggest to deprecate block.SetStateDB() function.
Can we take the state from the previous block only? The previous block isn't supposed to be modifiable?

peterlimg: We cannot guarantee that we always have the previous one, the situation is when a new miner join the network, it will try to fetch the previous blocks from the latest finalize block, but it would not sync the whole chain, instead it would only fetch a given number of blocks (e.g 50), the oldest block would definitely has no previous block. In this situation, we will calculate its state base on the state db and the root calculated base on the partial state changes. This situation is normal and would happens often.  Please do not deprecate block.SetStateDB(), as it is important for the fix of 'node not found' error.

peterlimg: > `MergeChanges() deletes only nodes specified in the 3rd argument `deletes []Node` which is nil, So no deletes will take place, they are not even transmitted over the network.

The first argument of `MPT.insertNode()` is the old node that will be deleted if it's not nil.

peterlimg: > Ok, we need to properly handle this error somehow. My point is -- we can't just apply node changes (at least in the form of []NodeChange to an arbitrary MPT -- this causes broken references (node not found during iteration) -- proven by tests). Why do we need this "block's state root will be set to the root of the block's own state (not previous blocks)"?

The 'node not found' error is generally caused by the incorrect blocks sync process, the MPT itself works fine. Anyway, it should has been fixed in `fix/node-state-synch-err` branch, it runs pretty well in my local env, but have some issues when deployed to k8s. Feel free to have a test, and let me know if it does not work.

rrrooommmaaa: > The first argument of MPT.insertNode() is the old node that will be deleted if it's not nil.

Well, `oldNode` isn't transferred over the network, it's always nil. BTW, why do we have this `oldNode`? Wouldn't it be simpler to store the oldNode in `Deletes` field of the ChangeCollector?

> The 'node not found' error is generally caused by the incorrect blocks sync process, the MPT itself works fine. 

Can you create a unit test that fails on the current `master` or `staging` and doesn't fail on your branch?

rrrooommmaaa: > it will try to fetch the previous blocks from the latest finalized block, but it would not sync the whole chain, instead it would only fetch a given number of blocks (e.g 50), the oldest block would definitely has no previous block. In this situation, we will calculate its state base on the state db and the root calculated base on the partial state changes. 

Are we fetching previous blocks with FBRequest? Partial state changes from this block onwards?

peterlimg: Its `MinerNotarizedBlockRequestor` I think. 

> Partial state changes from this block onwards?
No, just the block's own state changes.

peterlimg: > Well, oldNode isn't transferred over the network, it's always nil

Oh, yes, the oldNodes were not transferred, but we'd better pass `nil` as oldNode explicitly to avoid unexpected behavior.

peterlimg: > Can you create a unit test that fails on the current master or staging and doesn't fail on your branch?
Sure, I'll do it.

rrrooommmaaa: > or maybe the logic can be changed to it must be equal to the start root acquired from network or the state root of the block's own

Do sharders keep ClientStateHash of each block?

PR #494, opened 2021-08-19
Minersc update config

Sriep:
Required for https://github.com/0chain/gosdk/pull/215 and https://github.com/0chain/zwalletcli/pull/42


code/go/0chain.net/smartcontract/minersc/sc.go
```diff
@@ -182,70 +182,26 @@ func getHostnameAndPort(burl string) (string, int, error) {
...
-	}
-
-	if err == nil {
-		if err = gn.Decode(p.Encode()); err != nil {
-			return nil, fmt.Errorf("%w: %s", common.ErrDecoding, err)
+	if err != nil {
+		if err != util.ErrValueNotPresent {
+			return nil, err
+		}
+		gn.readConfig()
```
peterlimg, line 195: Perhaps, we should return error here instead of using values from local yaml file.

PR #503, opened 2021-08-27
Benchmark storage smartcontracts

Sriep:
Benchmark 0chain smart-contract endpoints, check the [readme.md](https://github.com/0chain/0chain/blob/bench-sc/code/go/0chain.net/smartcontract/benchmark/main/readme.md) for further details.

The MPT database created is designed to create a database that can be used for the benchmarks rather than running a chain. So information related to running a chain will often be missing from the simulated MPT.

This PR does the following.

1. Creates a new application for running benchmarks.
2. In each smart contract, adds new files containing lists of benchmark tests for each endpoint. 
3. Adds files for populating the applications MPT.  This needs to access unexported objects in the various smart contracts.
4. Some modifications to endpoints in `storagesc` to make their signature uniform.
5. Access the signature scheme through the `state.StateContextI` interface rather than the chain. Smart contracts should be independent of the chain package.
6. Incorporates https://github.com/0chain/0chain/pull/507 as the read/write pool lock/unlock fix is needed for benchmarking those endpoints.
7. Some endpoints were skipped in anticipation of them changing in the set config PRs,  0chain/0chain#484, 0chain/0chain#485, 0chain/0chain#487, 0chain/0chain#489 and 0chain/0chain#494.


MurashovVen:
Can we add `*testing.B` argument to method `Run` of the `BenchTestI` interface like [this](https://github.com/0chain/0chain/blob/021314e13a606852b7b3098925a24d451a21cdd9/code/go/0chain.net/smartcontract/benchmark/benchmark.go#L203)? In `magmasc` we need to use the different already generated state for stress-cases (with 1k consumers, 100k providers, 1m active sessions, 10m inactive), because generating stress-state on-fly takes a lot of time. To solve this problem we just copy generated databases and use them. You can look at this [there](https://github.com/0chain/0chain/blob/e930331306dff0ee3b27d3e0e42cbf355a1eb1ac/code/go/0chain.net/smartcontract/magmasc/bench/execute_cases.go#L189). So, for clearly using benchmarks I need to stop timer by calling `b.StopTimer` inside `Run`.

Sriep:
Merge so can be used. Will open another Pr to handle any review issues.

code/go/0chain.net/smartcontract/benchmark/main/cmd/bls0chain.go
```diff
@@ -0,0 +1,274 @@
+package cmd
+
```
peterlimg, line 2: Any reason for why we don't use the `encryption/bls0chain.go` directly, and do a copy here.

PR #506, opened 2021-09-01
Fix/node state synch err

rrrooommmaaa:


code/go/0chain.net/chaincore/block/entity.go
```diff
@@ -763,7 +767,8 @@ func (b *Block) ComputeState(ctx context.Context, c Chainer) error {
 	if pb == b {
 		b.PrevBlock = nil // reset (a real case, may be unexpected)
 	}
-	if pb == nil {
+
+	if pb == nil || (pb != nil && !pb.IsStateComputed()) {
```
dabasov, line 771: can be simplified to pb == nill || !pb.IsStateComputed()

PR #509, opened 2021-09-04
Rrrooommmaaa/zcn test

rrrooommmaaa:
It turned out ChangeCollector was keeping references to values (Serializable), so nodes in the change collector with subsequent MergeMPTChanges could be mutated after saving them to MPT.


code/go/0chain.net/chaincore/block/entity.go
```diff
@@ -458,13 +458,15 @@ func (b *Block) HashBlock() {
...
-	for _, txn := range b.Txns {
-		b.TxnsMap[txn.Hash] = true
-	}
+	go func() {
+		defer b.mutexTxns.Unlock()
+		b.TxnsMap = make(map[string]bool, len(b.Txns))
+		for _, txn := range b.Txns {
+			b.TxnsMap[txn.Hash] = true
+		}
+	}()
```
peterlimg, line 469: Do not use the `b.mutextTxns`  in different gorountines. This could cause deadlock, when the mutex is locked, but before the mutex is unlocked in the goroutine, another thread could acquire the `RLock`, before it calls the `RUnlock`, if the goroutine ended and try to call the `deter b.mutexTxns.Unlock()`, we will get deadlock.

peterlimg (changes requested):

code/go/0chain.net/chaincore/block/entity.go
dabasov: I don't think such a deadlock is possible, since no goroutine will be allowed to aquire lock after ```b.mutexTxns.Lock()``` is called. Goroutine with ```defer b.mutexTxns.Unlock()``` will be guaranteed to execute eventually, so this lock will be unlocked. @peterlimg is right, this code has a problem: we should try make locks as short as possible (this mutext is more like monitor for TxnMap) but in the code above we can't guarantee than goroutine will be started immediately, so this lock will be locked extra time which is not necessary

rrrooommmaaa: Right, in Golang (unlike some other programming languages) a locked RWMutex is not associated with a particular goroutine. One goroutine may RLock (Lock) a RWMutex and then arrange for another goroutine to RUnlock (Unlock) it.
My main concern was that upon initialization of a new block (in `protocol_block_main.go` function `GenerateBlock` we fill the block with transactions `b.Txns = append(b.Txns, txn)` and then start map calculation in a goroutine.
As the goroutine start may be delayed, `Block.HasTransaction()` and `Block.Clone()` may incorrectly handle the empty map.
In this patch, they have to wait until map calculation has finished and that seems logical to me.
How would you suggest to handle this situation?

code/go/0chain.net/core/util/merkle_patricia_trie.go
```diff
@@ -129,14 +129,15 @@ func (mpt *MerklePatriciaTrie) Insert(path Path, value Serializable) (Key, error
 		return mpt.Delete(path)
 	}
 
+	valueCopy := &SecureSerializableValue{eval}
```
rrrooommmaaa, line 132: as `wallet_test` failed because it saved an updated `State` object multiple times into MPT, an as MemoryNodeDB actually stores original objects in its map, subsequent MPT Validation failed because some nodes hashes were recalculated on AddChanges into different values.
So the safest solution is to convert any incoming value to bytes (SecureSerializableValue) right on input to the MPT.

code/go/0chain.net/chaincore/block/entity.go
peterlimg: Oh, yes, the deadlock would not exist here, my mistake. Basically, once you acquired the write lock, all others has to wait for the lock to be unlocked, so we don't need to put the `b.TxnsMap` updating code in a new goroutine.

code/go/0chain.net/core/util/merkle_patricia_trie.go
peterlimg: LGTM 👍 . This is the minimal changes we can do to make it immutable, but in the future, we will stop doing the `encoding` in the `Insert()` method of MPT, the encoding should happen only when we save the data into persistent db (i.e. *pnode), all other types of nodes store value in RAM directly.

peterlimg (changes requested):

code/go/0chain.net/core/util/merkle_patricia_trie.go
dabasov: @peterlimg it will be extremely useful to extract event sequence (updates, deletes, inserts, reads) before MPT persistence to track changes caused by transaction execution. it is the cornerstone of parallelizing smart contract execution which I'm working on now. I will describe how it can help parallelization soon in details, may be you thought of addition such a functionality previously?

rrrooommmaaa: Changes are currently collected in ChangeCollector object (each MPT has it), if that's what you're asking about.
The question is how to merge changes from several collectors and see if they really conflict or not. Current implementation doesn't allow to do this.

dabasov: Yes, something similar to ChangeCollector with additional AccessCollector. We will need some mechanics which can tell us what parts of the state SC accessed and what changed to make read and write sets for further contention detection.

rrrooommmaaa: sub-trees, that is?

PR #519, opened 2021-09-12
Auth ticked signed by client

Sriep:
The auth ticket is signed in gosdk, and can only be signed by the client, as the owner's public key is not in general known, only if the owner is the client.

This issue causes all 3rd party paid downloads to be free.
related to https://github.com/0chain/gosdk/issues/246


Sriep:
Its not this its the hash closing

PR #525, opened 2021-09-14
Rrrooommmaaa/zcn test (#509)

rrrooommmaaa:
* removed unused function

* don't expose uncalculated Block.TxnsMap

* store values in binary form throughout MPT

* compute TxnMap in-thread


peterlimg:
Already merged into staging by #509

PR #526, opened 2021-09-16
Remove "0chain.net/chaincore/chain" dependancy from storagesc

Sriep:
storagesc should not require a running chain, so should not import `"0chain.net/chaincore/chain"`. The problem is that in some places `chain.GetServerChain().GetSignatureScheme()` was used to get the signature scheme rather than `balances.GetSignatureScheme()`. This fixes that.

Some upcoming PRs will need this change, either because there is no chain (benchmarks https://github.com/0chain/0chain/pull/503 & https://github.com/0chain/0chain/pull/515) or to stop cyclic dependencies.


Sriep:
Replaced with https://github.com/0chain/0chain/pull/527

PR #527, opened 2021-09-17
Remove "0chain.net/chaincore/chain" dependancy from storagesc

Sriep:
storagesc should not require a running chain, so should not import `"0chain.net/chaincore/chain"`. The problem is that in some places `chain.GetServerChain().GetSignatureScheme()` was used to get the signature scheme rather than `balances.GetSignatureScheme()`. This fixes that.

Some upcoming PRs will need this change, either because there is no chain (benchmarks https://github.com/0chain/0chain/pull/503 & https://github.com/0chain/0chain/pull/515) or to stop cyclic dependencies https://github.com/0chain/0chain/pull/528.

### DIFF
--- a/code/go/0chain.net/smartcontract/storagesc/models.go
+++ b/code/go/0chain.net/smartcontract/storagesc/models.go
@@ -11,7 +11,7 @@ import (
 
 	chainstate "0chain.net/chaincore/chain/state"
 
-	"0chain.net/chaincore/chain"
+	//"0chain.net/chaincore/chain"
 	"0chain.net/chaincore/node"
 	"0chain.net/chaincore/state"
 	"0chain.net/core/common"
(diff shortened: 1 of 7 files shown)


code/go/0chain.net/smartcontract/storagesc/models.go
```diff
@@ -11,7 +11,7 @@ import (
 
 	chainstate "0chain.net/chaincore/chain/state"
 
-	"0chain.net/chaincore/chain"
+	//"0chain.net/chaincore/chain"
```
peterlimg: Remove the commented code if it's not being used

peterlimg (changes requested):
All looks good to me, only one line commented code need to be deleted.

PR #528, opened 2021-09-17
WIP Block rewards

Sriep:
build off https://github.com/0chain/0chain/pull/527

Update the total blobber capacity and used fields each time one of them changes. With planned transaction parallelisation changes hopefully, we will only need to update the MPT once per round.
image

Each time a smart contract is about to change one of the parameters used to calculate a blobber's stakeholders
block rewards, we calculate the blobber rewards and pay to the stakeholder.
image

todo


Sriep:
Depreciated

Sriep:
Closed as this solution is linear in block since last reward calculation, and takes too long.

PR #529, opened 2021-09-17
Migrate miner handler to grpc

niki290595:


code/go/0chain.net/miner/handler/grpc_middleware.go
```diff
@@ -0,0 +1,58 @@
...
+		return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+			if wrappedServer.IsGrpcWebRequest(r) {
+				wrappedServer.ServeHTTP(w, r)
+				return
+			}
+			h.ServeHTTP(w, r)
+		})
+	})
+
+	return srv
```
kushthedude: We would want to avoid this boilerplate rather we will follow the one official listed in their project

Issue #415, opened 2021-07-22
Bad stringified json format on transaction_output field from TransactionConfirmationHandler


shaktals:
As per [this issue](https://github.com/0chain/js-client-sdk/issues/35) opened on `js-client-sdk`, the _transaction_output_ field received from 0chain's [TransactionConfirmationHandler](https://github.com/0chain/0chain/blob/master/code/go/0chain.net/sharder/handler_main.go) has a stringified _json_ content, but it is badly formatted. Here is what is coming through to `go-sdk` now:

![Selection_038](https://user-images.githubusercontent.com/24679343/126577281-525441ba-e87f-480d-98cd-4ea0781910a4.png)

And here is what we get from the `js-client-sdk`:

![Selection_037](https://user-images.githubusercontent.com/24679343/126577641-1ad08f70-53c4-4401-8d5a-7f74a5d62423.png)

If we are to send a collection of objects to be parsed as json, they should be put in a comma separated object array `"[{}, {}, {}]"`.

shaktals:
Bump :+1:

shaktals:
Actually, this should not have been closed. It has been worked around, but the issue is still there.
If someone can help me with the draft PR #531 I'll be happy to finish it and fix this.

ma2b0043:
@shaktals when can we expect to close this issue?

shaktals:
@ma2b0043 can you help me with #531?

ma2b0043:
> @ma2b0043 can you help me with #531?

isnt it closed @shaktals ?

shaktals:
> > @ma2b0043 can you help me with #531?
> 
> isnt it closed @shaktals ?

It's closed because I wasn't able to finish the PR.
The web frontend team needed lots of attention, so I have done basically nothing on the backend yet.
Feel free to pick that PR and finish it.

uneet7:
Hi, @shaktals , can you please post a detailed payload or a bit about the command that you are executing? I think I got the issue.

PR #531, opened 2021-09-21
Fix txn_output json from TransactionConfirmationHandler response 

shaktals:
I'm writing my first Go PR to try and fix this issue #415.

To fix it, I wanted to run unit tests for the handler, update the code and then run the tests again, to make sure it is fine.
As there were no unit test for the handler, I wrote one, copying from other tests the best I can, although I have a very shallow understanding of the whole 0chain code.

I have been able to fmt.Println through the code to create a block, create a test sharder chain and add the block to the chain. But when reaching [this line](https://github.com/0chain/0chain/blob/fix-txnconfirmation-resp-json/code/go/0chain.net/sharder/transaction.go#L41-L42), on the `sc.BlockTxnCache.Get(hash)` call, I get the following error:

`missing key: key not found`

What am I missing?

### DIFF
--- a/code/go/0chain.net/core/persistencestore/connection.go
+++ b/code/go/0chain.net/core/persistencestore/connection.go
@@ -2,6 +2,7 @@ package persistencestore
 
 import (
 	"context"
+	"fmt"
 	"os"
 	"sync"
 	"time"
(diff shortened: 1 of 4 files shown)


shaktals:
Since I'm hitting issues running test locally, I'll just close this and delegate the issue to the backend team.

PR #397, opened 2021-07-11
Zcnsc latest

Dmdv:
This adds a new smart contract for converting between ERC-20 ethereum tokens and ZCN. It adds six new endpoints defined in sc.go. Details in 0Chain Token Bridge document.

image

PR #536, opened 2021-09-22
Zcnsc latest

Dmdv:
This is a replay of https://github.com/0chain/0chain/pull/397 into staging branch.

This adds a new smart contract for converting between ERC-20 ethereum tokens and ZCN. It adds six new endpoints defined in sc.go. Details in 0Chain Token Bridge document.

image


code/go/0chain.net/chaincore/state/mint.go
```diff
@@ -9,14 +9,13 @@ import (
...
 
-//Transfer - a data structure to hold state transfer from one client to another
 type Mint struct {
 	Minter     datastore.Key `json:"minter"`
 	ToClientID datastore.Key `json:"to"`
 	Amount     Balance       `json:"amount"`
 }
 
-//NewTransfer - create a new transfer
+// NewMint NewTransfer - create a new transfer
```
Sriep: New mint does not create a new transfer

code/go/0chain.net/smartcontract/zcnsc/README.MD
```diff
@@ -0,0 +1,4 @@
+![Diagram](https://github.com/Dmdv/0chain/blob/zcnsc_latest/code/go/0chain.net/smartcontract/zcnsc/docs/ZCNSC.png?raw=true)
```
Sriep, line 1: Nice pictures, but maybe add some text to introduce them?

docker.local/bin/build.miners.sh
```diff
@@ -9,6 +9,8 @@ DOCKER_DIR="$ROOT/docker.local/build.miner"
 DOCKER_FILE="$DOCKER_DIR/Dockerfile"
 DOCKERCOMPOSE="$DOCKER_DIR/docker-compose.yml"
 
+APP_ROOT="$ROOT/code/go/0chain.net"
```
Sriep: Does not seem to be used.

code/go/0chain.net/smartcontract/zcnsc/models.go
```diff
@@ -0,0 +1,586 @@
...
+
+func (pk *AuthorizerParameter) Decode(input []byte) error {
+	err := json.Unmarshal(input, pk)
+	return err
+}
+
+type AuthorizerNode struct {
+	ID        string                    `json:"id"`
+	PublicKey string                    `json:"public_key"`
+	Staking   *tokenpool.ZcnLockingPool `json:"staking"`
```
dabasov, line 285: It is not a good idea to organize data as a tree without proper normalization to be honest. 
Every change to pool for example, 
```
transfer, resp, err = ans.NodeMap[tran.ClientID].Staking.EmptyPool(gn.ID, tran.ClientID, tran)
```
will cause changing of AuthorizerNodes. 
- it is serialization/deserialization of pretty large json
- it is contention on all list instead of small it's part

It'll be better to normalize storage and use Ids of all nested entities, fortunately all of them do have one. Instead of storing entity or list we will store id or list of ids, nested entities will be stored in separate MPT node as independent structure. 

This approach has several drawbacks though: 
 - more read queries to MPT which shouldn't be a problem
 - bigger read set and wider MPT
 
But the biggest benefit is that when we change state, we touch narrow part of it, since addition of authorizers won't happen too frequently, but manipulations on their state like pools will. We had similar discussion with @Sriep.  
 
 Do you think such a normalization makes sense?

Dmdv: In that case how we should address the same issue in **miners smart contract** and **interestpool smart contract** ?

https://github.com/0chain/0chain/blob/staging/code/go/0chain.net/smartcontract/minersc/models.go#L587

```golang
// MinerNode struct that holds information about the registering miner.
type MinerNode struct {
	*SimpleNode `json:"simple_miner"`
	Pending     map[string]*sci.DelegatePool `json:"pending,omitempty"`
	Active      map[string]*sci.DelegatePool `json:"active,omitempty"`
	Deleting    map[string]*sci.DelegatePool `json:"deleting,omitempty"`
}
```

dabasov: I definitely suggest addressing it the same way. @platsko made good attempt to change the way we store lists in magmasc, I do not like local db for lists tbh, but the direction is correct.

Probably, we should extract this refactoring to  another task and make separate PR for this changes in all smart contracts at once. @guruhubb what do you think?

guruhubb: I agree we can scale to big number as per https://0chain.slack.com/archives/C020E8HA8TE/p1630528421006600 results that @platsko got.  @peterlimg @Sriep thoughts?

Sriep: A good idea, I was horrified when I first saw [getMinersList](https://github.com/0chain/0chain/blob/staging/code/go/0chain.net/smartcontract/minersc/models.go#L1204). 

But where is the balance between extra MPT reads, and unmarshalling data unnecessarily? How large do objects need to get before we switch to a list of ids?

guruhubb: I think authorizers, miners, sharders, consumer lists will be small and fixed. The providers list is extensive (millions) and thats why @platsko chose that way.  Perhaps we can use this for blobbers.  But perhaps better to have one data structure for all.

An additional method would be to have them in groups, say group A to Z where they are randomly assigned when a blobber or provider sign ups.  The groups need can be ordered lists so that its manageable and scale logN for r/w ops.

Dmdv: @guruhubb @Sriep @dabasov I'm taking into account all suggestions and I will rework this smart contract according to magma sc

Dmdv: I will implement lists in this PR, though is seems already huge

Dmdv: @Sriep We discussed with @guruhubb that issue with MPT update.
I'm making refactoring regarding how we register authorisers but it won't be like with magmasc, there are some other issues under consideration. As far as MPT update is concerned, the issue with multiple updates and MPT queries doesn't affect much the performance so I decided to make refactoring in a separate branch, meanwhile adding benchmarks to the authorisers. That way will be easier to estimate the effect of refactoring

Dmdv: So we can merge the request now

Dmdv: @Sriep Waiting for your approval

Dmdv: @peterlimg Could you review please?

Issue #534, opened 2021-09-22
Smart contract settings: restrict changes to set rounds.


Sriep:
For each smart contract, restrict setting changes to occur on particular rounds.
Add a `setting_change_round` to each smart contract entry in sc.yaml. 
So setting changes are collected and changed all at once when the round is an even multiple of `setting_chane_round`.

ma2b0043:
@Sriep when can we expect to close this issue?

PR #537, opened 2021-09-23
Make storeg sc setting changes periodic

Sriep:
1. Add a new 0chain entry `setting_update_period`, and the corresponding chain config field.
2. Change the `update_settings` to store the proposed changes in the MPT, rather than enacting them.
3. Every `setting_update_period` rounds run a storagesc `commit_settings_changes` transations
4.  `commit_settings_changes` will commit the accrued setting changes since the last `commit_settings_changes` transaction to the MPT.
5. Update unit tests with changes.

issue https://github.com/0chain/0chain/issues/534


code/go/0chain.net/chaincore/config/config.go
```diff
@@ -39,12 +39,13 @@ func SetupDefaultConfig() {
 	viper.SetDefault("server_chain.block.sharding.min_active_sharders", 100)
 	viper.SetDefault("server_chain.block.sharding.min_active_replicators", 100)
 	viper.SetDefault("server_chain.smart_contract.timeout", 550*time.Millisecond)
+	viper.SetDefault("server_chain.smart_contract.setting_update_period", 300)
```
peterlimg: Perhaps use the same default value of `200` as 0chain.yaml has.

code/go/0chain.net/smartcontract/storagesc/config_settigns.go
```diff
@@ -548,18 +555,57 @@ func (ssc *StorageSmartContract) updateSettings(
...
+
+	return "", nil
+}
+
+func (ssc *StorageSmartContract) commitSettingChanges(
+	t *transaction.Transaction,
+	_ []byte,
+	balances chainState.StateContextI,
+) (resp string, err error) {
+	if t.ClientID != balances.GetBlock().MinerID {
```
peterlimg: This will stop the transaction being executed by other miners. If a miner recives a block that is generated by others, and the block contains this transaction, the miner would fail to execute it. Therefore, this block could never be verified by others and will not be accepted by the blockchain.

peterlimg (changes requested):

peterlimg (changes requested):
https://github.com/0chain/0chain/pull/537/files#r716397419

code/go/0chain.net/chaincore/config/config.go
Sriep: Perhaps, but I was thinking that maybe we could stagger them, so everything does not happen on the same block?

code/go/0chain.net/smartcontract/storagesc/config_settigns.go
Sriep: I was copying the check from the [payFees transaction](https://github.com/0chain/0chain/blob/staging/code/go/0chain.net/smartcontract/minersc/fees.go#L358). Is the situation different here or should this be fixed?

peterlimg: The payFees transaction will be created after executing all the transactions in a block, and in the payFees transaction, the `ClientID` will be set to `MinerID` as we will only pay fee to the miner who generated the block. So the `if t.ClientID != balances.GetBlock().MinerID` is a must here, but for other normal transactions, we should not do this.

code/go/0chain.net/chaincore/config/config.go
peterlimg: I mean the `setting_update_period` is different from the value of `setting_update_period` in the 0chain.yaml, I think this `300` will be set to the value in the `0chain.yaml` after loading the yaml file.

PR #547, opened 2021-10-03
Add Partitions package and league tables

Sriep:
Adds a new partitions package, and implements a league table to be used for block rewards.

League tables hold N divisions of M members order by a numeric field.
* `Add` adds a new member to the league.
* `Remove` removes a member
* `Change` changes the numeric field for an existing member

For a total number of T members, these operations are of order(SqareRoot(T)). This assumes the league is split into SquareRoot(T) divisions of maximum size SquareRoot(T).

Each time a member of the league table is moved a callback function is triggered. This can do stuff, like pay block rewards.

To see how it works check out the [fuzzy test](https://github.com/0chain/0chain/blob/49cc4533d1fe5571175e8bfb98de3504fb99ffdb/code/go/0chain.net/smartcontract/partitions/league_table_test.go#L131)


code/go/0chain.net/smartcontract/partitions/league_table.go
```diff
@@ -0,0 +1,489 @@
...
+type leagueMember struct {
+	Id    string `json:"id"`
+	Value int64  `json:"value"`
+}
+
+func (lm leagueMember) Name() string {
+	return lm.Id
+}
+
+func (lm leagueMember) GraterThan(item PartitionItem) bool {
```
peterlimg, line 26: Typo `GraterThan`

code/go/0chain.net/smartcontract/partitions/league_table.go
```diff
@@ -0,0 +1,489 @@
...
+	}
+
+	newDiv.changed = true
+	oldDiv.changed = true
+	return nil
+}
+
+// Removes and returns the first member
+// If there are not enough members trys to fill up the numbers
+// by promoting frmo a lower division
```
peterlimg, line 324: Typo `frmo`

code/go/0chain.net/smartcontract/partitions/league_table.go
```diff
@@ -0,0 +1,489 @@
...
+	}
+	return b
+}
+
+func (d *divison) Decode(b []byte) error {
+	return json.Unmarshal(b, d)
+}
+
+func (d *divison) set(index int, opi OrderedPartitionItem) error {
+	lm, ok := opi.(leagueMember)
```
peterlimg, line 70: If the `OrderedPartitionItem` should always be `leagueMember`, maybe we can just pass the param in type of `leagueMember`.

code/go/0chain.net/smartcontract/partitions/league_table.go
```diff
@@ -0,0 +1,489 @@
...
+	Id    string `json:"id"`
+	Value int64  `json:"value"`
+}
+
+func (lm leagueMember) Name() string {
+	return lm.Id
+}
+
+func (lm leagueMember) GraterThan(item PartitionItem) bool {
+	lm2, ok := item.(leagueMember)
```
peterlimg, line 27: Add a method to `PartionItem` to get value and compare with the `leagueMember`'s value should be better, or we can just pass the `leagureMember` type of `item` instead of an interface.

code/go/0chain.net/smartcontract/partitions/league_table.go
```diff
@@ -0,0 +1,489 @@
...
+		return err
+	}
+	lt.Divisions[targetDivision].changed = true
+	if lt.Callback != nil {
+		if err := lt.Callback(in, NoPartition, PartitionId(targetDivision), balances); err != nil {
+			return fmt.Errorf("running callback, {in: %v, old position: %v, new poslitin: %v}",
+				in, nil, targetDivision)
+		}
+	}
+	lt.Divisions[targetDivision].changed = true
```
peterlimg, line 175: Dup `changed` updating.

code/go/0chain.net/smartcontract/partitions/league_table.go
```diff
@@ -0,0 +1,489 @@
...
+	if err != nil {
+		return err
+	}
+	if div == nil {
+		return fmt.Errorf("partition %v not found", index)
+	}
+
+	position := div.find(name)
+	if position == notFound {
+		return err
```
peterlimg, line 197: The error is nil, return a specific error or just nil maybe

code/go/0chain.net/smartcontract/partitions/league_table.go
```diff
@@ -0,0 +1,489 @@
...
+	}
+	if oldDiv == nil {
+		return fmt.Errorf("partition %v not found", from)
+	}
+
+	// move from (from, oldPosition)
+	oldPosition := oldDiv.find(changed.Name())
+	if oldPosition == notFound {
+		return err
```
peterlimg, line 250: nil error

code/go/0chain.net/smartcontract/partitions/league_table.go
```diff
@@ -0,0 +1,489 @@
...
+	}
+}
+
+func (lt *leagueTable) getDivision(i int, balances state.StateContextI) (*divison, error) {
+	if i == len(lt.Divisions) {
+		return nil, nil
+	}
+	if i > len(lt.Divisions) {
+		return nil, fmt.Errorf("partition id %v grater than numbr of partitions %v", i, len(lt.Divisions))
+	}
```
peterlimg, line 429: `i >= len(lt.Divisions)` is actually the same thing, index out of boundary.

peterlimg (changes requested):
Also, please comment all exported public methods and functions.

code/go/0chain.net/smartcontract/partitions/league_table.go
Sriep: Yes but i don't want `i == len(lt.Divisions)` to be an error. So I just return a nil for the value. Maybe I should create a new division and return that.

code/go/0chain.net/smartcontract/partitions/league_table.go
Sriep: I would then have to commit to a type, int, int64, state.Balance, float64 etc. Leave open for the moment.

Sriep:
Closing this as replaced with https://github.com/0chain/0chain/pull/588.

Issue #550, opened 2021-10-05
Create boilerplate code for the grpc-gateway


kushthedude:
#### Description
The boilerplate code being used by blobber is too difficult to navigate through or to understand, utilise the standard grpc-gateway boilerplate code https://github.com/johanbrandhorst/grpc-gateway-boilerplate

PR #567, opened 2021-10-06
Add GRPC gateway boilerplate

amanjain97:
## WHAT
- Add proto files 
- Add grpc gateway from boilerplate
- Create grpc server 
- Add buf tool for proto lint and generate. 

Issue: https://github.com/0chain/0chain/issues/550

Todo:

- [ ] Add rate limiter from core/common/ratelimiter.go
- [ ]


buf.lock
```diff
@@ -0,0 +1,10 @@
+# Generated by buf. DO NOT EDIT.
+version: v1
+deps:
+  - remote: buf.build
+    owner: beta
+    repository: googleapis
+    branch: main
+    commit: ac0d624de3f746c3b8ef55c74b6f5244
+    digest: b1-GubXb4hZ56wTohZYDL-Muute9Jxx8J4TtIYucS30TBM=
+    create_time: 2021-08-12T19:53:55.128813Z
```
kushthedude, line 10: Lets omit this for now as this protos would be changing a lot initially, What do you think?

code/go/0chain.net/miner/gateway/gateway.go
```diff
@@ -0,0 +1,59 @@
...
+
+	port := os.Getenv("PORT")
+	if port == "" {
+		port = "11000"
+	}
+	gatewayAddr := "0.0.0.0:" + port
+	gwServer := &http.Server{
+		Addr: gatewayAddr,
+		Handler: http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+			if strings.HasPrefix(r.URL.Path, "/api") {
```
kushthedude: Do we want the `api` suffix over here? Not sure if handlers have that suffix.

code/go/0chain.net/miner/m_handler.go
```diff
@@ -340,13 +340,20 @@ func NotarizedBlockHandler(ctx context.Context, entity datastore.Entity) (
 
 // NotarizedBlockSendHandler - handles a request for a notarized block.
 func NotarizedBlockSendHandler(ctx context.Context, r *http.Request) (interface{}, error) {
-	return getNotarizedBlock(ctx, r)
+	roundN, hash, err := getBlockInfoFromReq(r)
+	if err != nil {
+		return nil, err
+	}
+	return GetNotarizedBlock(ctx, roundN, hash)
```
kushthedude, line 347: What is this code change around? I can still see http.Request in the signature 🤔, is it just a refactor of the current http logic

code/go/0chain.net/miner/miner/main.go
```diff
@@ -421,3 +425,32 @@ func initWorkers(ctx context.Context) {
 	//miner.SetupWorkers(ctx)
 	transaction.SetupWorkers(ctx)
 }
+
+func initGRPCServer(grpcPort int) {
+	grpcServer := server.NewGRPCServerWithMiddlewares()
+
+	if config.Development() {
```
kushthedude: Lets not add this for now as we are not pushing the changes on master

buf.lock
amanjain97: This is the google proto library which I don't think would be changing. But it is my first time using `buf` tool, so, we can try it out.

code/go/0chain.net/miner/gateway/gateway.go
amanjain97: No, we donot need it.

code/go/0chain.net/miner/m_handler.go
amanjain97: Yes, I am refactoring this endpoint and will push the updated code soon for this. I will work on it in https://github.com/0chain/0chain/issues/552
Should I remove the logic from this PR.

code/go/0chain.net/miner/miner/main.go
amanjain97: Ok

amanjain97:
@twiny  Can you check the buf validate command, it is failing.

twiny:
@amanjain97 I think this is because of `bufbuild/buf-breaking-action@v1` trying to check breaking changes against this branch: `https://github.com/0chain/0chain.git#branch=grpc_miner` which does not have the .proto files yet.

https://github.com/0chain/0chain/pull/567/checks?check_run_id=3846251685#step:4:3

PR #574, opened 2021-10-07
grpc migration v2 - adding github workflow jobs

twiny:
added github workflow jobs to validate grpc .proto files and build buf.


.github/workflows/validate.yml
```diff
@@ -0,0 +1,17 @@
...
+on: [pull_request]
+
+jobs:
+  validate-grpc-stubs:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: bufbuild/buf-setup-action@v0.5.0
+      - uses: bufbuild/buf-lint-action@v1
+      - uses: bufbuild/buf-breaking-action@v1
```
kushthedude: This wont work until we have some proto

.github/workflows/cicd.yml
```diff
@@ -85,4 +85,28 @@ jobs:
...
+        uses: actions/checkout@v1
+      # There is no need to do this if you are not installing from source
+      - name: setup
+        if: success()
+        uses: actions/setup-go@v1
+        with:
+          go-version: 1.16.6
+      - name: make local
+        if: success()
+        run: make local
```
kushthedude, line 111: Why is it make local? Have you added the make target paths 🤔 ?

twiny: thats correct - I did not add the Make target paths.. let me fix that

PR #576, opened 2021-10-08
Benchmark control tests

Sriep:
This adds some simple control tests to the benchmark program. One gets an array of size `contrl_m`  from the MPT, the other gets `contorl_n` int64 objects from the MPT.

Added a benchmark for challenge_resopnse, which was earlier omitted.


Sriep:
https://github.com/0chain/system_test/actions/runs/1333641220 system test

PR #585, opened 2021-10-12
GRPC miner migration v2 `/_sign` Handler

twiny:


code/go/0chain.net/miner/server/grpc/handlers/sign.go
```diff
@@ -0,0 +1,40 @@
...
+	"0chain.net/core/encryption"
+	minerproto "0chain.net/miner/proto/api/src/proto"
+)
+
+// Sign
+func (m *minerGRPCService) Sign(ctx context.Context, req *minerproto.SignRequest) (*minerproto.SignResponse, error) {
+	privateKey := req.PrivateKey
+	publicKey := req.PublicKey
+	data := req.Data
+	timestamp := req.TimeStamp
```
kushthedude: They are unnecessary assignment we can directly use `req.*` variable whereever need be, HTTP Handler has such because it has to extract from the http.Request

code/go/0chain.net/miner/server/grpc/handlers/sign.go
```diff
@@ -0,0 +1,40 @@
...
+	key, err := hex.DecodeString(publicKey)
+	if err != nil {
+		return nil, err
```
kushthedude: use error wrapping 
```suggestion
		return nil, errors.New(err, "failed to decode string")
```

code/go/0chain.net/miner/server/grpc/handlers/sign.go
```diff
@@ -0,0 +1,40 @@
...
+	var hashdata string
+	if timestamp != "" {
+		hashdata = fmt.Sprintf("%v:%v:%v", clientID, timestamp, data)
+	} else {
+		hashdata = fmt.Sprintf("%v:%v", clientID, data)
+	}
+	hash := encryption.Hash(hashdata)
+	signature, err := encryption.Sign(privateKey, hash)
+	if err != nil {
+		return nil, err
```
kushthedude: Use error wrapping

kushthedude (commented):
Tests are missing for the PR, would you mind adding some unit tests or integration tests for the handler?

amanjain97 (approved):

PR #588, opened 2021-10-15
generate_challenges 

### LINKED ISSUE #578: GenerateChallenges - improve performance
[GenerateChallenges](https://github.com/0chain/0chain/blob/staging/code/go/0chain.net/smartcontract/storagesc/challenge.go#L524) needs to be improved. 

The biggest issue is the BlobberChallenge object https://github.com/0chain/0chain/issues/579, this takes too long to get from the MPT and needs refactoring. Fixing this should reveal several other issues below.

One obvious problem point is that it makes use of the list of all allocations. As there could be upwards of ten million allocations this is not a practical solution, so another way must be found.

We need to devise a way of using partitions to select a random allocation.

GenerateChallenges also uses a list of all validators. This list also will get too long, so another way will need to be found.

GenerateChallenges is linear in `max_challenges_per_generation` sc.yaml entry. But we can assume that this does not increase above 100.

Sriep:
This PR improves the efficiency of the  `generate_challenges` endpoint.

1. Create partitions, that allow random selection from a list that can support millions of entries.
2. Replaces the list of all allocations and all validators with partitions, so as to allow millions of objects.
3. Refactor validators. Recognises that only the list in 2 is used for validator information.
4. Run the main for loop in parallel using a `wait group`.
5. Take the stat update out of the `for loop`.
6. Add a hardcoded maximum to max challenges per generation. Set to ten initially.
7. Some refactoring of the code

Still to do.
Refactor the [BlobberChallenge](https://github.com/0chain/0chain/blob/staging/code/go/0chain.net/smartcontract/storagesc/models.go#L93). The challenge array 
```go
Challenges               []*StorageChallenge          `json:"challenges"`
```
will ultimately prove too big. Splitting up by allocation is one possibility. However this array is passed to the blobber, so any changes will need to be coordinated with the blobber.

## Partitions
The idea of partitions is to split a large slice that takes too long to manipulate, into sections that are easier to handle. So split a million objects into ten thousand partitions of a hundred objects.

The list of all allocations and all validators is held for the sole purpose of random selection.
1. To get random allocations to make challenges for.
2. To get random validators to assign to a challenge.

So the order of times is not important, we just need to add, remove and randomly select items. So assume 10,000 partitions of 100 items. The last partition in the list can, and in general, will have less than 100 items all others exactly 100.
- Add: We select a partition at random. Remove the last item and move it to the end of the last partition. Insert the added item into the vacated position. Return position for the object to know what partition it is held in.
- Remove. Enter the partition and id of the object. Remove object form partition and replace with the last item in the last partition.
- Whenever we change the location of an item, a callback is called. The callback is used to inform the moving object of its new partition.  Add and Remove can have a tax of an extra MPT update to inform an object (allocation) of its new partition.
- Random select, we randomly select a partition and a sub slice of that partition.


peterlimg:
#1055 is working on this.

PR #595, opened 2021-10-21
Stake Pool Offer fixes

Sriep:
Some non-intrusive changes to stake pools that improve the efficiency of several endpoints but particularly `cancel_allocation` and `finalize_allocation`. This is in advance of changes to interest and statsdb.

To do:
1. finalise only needs to [update the Blobber reward](https://github.com/0chain/0chain/blob/staging/code/go/0chain.net/smartcontract/storagesc/allocation.go#L1423). But is taking significant time to save the whole stake pool. Suggest move rewards and this code to the new stats_db
2. cancel allocations [extends the offers for an allocation](https://github.com/0chain/0chain/blob/staging/code/go/0chain.net/smartcontract/storagesc/allocation.go#L1253), but needs to get and save the whole stake pool with offers for all locations. Getting and saving from the MPT takes considerable time as it's in a for loop. stakePool.Offers needs to be reworked.


Sriep:
Depreciated replace with https://github.com/0chain/0chain/pull/854

PR #607, opened 2021-10-27
Update allocaion request changes

### LINKED ISSUE #538: Allow anyone to extend allocation
~~Modify `zbox update allocation`, with a new `--update-positive-expiry` option. This will allow anyone to lock tokens in an allocation to extend it's expiry.~~

~~Add a --no-expiry flag for an allocation. If this is set, the expiry period cannot be reduced and the allocation cannot be cancelled or finalized.~~

Sriep:
1. Allow anyone to lock tokens to increase expiry.
2. Do not allow allocation to have their expiry period reduced.
3. permanent: new allocation settings to forbid anyone from closing or finalising the allocation. Default = false.

Previously anyone can update an allocation by bypassing `Gosdk`, which has also been fixed as a result of this PR.

Matching GoSDK change https://github.com/0chain/gosdk/pull/270 and zbox change https://github.com/0chain/zboxcli/pull/118


Sriep:
Replaced with https://github.com/0chain/0chain/pull/620

Removing the permanate change.

PR #617, opened 2021-10-30
Remove restrictions on cancelling allocations

Sriep:


Sriep:
Replace with https://github.com/0chain/0chain/pull/636

PR #620, opened 2021-10-30
Allow anyone to extend allocaion

### LINKED ISSUE #538: Allow anyone to extend allocation

Sriep:
Allocations cannot have their duration reduced.
A new NFT allocation option which can be set in a new location request; defaulting to false. It allows anyone to extend the allocation.

Replaces https://github.com/0chain/0chain/pull/607 and https://github.com/0chain/0chain/issues/538
Previously anyone can update an allocation by bypassing `Gosdk`, which has also been fixed as a result of this PR.
Matching GoSDK https://github.com/0chain/gosdk/pull/288 and zbox Pr https://github.com/0chain/zboxcli/pull/122


stewartie4:
@Sriep should the allocation owner not be the only wallet able to update an allocation? 
where do the tokens come from if someone else updates it?

Sriep:
We are not using NFT optino any more. Change of plan. Closing.

PR #633, opened 2021-11-02
Fix min lock demand

### LINKED ISSUE #632: Cancel allocation min lock demand paid from challenge pool
In cancel allocation and finalise allocation the min lock demand payment is taken from the challenge pool rather than allocation write pools. This can result in `error not enough tokens in pool <pool id>:challengepool` errors.

Sriep:
Take min lock demand payments from allocation write pools, not the challenge pool.


Sriep:
https://github.com/0chain/system_test/actions/runs/1414184575 passed system tests.

PR #636, opened 2021-11-03
Remove restrictions on cancelling allocations

Sriep:
Allocations can always be cancelled independently of challenges.


Sriep:
Replaced with https://github.com/0chain/0chain/pull/637

PR #655, opened 2021-11-08
0chain events

### LINKED ISSUE #624: Add events to 0chain
Similar to go-ethereum events.

Use these events to emit information that will only be needed for REST API, and not by the blockchain.
Write a database app to monitor 0chain events and process the information. Support REST QPI endpoints with information from events.

Sriep:
This introduces 0chain Events, a way for smartcontracts to emit data that is no longer needed for future smartcontracts but will be needed for REST API queries. These events should be considered to take the same place as go-ethereum events.

1. A new posgreSQL database, eventDb. The sharder starts it during initialisation and performs any required auto migration.
2. For sc-REST API methods, a new StateContextI method GetEventDb. This allows REST API endpoints to query the event db for information. A link is thus passed by the sharder to incoming API calls.
3. Two new StatContextI methods were added for smart contracts. `EmitEvent` and `EmitError`. These emit data to be stored off-chain.
4. Any smartcontract returning an error has its events replaced with an Error event using `EmitError`. 
5. A new Events member of the block object to hold emitted events and later pass to the event database. The new StateContextI method `GetEvents` is used to populate this block Events field.
6. A new sc-REST API endpoint `GetEventsHandler` which provides a view on the evnetDb. Requires changes in GoSDK https://github.com/0chain/gosdk/pull/298 and zwalletCLI https://github.com/0chain/zwalletcli/pull/61.
7. Currently the only information being stored in the eventDb, is errors returned by smartcontracts. This will change over time. The new `./zwallet events --filters "tx_hash" --values "<tx hash>"` command can be used to get the error for a previously failed transaction.
8. Added mutex protection to some StateContextI methods. This is so that we can run concurrent `for { go ... }` loops in smart contracts to improve efficiency.


code/go/0chain.net/chaincore/block/entity.go
```diff
@@ -903,6 +916,11 @@ func (b *Block) ComputeStateLocal(ctx context.Context, c Chainer) error {
 		}
 	}
 
+	if len(b.Events) > 0 && c.GetEventDb() != nil {
```
dabasov, line 919: I'm not a fan of nil references, it would be better to use stub implementation of EventDb, with empty methods and use this implementation as default init value. So we can forget about nil checks, and if not initialized with real implementation it will do nothing

code/go/0chain.net/chaincore/block/entity.go
```diff
@@ -903,6 +916,11 @@ func (b *Block) ComputeStateLocal(ctx context.Context, c Chainer) error {
...
+		go c.GetEventDb().AddEvents(b.Events)
+		b.Events = nil
```
dabasov, line 921: Will we ever need  events later? May be we can keep them in block for future use or some logic depends on its emptiness?

code/go/0chain.net/chaincore/chain/entity.go
```diff
@@ -204,6 +208,32 @@ func (c *Chain) AsyncSyncBlocks(ctx context.Context, req SyncBlockReq) error {
 	return nil
 }
 
+func (c *Chain) SetupEventDatabase() error {
+	if c.EventDb != nil {
```
dabasov, line 212: with stub db simply can  call c.EventDb.Close()

code/go/0chain.net/chaincore/chain/state/state_context.go
```diff
@@ -159,6 +169,35 @@ func (sc *StateContext) GetMints() []*state.Mint {
...
+		Data:        data,
+	})
+}
+
+func (sc *StateContext) EmitError(err error) {
+	sc.events = []event.Event{
+		{
+			BlockNumber: sc.block.Round,
+			TxHash:      sc.txn.Hash,
+			Type:        "Error",
```
dabasov, line 197: hardcode

code/go/0chain.net/smartcontract/dbs/event/event.go
```diff
@@ -0,0 +1,92 @@
...
+
+	"0chain.net/smartcontract/dbs"
+	"gorm.io/gorm"
+)
+
+type Event struct {
+	gorm.Model
+	BlockNumber int64  `json:"block_number"`
+	TxHash      string `json:"tx_hash"`
+	Type        string `json:"type"`
```
dabasov, line 20: We can use int enum here or if we want to use string use something like this
```go
type EventType int
const (
    Payment EventType = iota
    Error
)

func (et EventType) String() string {
    return [...]string{"Payment", "Error"}[et]
}
```

code/go/0chain.net/smartcontract/dbs/event/event_test.go
```diff
@@ -0,0 +1,150 @@
...
+	filter = Event{
+		BlockNumber: 2,
+		Type:        "Error",
+	}
+	filterEvents, err = eventDb.FindEvents(filter)
+	require.NoError(t, err)
+	require.Len(t, filterEvents, 1)
+}
+
+/*
```
dabasov, line 113: can we delete this lines?

Sriep:
https://github.com/0chain/system_test/runs/4199832808?check_suite_focus=true Passing system tests

code/go/0chain.net/chaincore/chain/state/state_context.go
```diff
@@ -113,6 +127,8 @@ func (sc *StateContext) GetTransaction() *transaction.Transaction {
 
 //AddTransfer - add the transfer
 func (sc *StateContext) AddTransfer(t *state.Transfer) error {
+	sc.mutex.Lock()
```
dabasov, line 130: why do we need mutex in smart contracts? I assume it is executed sequentially always

code/go/0chain.net/chaincore/block/entity.go
Sriep: My intention is to convert them into a hash at this point. Then add to block hash.

PR #658, opened 2021-11-09
Split StateContextI

### LINKED ISSUE #657: Use different StateContext interface for smarcontracts and REST endpoints
Smart contracts and REST endpoints have a different sets of operations that they are permitted to perform. The StateContext interface they are passed should reflect that.

For example Rest endpoints might have `GetTrieNode` but not `InsertTrieNode`.

Sriep:
Currently, there is one [StateContextI](https://github.com/0chain/0chain/blob/staging/code/go/0chain.net/chaincore/chain/state/state_context.go#L32) interface that is used everywhere. This splits it up into three interfaces to better match its uses in various locations.

1. `StateContextI`: This is the pre-existing version that is used in smart contracts and in the chain for control operations. This could be split up further later if needed.
2. `RestStateContextI`: Methods that are needed for REST API endpoints. In particular, we can't change the state in these endpoints so these methods have no meaning. Also shortly we will introduce the EventDB and the StatsDB which will accessible from the REST endpoints but not smart contracts.
3. `ReadOnlyStateContextI`: Methods common to both smart-contracts and REST endpoints.

https://github.com/0chain/0chain/issues/657

### DIFF
--- a/code/go/0chain.net/chaincore/chain/state.go
+++ b/code/go/0chain.net/chaincore/chain/state.go
@@ -140,6 +140,17 @@ func (c *Chain) NewStateContext(b *block.Block, s util.MerklePatriciaTrieI,
 		c.GetSignatureScheme)
 }
 
+func (c *Chain) NewRestStateContext(b *block.Block, s util.MerklePatriciaTrieI,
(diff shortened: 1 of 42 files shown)


Sriep:
We need this, eventually. Close for the moment but can be reopened later.

PR #666, opened 2021-11-10
Refactor: organise the way to store grpc generated stubs for miner & shaders

### LINKED ISSUE #647: Reusable GRPC handlers for Miners and shraders
Miners and Sharders have common handlers in `chaincore` (Such as diagnostics.SetupHandlers() or transaction.SetupHandlers()).

Should ideally create a common GRPC package for Miners and Sharder and find a way to share the grpc-gateway handlers between them.

dhruv0000:
Signed-off-by: dhruv0000 <patel.4@iitj.ac.in>

Added a restructure:
```bash
.
├── buf.gen.yaml
├── buf.work.yaml
└── code
    └── go
        └── 0chain.net
            ├── gen
            │   └── proto
            │       ├── minerproto
            │       │   └── api
            │       │       └── v1
            │       │           ├── miner_contract.pb.go
            │       │           ├── miner_service_grpc.pb.go
            │       │           ├── miner_service.pb.go
            │       │           └── miner_service.pb.gw.go
            │       └── sharderproto
            │           └── api
            │               └── v1
            │                   ├── sharder_contract.pb.go
            │                   ├── sharder_service_grpc.pb.go
            │                   ├── sharder_service.pb.go
            │                   └── sharder_service.pb.gw.go
            ├── miner
            │   ├── buf.yaml
            │   └── minerproto
            │       └── api
            │           └── v1
            │               ├── miner_contract.proto
            │               └── miner_service.proto
            └── sharder
                ├── buf.yaml
                └── sharderproto
                    └── api
                        └── v1
                            ├── sharder_contract.proto
                            └── sharder_service.proto
```


buf.gen.yaml
```diff
@@ -9,3 +9,12 @@ plugins:
   - name: grpc-gateway
     out: code/go/0chain.net/miner/proto/api/src/proto
     opt: paths=source_relative,allow_delete_body=true
+  # - name: go
```
dhruv0000: The Sharder code seems to be generated even without specifing in `gen` file :confused: . A bit confused here.

twiny: I think this can be solved if we created a shared package for miner, sharder .. proto  if we have a directory like:
`code/go/0chain.net/grpc/` inside it `miner/proto/v1` & `sharder/proto/v1` and Go code will be generated in `code/go/0chain.net/grpc/{miner|sharder}/src`

I think will need to change `buf.yaml` and create new one inside `proto/{miner|sharder}`

PR #673, opened 2021-11-14
vrfs attack Byzantine test

rrrooommmaaa:
Add next integration test:
```
# 1. All nodes communicate honestly until round 10th
# 2. Lock notarization of round 10 in all nodes (block verification tickets sending)
# 3. The adversarial node sends VRF for round 11
# 4. All nodes receive adversarial node VRF for round 11, cache it and unblock the sending of pending verification tickets 
# 5. Verify whether the network continues operating correctly
```

Check the created integration test by running the next commands.
```
$ ./docker.local/bin/build.miners-integration-tests.sh && ./docker.local/bin/build.miners-integration-tests.sh

$ ./docker.local/bin/start.conductor.sh spam-miners
```

Although not used it also adds a directive to check whether a round was finalized.


code/go/0chain.net/miner/protocol_send_integration_tests.go
```diff
@@ -57,6 +67,36 @@ func (mc *Chain) SendVRFShare(ctx context.Context, vrfs *round.VRFShare) {
...
+		return // not requested or not initialized via SendVRFShare() yet
+	}
+	if vrfs != nil {
+		logging.Logger.Info("sendVRFSSpam() invoked by actual SendVRFShare()")
+		if setup.Vrfs != nil {
+			logging.Logger.Info("sendVRFSSpam() nothing to do")
+			return
+		}
+		// first send
+		setup.Vrfs = vrfs
```
dabasov, line 112: We should send VRFShare for the next round, I guess we should do something like 
```go
var (
	v = &round.VRFShare{}
	err  error
)
v.RoundTimeoutCount = r.GetTimeoutCount()
r := mc.StartNextRound(ctx, vrfs.Round) // possibly need *Round here not round number
v.Round = r.Round
v.Share, err = mc.GetBlsShare(ctx, r.Round)

v.SetParty(node.Self.Underlying())
setup.Vrfs = v
```

rrrooommmaaa: I tried this setup, however, there are still not suffucient configuration to `StartNewRound()` -- returns err

code/go/0chain.net/miner/miner/integration_tests.go
```diff
@@ -17,3 +23,29 @@ func initIntegrationsTests(id string) {
...
+		logging.Logger.Info("Setting VRFSSpam mode...")
+		miner.VRFSSpamFlag = true
+		go func(ctx context.Context) {
+			for {
+				select {
+				case <-ctx.Done():
+					return
+				default:
+					time.Sleep(100 * time.Millisecond)
+					miner.SendVRFSSpam(ctx, nil)
```
fabioDMFerreira: Shouldn't we do this after the miner has been setup?

code/go/0chain.net/miner/protocol_send_integration_tests.go
fabioDMFerreira: The job of this method is to send VRFs. Why do we have to force the start of a new round here? Is this the spamming mechanism?

fabioDMFerreira (dismissed):
What's the expected result of this test? Is this test to check whether the miners blacklist the spamming miner?

I couldn't also get the spamming mechanism. Are we considering the spam mechanism to be the miner forcing the existence of 2 uncompleted rounds?

code/go/0chain.net/miner/protocol_send_integration_tests.go
rrrooommmaaa: yes, [Attack round transfer](https://github.com/0chain/0chain/wiki/Byzantine-tests#attack-round-transfer) scenario means that the adversarial starts to send new VRF shares when it's not expected.

code/go/0chain.net/miner/miner/integration_tests.go
rrrooommmaaa: Do you mean to say `applyAdversarialMode()` function is called in a wrong place?

fabioDMFerreira: I don't know if it is being called in the wrong place. What's the relationship with `miner.SetupX2MResponders()`?

rrrooommmaaa: `SetupX2MResponders` puts in place handlers that react to VRF shares received from other nodes, e.g. `SignShareRequestHandler` (thus the node participates in a normal DKG process), and for this particular test I switched all of these handlers off and put frequent generation of own VRFShares instead to simulate VRF attack by simplest means I saw.
Note: these handlers were refactored after my initial PR.

docker.local/config/conductor.spam-miners.yaml
```diff
@@ -0,0 +1,49 @@
...
+tests:
+  # VRF
+  - name: "Spam VRFS to everyone"
+    flow:
+      - set_monitor: "sharder-1"
+      - cleanup_bc: {}
+      - start: ["sharder-1"]
+      - start: ["miner-1", "miner-2", "miner-3", "miner-4"]
+      - round_has_finalized:
+          spammers: ["miner-1"]
```
fabioDMFerreira: @dabasov as in the test case there is a reference to having a replica sending the VRF in the next round I will change this configuration to accept the node type and rank type (GetNodeTypeAndTypeRank) instead of the miner identification.

fabioDMFerreira: Wondering whether the spammer should send another "spamming" VRF after the first block has been finalized.

code/go/0chain.net/chaincore/chain/protocol_round.go
```diff
@@ -650,3 +650,37 @@ func getMagicBlockBrief(b *block.Block) *MagicBlockBrief {
...
+// If ranks is not computed, returns -1, -1.
+//
+// 	Explaining type rank example:
+//		Generators num = 2
+// 		len(miners) = 4
+// 		Generator0:	rank = 0; typeRank = 0.
+// 		Generator1:	rank = 1; typeRank = 1.
+// 		Replica0:	rank = 2; typeRank = 0.
+// 		Replica1:	rank = 3; typeRank = 1.
+func GetNodeTypeAndTypeRank(roundNum int64) (nodeType, typeRank int) {
```
dabasov: this method is only used in integration tests, so please move it to appropriate file

code/go/0chain.net/chaincore/chain/protocol_round_integration_tests.go
```diff
@@ -51,39 +47,6 @@ func addResultOnFinalizeRoundIfNeeded(r round.RoundI) {
 	cfg.Resulted = true
 }
 
-const (
```
dabasov: why is it removed?

code/go/0chain.net/miner/m_handler_integration_tests.go
```diff
@@ -87,6 +95,22 @@ func NotarizationReceiptHandler(ctx context.Context, entity datastore.Entity) (i
 		}()
 	}
 
+	state := crpc.Client().State()
+
+	if state.RoundHasFinalized != nil && state.RoundHasFinalized.Round == int(not.Round) && utils.IsSpamReceiver(state, not.Round) {
+		m := GetMinerChain()
+		mr := m.getOrCreateRound(ctx, (int64)(state.RoundHasFinalized.Round+1))
```
dabasov: so we get the next to finalized round and wait for VRF there, but it's not correct, since this round is guaranteed to be notarized and it is finished, we need the next round to it, so if r0 is finalized the current round is r0+2 or r0+3. Why we can't take simply current round and wait for VRFs there?

code/go/0chain.net/miner/protocol_receive_integration_tests.go
```diff
@@ -308,6 +311,28 @@ func getNotarisationInfo(not *Notarization) *cases.NotarisationInfo {
...
+func (mc *Chain) HandleNotarizedBlockMessage(ctx context.Context,
+	msg *BlockMessage) {
+	mc.handleNotarizedBlockMessage(ctx, msg)
+}
+
+// handleVRFShare - handles the vrf share.
+func (mc *Chain) HandleVRFShare(ctx context.Context, msg *BlockMessage) {
+	state := crpc.Client().State()
+
+	if state.RoundHasFinalized != nil && state.RoundHasFinalized.Round+1 == int(msg.VRFShare.Round) && utils.IsSpamReceiver(state, msg.VRFShare.Round-1) {
```
dabasov: this round is finished already

dabasov (dismissed):

code/go/0chain.net/chaincore/chain/protocol_round.go
fabioDMFerreira: I started to use it on https://github.com/0chain/0chain/pull/673/files#diff-b43ca4d90940c9705539979478cc66f924542be9af5ae0f3386ae1b94edd1b84R56, but as this code is only used on integration tests I will adapt it to be compiled only on integration tests builds.

fabioDMFerreira: fixed on [39fc22c](https://github.com/0chain/0chain/pull/673/commits/39fc22c0e7097c8372abb8b1559e0c8c3581de47)

code/go/0chain.net/chaincore/chain/protocol_round_integration_tests.go
fabioDMFerreira: fixed on [39fc22c](https://github.com/0chain/0chain/pull/673/commits/39fc22c0e7097c8372abb8b1559e0c8c3581de47)

code/go/0chain.net/miner/m_handler_integration_tests.go
fabioDMFerreira: Here we are putting on hold the handling of the notarization of the current round to prevent the current round from being finalized. Only after receiving the spammed VRF of the next round, we will proceed to the handling.

I guess the idea was to lock the notarization. @dabasov what do you suggest to do?

code/go/0chain.net/miner/protocol_receive_integration_tests.go
fabioDMFerreira: This is a VRF of the next round. What do you mean?

code/go/0chain.net/miner/protocol_receive_integration_tests.go
```diff
@@ -308,6 +310,28 @@ func getNotarisationInfo(not *Notarization) *cases.NotarisationInfo {
...
+func (mc *Chain) HandleNotarizedBlockMessage(ctx context.Context,
+	msg *BlockMessage) {
+	mc.handleNotarizedBlockMessage(ctx, msg)
+}
+
+// handleVRFShare - handles the vrf share.
+func (mc *Chain) HandleVRFShare(ctx context.Context, msg *BlockMessage) {
+	state := crpc.Client().State()
+
+	if state.RoundHasFinalizedConfig != nil && state.RoundHasFinalizedConfig.Round+1 == int(msg.VRFShare.Round) && chain.IsSpamReceiver(state, msg.VRFShare.Round-1) {
```
fabioDMFerreira: @dabasov I've changed the property name to prevent misleading.

fabioDMFerreira:
@dabasov any other changes, you want me to do?

sagchain:
@dabasov should we even track this? This has open without any activity since long.

code/go/0chain.net/conductor/cases/round_has_finalized.go
```diff
@@ -0,0 +1,85 @@
...
+	}
+
+	fmt.Printf("adding result -> %+v %+v\n", roundInfo, n.roundID)
+
+	if roundInfo.Num == int64(n.roundID) && (n.res == nil || n.res.Num == 0) {
+		defer func() {
+			n.prepared <- struct{}{}
+		}()
+		n.res = new(RoundInfo)
+		return n.res.Decode(blob)
```
peterlimg: Instead of decoding again, we can use the `roundInfo` directly.

code/go/0chain.net/chaincore/chain/utils_integration_tests.go
```diff
@@ -0,0 +1,34 @@
...
+package chain
+
+import (
+	crpc "0chain.net/conductor/conductrpc"
+	"0chain.net/conductor/config/cases"
+)
+
+// IsSpammer checks whether a node is a spammer.
+// A list of spammers names with format "miner-x" are passed, then the x is extracted and compared with the node index.
+func IsSpammer(spammers []cases.NodeTypeTypeRank, roundNum int64) (isSpammer bool) {
```
peterlimg, line 28: Duplicate code of the utils_integration_tests.go in utils package, should only keep one of them

code/go/0chain.net/conductor/utils/utils_integration_tests.go
```diff
@@ -0,0 +1,62 @@
...
+	}
+	return
+}
+
+// IsSpammer checks whether a node is a spammer.
+// A list of spammers names with format "miner-x" are passed, then the x is extracted and compared with the node index.
+func IsSpammer(spammers []cases.NodeTypeTypeRank, roundNum int64) (isSpammer bool) {
+	isSpammer = false
+
+	nodeType, typeRank := chain.GetNodeTypeAndTypeRank(roundNum)
```
peterlimg: Using `chain` package here may cause cycle import issue later if the `chain` package need to call methods from `utils` package.

code/go/0chain.net/miner/protocol_receive_integration_tests.go
peterlimg: The finalized round is usually 2 or 3 rounds behind the current round, basically if round 10 is finalized, the next round 11 must have been notarized(guaranteed by protocol). So perhaps name it as RoundHasNotarizedConfig would be more make sense.

peterlimg (changes requested):

code/go/0chain.net/conductor/cases/round_has_finalized.go
fabioDMFerreira: resolved by [ffdcb20](https://github.com/0chain/0chain/pull/673/commits/ffdcb206bb5aebc4544d12090ed493f0a83eeb45)

code/go/0chain.net/chaincore/chain/utils_integration_tests.go
fabioDMFerreira: resolved by [ffdcb20](https://github.com/0chain/0chain/pull/673/commits/ffdcb206bb5aebc4544d12090ed493f0a83eeb45)

code/go/0chain.net/conductor/utils/utils_integration_tests.go
fabioDMFerreira: resolved by [ffdcb20](https://github.com/0chain/0chain/pull/673/commits/ffdcb206bb5aebc4544d12090ed493f0a83eeb45)

peterlimg (approved):
LGTM, good job 👍

PR #676, opened 2021-11-16
fix: bring out errors with stack trace

kushthedude:
Signed-off-by: Kush Trivedi <kushthedude@gmail.com>

### DIFF
--- a/code/go/0chain.net/chaincore/chain/state.go
+++ b/code/go/0chain.net/chaincore/chain/state.go
@@ -3,9 +3,8 @@ package chain
 import (
 	"context"
 	"fmt"
+	"github.com/pkg/errors"
 	"time"
-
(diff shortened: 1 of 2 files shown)


peterlimg (changes requested):
Wraps errors is good, but we have to change the error processing code too, otherwise the code would cause unexpected result. For example, if we wrap errors of `util.NodeNotFound` , `context.Canceled` `context.DeadlineExceeded`, then the process logic [here](https://github.com/0chain/0chain/blob/411af96acee3d9bf9b042e7dd45e7a77546730a7/code/go/0chain.net/chaincore/block/entity.go#L880-L881) will break.

peterlimg:
The changes will break smart contract error process logic, so I'll close this as this is not being updated.

PR #678, opened 2021-11-18
fix(docker): improved docker build with go mod cache

cnlangzi:
- added go mod cache to improve docker build 
- moved gozstd to base image


dabasov (commented):
lgtm

PR #683, opened 2021-11-21
GRPC miner migration v2 - refactor

twiny:


buf.gen.yaml
```diff
@@ -1,4 +1,4 @@
-version: v1beta1
+version: v1
```
kushthedude, line 1: Why?

buf.lock
```diff
@@ -2,9 +2,9 @@
 version: v1
 deps:
   - remote: buf.build
-    owner: beta
+    owner: googleapis
     repository: googleapis
     branch: main
```
kushthedude, line 7: Was this auto-changed?

kushthedude (commented):
Did you try to start the server on your local? Did it work?

buf.gen.yaml
twiny: this is an update - v1 is a released after the v1beta1.

buf.lock
twiny: Repository `buf.build/beta/googleapis` is deprecated: This module has deprecated in favor of `buf.build/googleapis/googleapis`. Please use that instead.

buf.gen.yaml
kushthedude: Understood

Issue #540, opened 2021-09-25
Make the "Not enough blobbers to honor the allocation" error more informative


Sriep:
Make the ["Not enough blobbers to honor the allocation" ](https://github.com/0chain/0chain/blob/staging/code/go/0chain.net/smartcontract/storagesc/allocation.go#L433) more informative.

It often confuses people when they see this message and don't know what is the problem that is preventing enough blobbers being selected.

Sriep:
Moved to https://github.com/0chain/0chain/issues/569 which is now using preferred blobbers for input. It is not practical to list why every blobber in the blockchain is not eligible, but we can do this for the user inputted list of blobbers.

PR #684, opened 2021-11-22
Adding blobbers to newAllocation and removing prefferedBlobbers

### LINKED ISSUE #668: New Allocation Request
New Allocation will no longer choose the blobbers, instead, the client will include a list of blobbers with the new allocation request. 

The [selectBlobbers](https://github.com/0chain/0chain/blob/staging/code/go/0chain.net/smartcontract/storagesc/allocation.go#L342) method can be reduced. Instead, it will only verify the user-selected blobbers as in [filterBlobbers](https://github.com/0chain/0chain/blob/staging/code/go/0chain.net/smartcontract/storagesc/allocation.go#L428); which can be improved to provide better error messages for the end user https://github.com/0chain/0chain/issues/540. 

The preferred blobbers field can be repurposed and renamed as needed. 

Changes to gosdk https://github.com/0chain/gosdk/issues/297 and zbox https://github.com/0chain/zboxcli/issues/125 will be needed to support this.

This is a rewrite of https://github.com/0chain/0chain/issues/569.

moldis:
- removing PreferredBlobbers from NewAllocationRequest
- removing the functionality to fetch all blobbers from MPT and then find appropriate blobber to handle allocation
- instead the list of blobbers is to be expected in request and validated further
- freeAllocation is also updated with the same, it has now a blobbers key which take []string as input key


cnlangzi:
@moldis  don't forget to remove `prefferedBlobbers` stuff from gosdk and zbox

code/go/0chain.net/smartcontract/storagesc/allocation.go
```diff
@@ -333,14 +325,9 @@ func (sc *StorageSmartContract) newAllocationRequestInternal(
...
-	var seed int64
-	if seed, err = strconv.ParseInt(t.Hash[0:8], 16, 64); err != nil {
-		return "", common.NewError("allocation_creation_failed",
-			"Failed to create seed for randomizeNodes")
-	}
-
 	blobberNodes, bSize, err := sc.selectBlobbers(
-		t.CreationDate, *allBlobbersList, sa, seed, balances)
+		t.CreationDate, *allBlobbersList, sa, balances, request.Blobbers)
+
```
Sriep, line 421: We should not need to pass the `allBlobberList` here. 

We should expect hundreds of thousands of blobbers here. Use of  `allBlobbers` objects is deprecated and will be removed as part of optimisation.

code/go/0chain.net/smartcontract/storagesc/handler.go
```diff
@@ -98,7 +98,8 @@ func (ssc *StorageSmartContract) GetAllocationMinLockHandler(ctx context.Context
 	var sa = request.storageAllocation()
 
 	blobberNodes, bSize, err := ssc.selectBlobbers(
-		creationDate, *allBlobbersList, sa, int64(creationDate), balances)
+		creationDate, *allBlobbersList, sa, balances, request.Blobbers)
+
```
Sriep, line 272: We do not need the `allBlobbersList` here.

code/go/0chain.net/smartcontract/storagesc/allocation.go
```diff
@@ -426,7 +413,7 @@ func (sc *StorageSmartContract) selectBlobbers(
 	// size of allocation for a blobber
 	var bSize = (sa.Size + int64(size-1)) / int64(size)
 	var list = sa.filterBlobbers(allBlobbersList.Nodes.copy(), creationDate,
-		bSize, filterHealthyBlobbers(creationDate),
```
Sriep, line 432: We should not pass the `allBlobberList` here. Instead, pass the `blobbers` input parameter.

Previously the `filterBlobbers` method iterated through `all` the blobbers filtering out the ones that did not fit the new allocation request parameters. 

We want to repurpose `filterBlobbers` to filter out any input blobbers that do not fit the allocation request parameters.

If any blobbers are filtered then we should expect a not enough blobbers error. 

https://github.com/0chain/0chain/issues/540 Uses complain that the error messaging here is unhelpful. Could we change `filterBlobbers` to be more informative about filtered out blobbers? 

One solution would be to return a list of filtered out blobber ids with the reason they were filtered. This could be appended to an error message if it turns out that there are not enough valid blobbers to form an allocation.

code/go/0chain.net/smartcontract/storagesc/allocation_test.go
```diff
@@ -189,8 +186,9 @@ func TestSelectBlobbers(t *testing.T) {
 			t.Parallel()
 			ssc, sa, blobbers, balances := setup(t, tt.args)
 
+			var bl []string
 			outBlobbers, outSize, err := ssc.selectBlobbers(
-				now, blobbers, &sa, randomSeed, balances,
+				now, blobbers, &sa, balances, bl,
```
Sriep: remove `blobbers`

code/go/0chain.net/smartcontract/storagesc/benchmark_tests.go
```diff
@@ -172,7 +171,6 @@ func BenchmarkTests(
 					Expiration:                 common.Timestamp(viper.GetDuration(bk.StorageMinAllocDuration).Seconds()) + now,
 					Owner:                      data.Clients[0],
 					OwnerPublicKey:             data.PublicKeys[0],
-					PreferredBlobbers:          blobberUrls[:8],
```
Sriep, line 178: Could rename this to the new Blobbers field rather than lose completely, otherwise I would expect a not enough blobbers error. 
Also, rename the test to 
```go
name:     "storage.new_allocation_request
```

code/go/0chain.net/smartcontract/storagesc/benchmark_tests.go
```diff
@@ -140,7 +140,6 @@ func BenchmarkTests(
 					Expiration:                 common.Timestamp(viper.GetDuration(bk.StorageMinAllocDuration).Seconds()) + now,
 					Owner:                      data.Clients[0],
 					OwnerPublicKey:             data.PublicKeys[0],
-					PreferredBlobbers:          []string{},
```
Sriep, line 146: We do not need this test anymore so can be removed.

code/go/0chain.net/smartcontract/storagesc/benchmark_rest_tests.go
```diff
@@ -103,7 +103,6 @@ func BenchmarkRestTests(
 					Expiration:                 2*common.Timestamp(viper.GetDuration(bk.StorageMinAllocDuration).Seconds()) + now,
 					Owner:                      data.Clients[0],
 					OwnerPublicKey:             data.PublicKeys[0],
-					PreferredBlobbers:          []string{},
```
Sriep, line 124: We should probably replace the line with `Blobbers:          blobberUrls[:8],` rather than delete it. Otherwise, we should expect a not enough blobbers error.

Sriep (changes requested):
I think we need a  `max_blobbers_per_allocation` `sc.yaml` setting. Can't have users asking for, and getting 80,000 blobbers. This could default to 40.

code/go/0chain.net/smartcontract/storagesc/allocation.go
moldis: @Sriep Maybe I misunderstood. My idea is to pass the list of blobbers (urls of blobbers) in the allocation endpoint, then this filter is picking up blobbers by URL and a few other filters like healthy, size etc. In any case, I will need to iterate through blobbers to get Blobber object?

Sriep: A list of 100,000+ blobbers is too much to keep in the MPT. Takes over 50ms to get and update. So that is being removed soon. So pretend the allBlobbers list is not there, but leave the existing code, it will be removed shortly.

Instead, the list of blobbers checked here has to be loaded individually. Updating 40 items from the MPT is bad, but probably only about 10-20ms so we can handle it. 

I can's see any reason the same `filterBlobbers` method can't be used to check the user-selected list of blobbers the same way it filtered the all blobbers list. But could be wrong.

We want to avoid doing anything that potentially involves manipulating hundred of thousands of blobbers.

Sriep: Actually, if you could remove the all blobbers list entirely from new allocation request, that would help.

Sriep: I just modified https://github.com/0chain/0chain/pull/691 to supply `/getBlobbers` with information from the eventDb. I believe that the [all blobbers list](https://github.com/0chain/0chain/blob/staging/code/go/0chain.net/smartcontract/storagesc/blobber.go#L17) is only used in new allocation requests and [/getBlobbes](https://github.com/0chain/0chain/blob/blobber-events/code/go/0chain.net/smartcontract/storagesc/handler.go#L53). The other references are just maintaining the list. 

Any endpoint that maintains the all blobbers list has to take >50ms with 100,000+ blobbers so removing it will be a solid step towards making our smartcontracts scalable.

code/go/0chain.net/smartcontract/storagesc/benchmark_tests.go
```diff
@@ -119,36 +119,6 @@ func BenchmarkTests(
...
-					ReadPriceRange:             PriceRange{0, state.Balance(viper.GetInt64(bk.StorageMaxReadPrice) * 1e10)},
-					WritePriceRange:            PriceRange{0, state.Balance(viper.GetInt64(bk.StorageMaxWritePrice) * 1e10)},
-					MaxChallengeCompletionTime: viper.GetDuration(bk.StorageMaxChallengeCompletionTime),
-					DiversifyBlobbers:          false,
-				}).encode()
-				return bytes
-			}(),
-		},
 		{
 			name:     "storage.new_allocation_request_preferred",
```
Sriep: `storage.new_allocation`,  remove request_preferred.

code/go/0chain.net/smartcontract/storagesc/benchmark_tests.go
```diff
@@ -172,7 +142,6 @@ func BenchmarkTests(
 					Expiration:                 common.Timestamp(viper.GetDuration(bk.StorageMinAllocDuration).Seconds()) + now,
 					Owner:                      data.Clients[0],
 					OwnerPublicKey:             data.PublicKeys[0],
-					PreferredBlobbers:          blobberUrls[:8],
 					ReadPriceRange:             PriceRange{0, state.Balance(viper.GetInt64(bk.StorageMaxReadPrice) * 1e10)},
```
Sriep: Why no `newAllocationRequest.Blobbers = blobberUrls[0:4]` or similar  field? Should this not return a not enough blobbers error?

code/go/0chain.net/smartcontract/storagesc/benchmark_tests.go
```diff
@@ -172,7 +142,6 @@ func BenchmarkTests(
...
 					WritePriceRange:            PriceRange{0, state.Balance(viper.GetInt64(bk.StorageMaxWritePrice) * 1e10)},
 					MaxChallengeCompletionTime: viper.GetDuration(bk.StorageMaxChallengeCompletionTime),
```
Sriep: Remove `torage.new_allocation_request_diversify` benchmark below. We don't need it any more.

code/go/0chain.net/smartcontract/storagesc/blobber_test.go
```diff
@@ -215,19 +212,15 @@ func Test_flow_reward(t *testing.T) {
...
-			b1 = b
-			break
-		}
-	}
-	require.NotNil(t, b1)
+	require.EqualValues(t, len(blobs), len(alloc.BlobberDetails))
+	var b1 *Client = blobs[0]
 
-	require.EqualValues(t, 202546280, alloc.restMinLockDemand())
+	// QUESTION HERE
```
Sriep, line 222: The blobber objects used int `newAllocationRequestInternal`  don't have a valid Terms.MinLockDemand value. Get blobber from blockchain to make sure.

code/go/0chain.net/smartcontract/storagesc/blobber_test.go
```diff
@@ -286,18 +283,21 @@ func Test_flow_reward(t *testing.T) {
...
+		require.EqualValues(t, readPoolFund-1e8, rp.allocTotal(allocID, tp))
+		require.EqualValues(t, 199e8, rp.allocBlobberTotal(allocID, b1.id, tp))
 
 		// min lock demand reducing
 		alloc, err = ssc.getAllocation(allocID, balances)
 		require.NoError(t, err)
-		require.EqualValues(t, 192418966, alloc.restMinLockDemand())
+
+		// TODO not clear for me
+		//require.EqualValues(t, 192418966, alloc.restMinLockDemand())
```
Sriep, line 300: As above blobbers do not have valid Terms.MinLockDemand. Use blobbers from MTP not dodgy user input.

code/go/0chain.net/smartcontract/storagesc/allocation.go
```diff
@@ -171,11 +170,11 @@ type newAllocationRequest struct {
 	Expiration                 common.Timestamp `json:"expiration_date"`
 	Owner                      string           `json:"owner_id"`
 	OwnerPublicKey             string           `json:"owner_public_key"`
-	PreferredBlobbers          []string         `json:"preferred_blobbers"`
 	ReadPriceRange             PriceRange       `json:"read_price_range"`
 	WritePriceRange            PriceRange       `json:"write_price_range"`
 	MaxChallengeCompletionTime time.Duration    `json:"max_challenge_completion_time"`
 	DiversifyBlobbers          bool             `json:"diversify_blobbers"`
+	Blobbers				   []*StorageNode	`json:"blobbers"`
```
Sriep: []String = list of blobber ids should be good enough. 
Its painful getting all the blobbers from the MPT, but we can't assume users will enter valid data.

code/go/0chain.net/smartcontract/storagesc/allocation.go
```diff
@@ -336,20 +322,21 @@ func (sc *StorageSmartContract) newAllocationRequestInternal(
 
 	var sa = request.storageAllocation() // (set fields, including expiration)
 
-	var seed int64
-	if seed, err = strconv.ParseInt(t.Hash[0:8], 16, 64); err != nil {
-		return "", common.NewError("allocation_creation_failed",
-			"Failed to create seed for randomizeNodes")
+	if len(request.Blobbers) > conf.MaxBlobbersPerAllocation{
```
Sriep: We should probably check len(request.Blobbers) > request.DataShards + request.ParityShards as well, somewhere around here.

code/go/0chain.net/smartcontract/storagesc/allocation.go
```diff
@@ -336,20 +322,21 @@ func (sc *StorageSmartContract) newAllocationRequestInternal(
...
+		return "", common.NewErrorf("allocation_creation_failed",
+			"Too many blobbers selected, max available %d", conf.MaxBlobbersPerAllocation)
 	}
 
 	blobberNodes, bSize, err := sc.selectBlobbers(
-		t.CreationDate, *allBlobbersList, sa, seed, balances)
+		t.CreationDate, sa, balances, request.Blobbers)
```
Sriep: We should get the blobbers from the MPT rather than rely on user input to be valid. 

Its painful but we  have `max_blobbers_per_allocation` to ensure that its ok. Plus its been budgeted for, check spreadsheet https://docs.google.com/spreadsheets/d/1oqY-gyt5itAJLxsh748CkFB8OOQsR8KClbzTrUfDf2A/edit#gid=677534893. Note spreadsheet assumes upcoming changes including partitions.


I suspect this is why you are having problems with the blobber_test.go minLockDemand results. The blobber.Terms.MinLockDemand field is zero in the input so givers wrong value on line 353 below.

code/go/0chain.net/smartcontract/storagesc/free_allocation.go
```diff
@@ -30,8 +30,9 @@ func (frm *freeStorageMarker) decode(b []byte) error {
 }
 
 type freeStorageAllocationInput struct {
-	RecipientPublicKey string `json:"recipient_public_key"`
-	Marker             string `json:"marker"`
+	RecipientPublicKey string         `json:"recipient_public_key"`
+	Marker             string         `json:"marker"`
+	Blobbers           []*StorageNode `json:"blobbers"`
```
Sriep: As for allocation. I think []string = list of blobber ids is all that's needed.

Sriep (changes requested):

code/go/0chain.net/smartcontract/storagesc/blobber_test.go
ssardana96: @Sriep didn't get it
I'm taking now blobbers from MPT only but, still MinLockDemand is 0
as all blobbers are getting entered in allAlloc() func, where blobbers are created for test and pushed to MPT

ssardana96:
@Sriep can you review it

code/go/0chain.net/smartcontract/storagesc/benchmark_tests.go
ssardana96: outdated

code/go/0chain.net/smartcontract/storagesc/allocation.go
```diff
@@ -384,6 +459,8 @@ func (sc *StorageSmartContract) newAllocationRequestInternal(
 		return "", common.NewError("allocation_creation_failed", err.Error())
 	}
 
+	var allBlobbersList *StorageNodes
+	allBlobbersList, err = sc.getBlobbersList(balances)
```
Sriep, line 464: Should be ok to remove this block of code.
```go
	var allBlobbersList *StorageNodes
	allBlobbersList, err = sc.getBlobbersList(balances)
	err = updateBlobbersInAll(allBlobbersList, blobberNodes, balances)
	if err != nil {
		return "", common.NewError("allocation_creation_failed", err.Error())
	}
```
The all blobbers list is depreciated and should be safe to remove. Currently very expensive and does nothing.

Sriep (changes requested):
Looks ok. 
The `zbox` and event database changes could allow system tests to pass, the exception is preferred blobbers - this might need renaming or changing in the sytem_tests.

`zbox newallocaion` remaining unchanged from a black-box perspective.

code/go/0chain.net/smartcontract/storagesc/allocation.go
```diff
@@ -283,6 +283,59 @@ func (sc *StorageSmartContract) filterBlobbersByFreeSpace(now common.Timestamp,
...
+		if _, ok := inputBlobberMap[b]; !ok {
+			inputBlobberMap[b] = true
+		}
+	}
+
+	input := make(chan string, len(inputBlobberMap))
+	output := make(chan *StorageNode, len(inputBlobberMap))
+	defer close(output)
+	for pool := 0; pool < poolSize; pool++ {
+		go func() {
```
dabasov: I don't think we should use concurrency in smart contracts tbh

dabasov: do we get any performance boost?

Sriep: My benchmarks suggest it can halve the speed of smart contracts. So using parallel for loops could well allow us to have a maximum of say 40 blobbers per allocation rather than 20.

What is the problem?

ssardana08: 40 blobbers, concurrency:8 get_blobbers	{"time_taken": "167ns"}
40 blobbers, concurrency:4 get_blobbers	{"time_taken": "208ns"}
40 blobbers, concurrency disabled get_blobbers	{"time_taken": "250ns"}

code/go/0chain.net/smartcontract/storagesc/allocation.go
ssardana96: @Sriep couple of unit_tests again start to break because of this.
for now can we keep this? maybe we can take this in some other issues.

code/go/0chain.net/smartcontract/storagesc/allocation.go
ssardana08: these are timings for this function with and without cocurrency

dabasov: afaik we never used it and all structures used in smartcontracts don't have proper synchronization, if we allow to use it somewhere we should code accordingly, for example StateContextI is not designed for concurrent usage, look here [field access](https://github.com/0chain/0chain/blob/c73e2717354d70f33a33fe1a73487ca85f506d14/code/go/0chain.net/chaincore/chain/state/state_context.go#L290)

dabasov: in my opinion we should parallelize work on some higher level and implement smartcontracts just as not concurrent code

peterlimg:
Please fix the conflicts, thanks

code/go/0chain.net/smartcontract/storagesc/handler.go
```diff
@@ -254,23 +256,20 @@ func (ssc *StorageSmartContract) GetAllocationMinLockHandler(ctx context.Context
 	allocData := params.Get("allocation_data")
 	var request newAllocationRequest
 	if err = request.decode([]byte(allocData)); err != nil {
-		return "", common.NewErrInternal("can't decode allocation request", err.Error())
+		logging.Logger.Error("can't decode allocation request",
+			zap.Error(err))
```
peterlimg, line 260: Return error instead of logging only

peterlimg (changes requested):

ssardana08:
@samtholiya is this completed? @dabasov ?
if not needed, can we close it?

samtholiya:
This is not needed

Issue #432, opened 2021-07-26
Hard coded smatcontract owner


Sriep:
This could be a security problem. Would it be better to set up these values as part of chain genesis?

guruhubb:
yes, these need to different along with chain ID for every network we deploy, so needs to be generated at the onset and then used in yaml to deploy the network. But this is more of a deployment issue, but I agree we need to parameterize this and have a yaml to generate these address at the time of deployment.

anish-squareops:
@Sriep please add more details on "how to generate these".

Sriep:
I assume the same way as any other (id, public key, private key) triplet the administrator of the chain presumable guards the private key. Check out `./zwallet register`.

PR #685, opened 2021-11-22
Moved smart contract owner to config

dabasov:
Introduced new interface `Authorizer` to smart contracts. It is similar to ethereum (openeppelin) Ownable. 
When sc needs to authorize access to any of its funcs it can embed implementation of this interface in smart contract and call `Authorize()` at the start of func.
```
func (fc *FaucetSmartContract) updateSettings(t *transaction.Transaction, inputData []byte, balances c_state.StateContextI, gn *GlobalNode,) (string, error) {
	if err := fc.Authorize(t.ClientID, "update_settings"); err != nil {
		return "", err
	}
```


dabasov:
There is alternative approach though, with slightly different `Authorize()` interface

```go
type Authorizer interface {
	Authorize(clientId, f func()) error
}
```
and wrap existing code with authorize logic, it can be more pure as we can hide `return error`  inside func, but it brings more boilerplate, since we have to introduce new func to be wrapped. It worth discussing, but I prefer the first variant

code/go/0chain.net/chaincore/smartcontractinterface/auth.go
```diff
@@ -0,0 +1,13 @@
+package smartcontractinterface
+
+import (
+	"0chain.net/core/common"
+)
+
+func AuthorizeWithOwner(funcName string, hasAccess func() bool) error {
```
Sriep, line 7: Why define the `hasAccess` function outside `AuthorizeWithOwner` as it's always the same? Will we ever want to use a different `hasAccess` funcion? What does
```go
	if err := smartcontractinterface.AuthorizeWithOwner("update_settings", func() bool {
		return gn.FaucetConfig.OwnerId == t.ClientID
	}); err != nil {
		return "", err
	}
```
gain over
```go
	if err := smartcontractinterface.AuthorizeWithOwner(
		gn.FaucetConfig.OwnerId, t.ClientID,
	); err != nil {
		return "", common.NewError("update_settings", err)
	}
```
?

code/go/0chain.net/smartcontract/storagesc/curator_test.go
```diff
@@ -40,6 +40,7 @@ func TestAddCurator(t *testing.T) {
 			ClientID: p.clientId,
 		}
 		var ssc = &StorageSmartContract{
+
```
Sriep: There does not appear to be any meaningful changes in this file. Probably best to revert them.

code/go/0chain.net/smartcontract/storagesc/challenge_test.go
```diff
@@ -153,6 +153,7 @@ func TestAddChallenge(t *testing.T) {
 			t.Parallel()
 			args := parametersToArgs(tt.parameters)
 			var ssc = &StorageSmartContract{
+
```
Sriep: No meaningful changes, maybe remove this file from PR.

code/go/0chain.net/smartcontract/storagesc/block_reward_test.go
```diff
@@ -85,6 +85,7 @@ func TestPayBlobberBlockRewards(t *testing.T) {
 			Nodes: []*StorageNode{},
 		}
 		var ssc = &StorageSmartContract{
+
```
Sriep: No real changes, can we remove this file from Pr.

code/go/0chain.net/smartcontract/storagesc/benchmark_setup.go
```diff
@@ -24,6 +24,7 @@ func AddMockAllocations(
 	const mockMinLockDemand = 1
 	var (
 		sscId = StorageSmartContract{
+
```
Sriep: No meaningful changes can we remove from Pr?

code/go/0chain.net/smartcontract/storagesc/funded_pools_test.go
```diff
@@ -26,6 +26,7 @@ func TestAddToFundedPools(t *testing.T) {
 	) (*StorageSmartContract, cstate.StateContextI) {
 		var balances = &mocks.StateContextI{}
 		var ssc = &StorageSmartContract{
+
```
Sriep: No real changes, can we revert and remove from PR?

Sriep (changes requested):

code/go/0chain.net/chaincore/smartcontractinterface/auth.go
dabasov: This is common logic that we use in different places, we might want something more complex and hasAccess can differ from time to time (we can use several fields, some check to authorized maps and so on)
In your example, you can't  have different equality logic for ownerId and ClientId (imagine you want to call isEqual() instead of '==' ) or some other logic like less than (stake < 1000) or contains in the map. It should give us  agility in the future and creates near zero boilerplate. Maybe it'll be better to rename `AuthorizeWithOwner ` to `Authorize`

```go

func (sc *SC) somefunc(..) {
 // authorize for big stack
if err := smartcontractinterface.Authorizer("some_func", func() bool {
		return sc.getStake(t.ClientID) > 100000
	}); err != nil {
		return "", err
	}
}
```

code/go/0chain.net/smartcontract/storagesc/curator_test.go
dabasov: it left from draft, should have removed it

Sriep (approved):

Sriep:
We can link this to https://github.com/0chain/0chain/issues/432

stewartie4:
ignore the cancelled check below, it just means that the system tests against the default branch failed but they have since passed manually against the feature/block-rewards-test-updates branch

PR #688, opened 2021-11-22
Build time improvement for tests

dabasov:
Added caching of `go mod download` to dockerfile

### DIFF
--- a/docker.local/build.unit_test/Dockerfile
+++ b/docker.local/build.unit_test/Dockerfile
@@ -8,6 +8,11 @@ COPY ./code/go/0chain.net/go.mod $SRC_DIR/go/0chain.net/
 COPY ./code/go/0chain.net/go.sum $SRC_DIR/go/0chain.net/
 
 RUN cd $SRC_DIR/go/0chain.net && go mod download -x
+
+RUN cd $GOPATH/pkg/mod/github.com/valyala/gozstd@v1.14.* && \
+    chmod -R +w . && \
+    make clean libzstd.a
+
 RUN go get github.com/vektra/mockery/v2/.../
 
 COPY ./code/go/0chain.net $SRC_DIR/code/go/0chain.net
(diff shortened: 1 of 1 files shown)


kushthedude (commented):
Are we sure getting faster times? Seeing from the history of runs the time is similar to old tests cc: @cnlangzi

cnlangzi:
> Are we sure getting faster times? Seeing from the history of runs the time is similar to old tests cc: @cnlangzi

yes. it is faster on local. but it is only improvement for local, not github action. if you want to improve it on github action, please improve it like `make local-run` on blobber

docker.local/build.unit_test/Dockerfile
```diff
@@ -4,14 +4,17 @@ ENV GO111MODULE=on
...
+RUN go get github.com/vektra/mockery/v2/.../
+
 COPY ./code/go/0chain.net $SRC_DIR/code/go/0chain.net
 COPY ./generate_mocks.sh $SRC_DIR/generate_mocks.sh
 
-RUN go get github.com/vektra/mockery/v2/.../
 RUN cd $SRC_DIR && ls && ./generate_mocks.sh
 
-RUN cd $SRC_DIR/code/go/0chain.net && \
-    go mod download
```
cnlangzi, line 23: copy gozstd from zbuild_base, it will save some time to build it again. https://github.com/0chain/0chain/blob/556fedc2aad08b9efeb0b4901ad377d2928cd611/docker.local/build.sharder/Dockerfile#L19

cnlangzi (changes requested):

docker.local/build.unit_test/Dockerfile
dabasov: will it be faster if we move existing code upper, so we could cache it? 
In sharder we do this step after code copy so it can not be cached

cnlangzi: yes. `move upper` should be a solution too.

cnlangzi (approved):
LGTM

PR #691, opened 2021-11-25
Blobber events table

### LINKED ISSUE #680: Blobber events database table plus endpoint.
We need to add a new table to the event database. This is to hold a blobber table. 
0chain to emit events each time blobber information changes. The event database can then respond by updating the blobber table.
A new endpoint for querying the blobber table. The initial use of this will be to select a suitable set of blobbers for an allocation request.

Waiting on https://github.com/0chain/0chain/pull/679

Sriep:
Creates a blobber table in the event database.
Populates the table by emitting events from 0chian smartcontracts.
Rewrite GetBlobberHandler to use the event database rather than the MPT.
Integrate with benchmark program. If enabled requires a local Postgres installation.


docker.local/build.sharder/b0docker-compose.yml
```diff
@@ -39,19 +39,6 @@ services:
...
-      POSTGRES_PORT: 5432
-      POSTGRES_HOST: postgres
-      POSTGRES_USER: postgres
-      POSTGRES_HOST_AUTH_METHOD: trust
-    volumes:
-      - ../bin:/0chain/bin
-      - ../sql:/0chain/sql
-      - ../sql_script/:/docker-entrypoint-initdb.d/
-    links:
-      - postgres:postgres
```
kushthedude, line 54: Why did we remove the psql config?

docker.local/build.sharder/b0docker-compose-debug.yml
```diff
@@ -39,19 +39,6 @@ services:
...
-  postgres-post:
-    image: postgres:14
-    environment:
-      POSTGRES_PORT: 5432
-      POSTGRES_HOST: postgres
-      POSTGRES_USER: postgres
-      POSTGRES_HOST_AUTH_METHOD: trust
-    volumes:
-      - ../bin:/0chain/bin
-      - ../sql:/0chain/sql
```
kushthedude, line 51: Can we also add a psql for the eventsDB if we want to test setup with local chain?

code/go/0chain.net/smartcontract/dbs/postgresql/postresql.go
```diff
@@ -76,3 +76,49 @@ func (store *PostgresStore) Close() {
...
+
+
+
+
+
+
+
+
+
+
```
kushthedude: What is this empty block 🤔 ?

docker.local/build.sharder/b0docker-compose.yml
Sriep: I was not sure it was needed. I can put it back easily enough.

Sriep:
Passing system tests https://github.com/0chain/system_test/runs/4325200913?check_suite_focus=true

code/go/0chain.net/smartcontract/dbs/event/interface.go
```diff
@@ -0,0 +1,29 @@
...
+	return nil
+}
+
+func (_ *NoOpEventDb) AutoMigrate() error {
+	return nil
+}
+
+func (_ *NoOpEventDb) Close() {}
+
+func (_ *NoOpEventDb) AddEvents(_ []Event) {}
```
kushthedude: Are these NoOp funcs being used?

code/go/0chain.net/smartcontract/dbs/event/blobber.go
```diff
@@ -0,0 +1,162 @@
...
+	var blobber = Blobber{BlobberID: updates.Id}
+	exists, err := blobber.exists(edb)
+
+	if err != nil {
+		return err
+	}
+	if !exists {
+		return fmt.Errorf("blobber %v not in database cannot update",
+			blobber.BlobberID)
+	}
```
kushthedude, line 77: Can we not merge these two `err!=nil || !exists` ?

code/go/0chain.net/smartcontract/dbs/event/event.go
```diff
@@ -108,11 +60,50 @@ func (edb *EventDb) GetEvents(block int64) ([]Event, error) {
...
 
 func (edb *EventDb) drop() error {
-	return edb.Store.Get().Migrator().DropTable(&Event{})
+	err := edb.Store.Get().Migrator().DropTable(&Event{})
+	if err != nil {
+		return err
+	}
+
+	err = edb.Store.Get().Migrator().DropTable(&Blobber{})
+	if err != nil {
```
kushthedude: Can we use error wrapping here? It makes traces easy to trace https://pkg.go.dev/github.com/pkg/errors#Wrap

code/go/0chain.net/smartcontract/dbs/event/event.go
```diff
@@ -108,11 +60,50 @@ func (edb *EventDb) GetEvents(block int64) ([]Event, error) {
...
+		return false, fmt.Errorf("error counting events matching %v, error %v",
+			event, result.Error)
+	}
+	return count > 0, nil
+}
+
+func (edb *EventDb) removeDuplicate(events []Event) []Event {
+	for i := len(events) - 1; i >= 0; i-- {
+		exists, err := edb.exists(events[i])
+		if err != nil {
```
kushthedude, line 78: Shall we not return the error here & stop the execution of function?

kushthedude (commented):
Some suggestion & minor neats.

code/go/0chain.net/smartcontract/dbs/event/eventdb.go
```diff
@@ -0,0 +1,36 @@
...
+	}
+	return eventDb, nil
+}
+
+type EventDb struct {
+	dbs.Store
+}
+
+func (edb *EventDb) AutoMigrate() error {
+	if err := edb.Store.Get().AutoMigrate(&Event{}); err != nil {
```
moldis: AutoMigrate accept few income params, it can be AutoMigrate(&Event{}, &Blobber{})

code/go/0chain.net/smartcontract/dbs/event/blobber.go
```diff
@@ -0,0 +1,162 @@
...
+	return count > 0, nil
+}
+
+func (bl *Blobber) create(edb *EventDb) error {
+	result := edb.Store.Get().Create(bl)
+	return result.Error
+}
+
+func (edb *EventDb) GetBlobber(id string) (*Blobber, error) {
+	exists, err := (&Blobber{
```
moldis: I think can avoid  'exists' call, as the next DB call will return an error if the item does not exist

code/go/0chain.net/smartcontract/dbs/event/blobber.go
```diff
@@ -0,0 +1,162 @@
...
+
+func (edb *EventDb) deleteBlobber(data []byte) error {
+	result := edb.Store.Get().
+		Where("blobber_id = ?", string(data)).Delete(&Blobber{})
+	return result.Error
+}
+
+func (edb *EventDb) updateBlobber(data []byte) error {
+	var updates dbs.DbUpdates
+	err := json.Unmarshal(data, &updates)
```
moldis: As this is DB layer, we might better pass Blobber object in params instead of raw []byte and avoid conversion here

code/go/0chain.net/smartcontract/dbs/event/blobber.go
```diff
@@ -0,0 +1,162 @@
...
+
+	result := edb.Store.Get().
+		Model(&Blobber{}).
+		Where("blobber_id = ?", updates.Id).
+		Updates(updates.Updates)
+	return result.Error
+}
+
+func (edb *EventDb) addBlobber(data []byte) error {
+	var blobber Blobber
```
moldis: Same comment as above to replace param from []byte to Blobber

code/go/0chain.net/smartcontract/dbs/event/blobber.go
```diff
@@ -0,0 +1,162 @@
...
+		return err
+	}
+
+	exists, err := blobber.exists(edb)
+
+	if err != nil {
+		return err
+	}
+	if exists {
+		return edb.overwriteBlobber(blobber)
```
moldis, line 117: As function name `addBlobber` it might give a confusion for another developer. Consider removing overwriteBlobber from DB layer and moving it to condition, like if blobber exists then add otherwise ovveride, outside of this file

code/go/0chain.net/smartcontract/dbs/event/blobber.go
```diff
@@ -0,0 +1,162 @@
...
+		return err
+	}
+	if !exists {
+		return fmt.Errorf("blobber %v not in database cannot update",
+			blobber.BlobberID)
+	}
+
+	result := edb.Store.Get().
+		Model(&Blobber{}).
+		Where("blobber_id = ?", updates.Id).
```
moldis: It can be Where(&Blobber{blobber_id: updates.Id}). to follow GORM style

code/go/0chain.net/smartcontract/dbs/event/blobber.go
```diff
@@ -0,0 +1,162 @@
...
+		return nil, err
+	}
+	if !exists {
+		return nil, fmt.Errorf("blobber %v not in event db", id)
+	}
+
+	var blobber Blobber
+	result := edb.Store.Get().
+		Model(&Blobber{}).
+		Find(&Blobber{}).
```
moldis: I think you can avoid using Find and just use First instead. Also I am not sure how it will unwrap Blobber object, in where you adding blobber_id, but Blobber doesn't have blobber_id field

moldis (changes requested):
Minor changes with DB logic

docker.local/build.sharder/b0docker-compose-debug.yml
Sriep: Don't follow, could you explain further?

code/go/0chain.net/smartcontract/dbs/event/blobber.go
Sriep: They are logically different.
If you have an error you want to pass that error up the stack.
If the blobber can't be found you want to create a new error.

code/go/0chain.net/smartcontract/dbs/event/event.go
Sriep: https://pkg.go.dev/github.com/pkg is not being used at the  moment. This method is only called in tests. I'll move the method to a `_test.go` file.

code/go/0chain.net/smartcontract/dbs/event/event.go
Sriep: Just because one of the events is bad does not mean that all the other events emitted this block should be affected. We don't want one bad transaction to damage other good transactions just because they happened in the same block.

code/go/0chain.net/smartcontract/dbs/event/interface.go
Sriep: No. Ok i will remove I can always get them from here if I need them later.

code/go/0chain.net/smartcontract/dbs/event/blobber.go
Sriep: Blobber has a `BlobberID` field, and gorm seems to convert that to `blobber_id`, `BlobberID` would probably also work. I should be consistent here, for example [overwriteBlobber](https://github.com/0chain/0chain/blob/blobber-events/code/go/0chain.net/smartcontract/dbs/event/blobber.go#L100).

code/go/0chain.net/smartcontract/dbs/event/blobber.go
Sriep: Ok, it's a slightly more compact notation.

Sriep:
System test pass https://github.com/0chain/system_test/runs/4333958756?check_suite_focus=true

docker.local/build.sharder/b0docker-compose-debug.yml
Sriep: I don't follow why we need it, but I put the ` postgres-post:` back.

Issue #601, opened 2021-10-25
Do not verify block if the VRF shares have not reached threshold


peterlimg:
A miner should not verify a block if it does not have enough VRFShares consensus for the RRS of the block.

Check details here: https://0chain.slack.com/archives/C01V27E2Z9N/p1634872520004600

peterlimg:
The chain moves pretty slow after adding the logic to do not verify block if the VRF shares have not reached threshold. Sometimes it takes more than 10 seconds to move on to the next round. The chain moves slow because some miners could not verify the proposed block as they have not yet collect enough VRF shares. Because they could not verify the block, they would not send out the block verification tickets, therefore, slowed down the tickets collection. This could be even worse if the miner has limited CPU resources and when the CPU is busy, they could drop the VRF share messages due the timeout for validating N2N messages. 

We can fix the above issue by adding VRF shares to the verify block message, the miner who sends out the verify block message must already have collected enough VRF shares. So when a miner received a proposed block for verification, we can verify and merge the VRF shares from the message into local round. Only when the VRF shares are valid and reached the threshold, will the miner move to next step to do the block verification. I did some tests, and this worked pretty well, basically the chain moves the same speed as before. 

What do you think @guruhubb @dabasov

dabasov:
@peterlimg do you have any idea, why network started to move slowly? in theory majority of miners should move at the network speed, since they all reach threshold (well it is the way to find out what is the network speed). 
I want to say, that couple of slow miners must not affect the majority and they have to move fast despite of 1/3 of delayed or even Bizantine. 

Your solution can work but for me it can be excessive and bring network overhead. With such approach we try to solve network issues, in fact we resend messages that was not received earlier, well, for protocol to work well we still need synchrony assumption.

I'm seriously worried not even about network traffic, but that this "slowness" can hide some other serious issues, which we can hide, for example some networking bugs  (a lot of network failures we talk with @guruhubb recently) and so on.

Resending messages in general can be reasonable, we can piggy back even previous blocks, to reduce number of fetching requests for example, and it can be good way to sacrifice some network traffic for latency. Just suggest to investigate this slowness issue.

PR #712, opened 2021-11-27
Do not verify block and use the RRS of the block before VRF shares meet the threshold

peterlimg:
Fixes: 
- #601 

Changes:
- Do not verify block if the round does not have VRF shares that could meet consensus threshold.
- Add VRF shares to the `verify block` message so that receiver could use them to verify and merge the VRF shares into local round, and then start to verify the block. Return error if the VRF shares are invalid.
- Do not send verify block message when received VRFShare.Round is < current round, send notarized block should be fine.


dabasov:
I don't think we should resend VRFShare, synchronous protocol guarantees that we receive shares before the verify block, the only way we could miss it due to node offline, so it is not a problem for delayed replica to synchronize on notarization for this round.
The cons of this solution I'm worried about is network traffic, Imagine we have k generators and n miners with vrf threshold like 2/3. We will increase the size of traffic for verify round by 2/3*k*n*size(share) (we assume honest generators to have all threshold of shares)
Since we store full node structure in the vrf share, wonder why we can't store only miner key, traffic can grow tremendously.

I suppose at least to calculate message size of vrf share and figure out how much traffic we should to trade of

dabasov (changes requested):
I don't think we should resend VRFShare, synchronous protocol guarantees that we receive shares before the verify block, the only way we could miss it due to node offline, so it is not a problem for delayed replica to synchronize on notarization for this round.
The cons of this solution I'm worried about is network traffic, Imagine we have k generators and n miners with vrf threshold like 2/3. We will increase the size of traffic for verify round by 2/3kn*size(share) (we assume honest generators to have all threshold of shares)
Since we store full node structure in the vrf share, wonder why we can't store only miner key, traffic can grow tremendously.

I suppose at least to calculate message size of vrf share and figure out how much traffic we should to trade of

peterlimg:
@dabasov you're right, the network traffic would be a problem when nodes number increases. I've figured out why the chain is slow, there are mainly two places that can lead to slow collecting of VRF shares:
1. VRF shares could be dropped before they could be processed if the N2N sender request signature verification are failed due to CPU busy caused timeout. 
2. VRF shares will be ignored if it's previous round has no seed. 

So I did some tests to fix the above issues by
- Do not start the sender request signature verification immediately when the message is received, instead we can read out the data and do some basic check, and if it can not pass the check, we can ignore the message without doing the signature verification. In this way, we can save lot of CPU resources, and reduce the possibility that VRF shares are dropped due to the signature verification timeout error.
- Cache the received VRF shares for a round if the previous round has no seed yet, verify and collect those VRF shares once the previous round is ready. In this way, we don't need to wait for the re-sending of VRF shares.

So far, the fix looks good, check it [here](http://216.218.228.197:7087/_diagnostics).

code/go/0chain.net/miner/m_handler.go
```diff
@@ -191,11 +191,12 @@ func VRFShareHandler(ctx context.Context, entity datastore.Entity) (
...
-		go func() {
-			mb.Miners.SendTo(ctx, VerifyBlockSender(hnb), found.ID)
-			mb.Miners.SendTo(ctx, MinerNotarizedBlockSender(hnb), found.ID)
-		}()
+		if err := node.ValidateSenderSignature(ctx); err != nil {
+			return nil, err
+		}
+
+		// send notarized block
+		go mb.Miners.SendTo(ctx, MinerNotarizedBlockSender(hnb), found.ID)
```
dabasov, line 199: I think it is not necessary to resend notarized block to delayed miner

dabasov (commented):
LGTM

code/go/0chain.net/miner/m_handler.go
dabasov: We can create an issue or discuss it in view-change

dabasov (approved):

Issue #728, opened 2021-12-01
VRF shares collect issue caused chain stuck or slow


peterlimg:
RRS generation is slow after running for a while, and then caused the chain moves pretty slow.

PR #729, opened 2021-12-01
Fix round restarting bug that caused RRS generation slow

peterlimg:
Fixes:
- #728 

Changes:
- Do not restart round if the round's soft timeout count is less than the `mc.RoundRestartMult`.


dabasov (commented):
lgtm

PR #731, opened 2021-12-01
Update configs on block finalization

### LINKED ISSUE #541: Possable problem with global settings.
@peterlimg point in https://github.com/0chain/0chain/pull/501 needs to be considered.

@Peter raised a point about updating global-settings. Prs501 and PR524.
Because global settings can be accessed at any time, not just while running a smart contract, its possible that things are not set up properly when queried. In particular the MPT might not be fully configure and the config settings are missing. We need to be able to handle this situation differently to when they are missing because the chain has just started and we should just get them from 0chian.yaml.

Peter  [05:01]
We could have partial state of MPT for new joined miners,  basically we can not sync the whole MPT from remote when it starts, it is highly possible that the miner could not get settings from the MPT, we shall wait till MPT has enough state to work. We may have changed the settings in MPT, if the new joined miner get settings from local, it would use different settings from others, which could lead problems. So the situations are:
a) The chain just started, we should init the MPT with local settings
b) The chain has run for a while, there are changes applied to the settings, a new miner joined the network, before it syncs up the MPT state,  it could not get settings from MPT, however, we shall not use local settings as well, because the settings could be different from others, and could cause problem if it use the different settings to participant the block VRF share verification, and block generation. (edited)

### LINKED ISSUE #716: Update configs on block finalization
Config can be updated through update_globals call on minersc. We need to update config struct after block is finalized, so MPT will never be altered and config stored on chain and in MPT will be consistent

dabasov:
Update Config on block finalization. Config will be updated in case there is new version of config in block's state


peterlimg:
Seems the `config` is not thread safe, if we update it in the `SetLatestFinalizedBlock()`, we'd better protect the config with locks as the `SetLatestFinalizedBlock()` could be called in different places concurrently.

code/go/0chain.net/chaincore/chain/entity.go
```diff
@@ -1367,6 +1370,68 @@ func (c *Chain) SetLatestFinalizedBlock(b *block.Block) {
...
+	}
+
+	return gl, nil
+}
+
+func (mc *Chain) updateConfig(pb *block.Block) {
+	clientState, err := mc.getClientState(pb)
+	if err != nil {
+		// This might happen after stopping and starting the miners
+		// and the MPT has not been setup yet.
```
Sriep, line 1327: Does this handle https://github.com/0chain/0chain/issues/541, if so can we link the issue to this Pr?

Sriep (approved):

dabasov:
@peterlimg Added synchronized access to config structure:
1. Extracted interface
2. Moved json structure to DTO, because it is not possible to parse json to not exported fields. 
3. Made Config structs not exportable, to guard access with RW lock
4. All updates to DTO is made through write lock

It is important to notice, that working to struct is still not fully concurrent safe, if we need to change or read several properties atomically we must guard such action with lock. 
Probably, it is not needed  for writing now and most likely won't be needed in the future until update will be made outside MPT, as now we make all changes through blockchain transaction which is atomic by nature. 
Reads can still need  some atomicity, when several properties depend on each other. Such atomicity can be achieved by extracting such dependant properties to single structure.

@peterlimg @Sriep do you know such dependant properties?

peterlimg (approved):

Sriep (approved):

PR #743, opened 2021-12-05
Fix LFMB sync isse

### LINKED ISSUE #742: Latest finalized magic block is not sycned correctly
The latest finalized magic block is not synced after `chain.ViewChange()` is called when `VC:false`. Actually, we should not call this function when `VC:false`. Two things need to fix:
- [x] Sync latest finalized magic block in the `chain.ViewChange()` 
- [x] Do not update the magic block in `chain.ViewChange()` when `VC:false`

The problem of the non-synced latest finalized magic block is: n2n will only send messages to nodes in magic block that is acquire from `chain.MagicBlockStorage()`, while status monitoring would only update nodes' active status on latest finalized magic block. Therefore, if any of the nodes in the `chain.MagicBlockStorage().MB` became inactive, they would never become to active even the nodes are reachable. The pool would not send messages to those nodes are inactive, so they will not able to receive messages, and would not be able to run correctly.

peterlimg:
Fix #742 

The latest finalized magic block should not be changed when `vc:false`, even if it is changed, it should be synced with the mb in `chain.MagicBlockStorage` so that the nodes' active status can be updated correctly.

### DIFF
--- a/code/go/0chain.net/chaincore/chain/entity.go
+++ b/code/go/0chain.net/chaincore/chain/entity.go
@@ -1501,6 +1501,7 @@ func (c *Chain) LatestOwnFinalizedBlockRound() int64 {
 }
 
 // SetLatestFinalizedBlock - set the latest finalized block.
+// TODO: this should be called when UpdateMagicBlock is called successfully
 func (c *Chain) SetLatestFinalizedMagicBlock(b *block.Block) {
 
 	if b == nil || b.MagicBlock == nil {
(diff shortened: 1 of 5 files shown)


peterlimg:
@dabasov commented:
> I’m a little skeptical about switching off mb update, we might need it even with change-view offed, for example if one of key will be compromised on mainnet, I’ll think it over and comment this issue later today.

dabasov:
Yeah, we might need to update magic block manually. Imagine situation when several miners compromise their keys after mainnet start and we need urgently regenerate magic block and change it on some round number. To make it happen, we will need some mechanics to change magic block and dkg-summaries. The easiest one is through config file after restart or maybe through REST call. 
Maybe it is better to create new issue for that as if it looks more like new functionality than changing this one.

dabasov (commented):
lgtm

PR #751, opened 2021-12-06
Make PR691 backward compatible, chains with no event database

### LINKED ISSUE #746: Make chain work with previous logic when db events is off
The chain should be able to function well when db events is off. Use the previous logic when the db events is off or when the config does not have the `dbs.events` field.

Sriep:
Change getBlobber and getBlobbers endpoints so they are backward compatible with chains that do not have event database installed.

Runs old pre-event database version if the new version fails.


peterlimg:
Except the blobbers API endpoints, are there other endpoints would be affected by the db events?

Sriep:
The event database is used in three endpoints at the moment `getBlobber`, `getBlobbers` and `getEvents`. getEvents is new so there is no backward compatibility issue.

PR #780, opened 2021-12-12
Run system tests automatically on PR commit

stewartie4:


.github/workflows/unit-test.yml
```diff
@@ -16,4 +16,15 @@ jobs:
         run: ./docker.local/bin/build.base.sh
 
       - name: unit tests
-        run: ./docker.local/bin/unit_test_build.sh
\ No newline at end of file
+        run: ./docker.local/bin/unit_test_build.sh
+
+  lint:
```
stewartie4, line 21: combined 2 files

.github/workflows/build-&-publish-docker-image.yml
```diff
@@ -1,8 +1,13 @@
 name: build-&-publish-docker-image
 
+concurrency:
+  group: "${{ github.ref }}"
+  cancel-in-progress: true
+
 on:
   push:
     branches: [ staging ]
+  pull_request:
```
stewartie4, line 10: any reason why this shouldn't publish on PR like blobber @cnlangzi ?

cnlangzi: I remember @anish-squareops seems says it makes too heavy tasks on our github runner. I would suggest to enable it too.

Issue #773, opened 2021-12-10
New storagesc update_validator_settings endpoint


Sriep:
We need a new storage smart contract endpoint, update_validator_settings. This will act in much the same way as [update_blobber_settings](https://github.com/0chain/0chain/blob/staging/code/go/0chain.net/smartcontract/storagesc/blobber.go#L224) but will update a validator's stake pool settings. 

Related to https://github.com/0chain/0chain/issues/772

ma2b0043:
@sriep when can we expect to close this issue?

Sriep:
ref https://github.com/0chain/0chain/issues/905

Issue #587, opened 2021-10-14
Validators - remove if stake is too low.


Sriep:
Have a minimum stake for validators. Remove validators that have too low a stake.

Sriep:
To do this the validator should be added or removed from the list of validators used during challenge creation during StakePoolLock and StakePoolUnlock respectively. 

the existing `all validators list` is soon to be replaced by `all validator partitions`. It would be best to delay this change until then. https://github.com/0chain/0chain/issues/785

ma2b0043:
@Sriep when can we expect to close this issue?

iamrz1:
TBD after https://github.com/0chain/0chain/pull/1439

PR #811, opened 2021-12-21
Partitions and all validator list

### LINKED ISSUE #785: Partition tables for validators, blobbers and blobber's allocations
We need to replace the all object lists with partition tables. We need them for validators, blobbers and blobber's allocations.

Most of the work has been done in the depreciated https://github.com/0chain/0chain/pull/588. We need to separate out the partition code.

This is a blocker for challenge and block reward issues. https://github.com/0chain/0chain/issues/753 and https://github.com/0chain/0chain/issues/741.

Sriep:

Adds a partition package. This is used to partition 0chain object so that objects can be selected at random. These changes are needed to support large numbers of objects, 100,000 and over and will be used to replace the all_allocations, all_validators and all_blobbers lists.
```go
type Partition interface {
	util.Serializable
	Add(PartitionItem, state.StateContextI) (int, error)
	Remove(PartitionItem, int, state.StateContextI) error

	SetCallback(ChangePartitionCallback)
	Size(state.StateContextI) (int, error)
	Save(state.StateContextI) error
}

type RandPartition interface {
	Partition
	AddRand(PartitionItem, *rand.Rand, state.StateContextI) (int, error)
	GetRandomSlice(*rand.Rand, state.StateContextI) ([]PartitionItem, error)
}
```
- `Add` adds to the end then swaps with a random item.
- `Remove` An object knows where it is. So remove then swap in the last item.
- `Random slice` Selects a partition at random.
- `SetCallback` There is a callback that is used to inform objects if their location changes.

The list of all validators is replaced with a validator partition. Each challenge needs a random selection of validators to be assigned. This uses the new partition to randomly select a slice of partitions. 

Validators are immortal so are never removed and the validator objects stored in the MPT are never used. Hence no need for callback. This will change https://github.com/0chain/0chain/issues/587 and https://github.com/0chain/0chain/issues/773.

The new partition package will also be needed In upcoming PRs. We will need a partition of blobbers, and a partition for allocations assigned to each blobber.

Fixes some unit tests to work with the new validator partitions.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,6 +2,7 @@
 
 ## Table of Contents
 
+
 - [Initial Setup](#initial-setup) 
   - [Directory Setup for Miners & Sharders](#directory-setup-for-miners-and-sharders) 
   - [Setup Network](#setup-network)
--- a/README.md
+++ b/README.md
@@ -2,7 +2,6 @@
 
 ## Table of Contents
 
-
 - [Initial Setup](#initial-setup) 
   - [Directory Setup for Miners & Sharders](#directory-setup-for-miners-and-sharders) 
   - [Setup Network](#setup-network)
--- a/code/go/0chain.net/smartcontract/benchmark/main/cmd/mpt.go
+++ b/code/go/0chain.net/smartcontract/benchmark/main/cmd/mpt.go
@@ -3,6 +3,8 @@ package cmd
 import (
 	"encoding/hex"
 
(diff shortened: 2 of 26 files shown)


Dmdv (approved):
Looks good
For inspiration: Random, Hash, Custom, RoundRobin partitioning:

https://github.com/Shopify/sarama/blob/main/partitioner.go

PR #816, opened 2021-12-23
fix(storagesc): ignore expriation checking if it is zero

cnlangzi:


stewartie4:
I see the has been ticked 
image
Has a new system test been added to ensure that when we set an expiry of 0 the expiration check does not run?

peterlimg (approved):
lgtm

PR #826, opened 2021-12-27
Fix SendEntityHandler canceled early issue

peterlimg:
- [x]  Do any new system tests need added to test this change? do any existing system tests need updated? If so create a branch at [0chain/system_test](https://github.com/0chain/system_test) then add and merge your test changes to master BEFORE merging this PR

The request is canceled too early

### DIFF
--- a/code/go/0chain.net/chaincore/node/n2n_send.go
+++ b/code/go/0chain.net/chaincore/node/n2n_send.go
@@ -141,7 +141,12 @@ func (np *Pool) sendTo(ctx context.Context, numNodes int, nodes []*Node, handler
 		case <-done:
 			doneCount++
 			if doneCount >= numNodes+THRESHOLD || doneCount >= activeCount {
-				logging.N2n.Info("send message (done)", zap.Int("all_nodes", len(nodes)), zap.Int("requested", numNodes), zap.Int("active", activeCount), zap.Int("sent_to", len(sentTo)), zap.Duration("time", time.Since(start)))
+				logging.N2n.Info("send message (done)",
+					zap.Int("all_nodes", len(nodes)),
+					zap.Int("requested", numNodes),
+					zap.Int("active", activeCount),
+					zap.Int("sent_to", len(sentTo)),
+					zap.Duration("time", time.Since(start)))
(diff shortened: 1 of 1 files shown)


code/go/0chain.net/chaincore/node/n2n_send.go
```diff
@@ -296,13 +303,13 @@ func SendEntityHandler(uri string, options *SendOptions) EntitySendHandler {
 				selfNode.SetLastActiveTime(ts)
 				selfNode.InduceDelay(receiver)
 
-				cctx, cancel := context.WithTimeout(ctx, timeout)
-				defer cancel()
+				cctx, cancel = context.WithTimeout(ctx, timeout)
```
cnlangzi, line 306: do you think it is better to move `cctx, cancel = context.WithTimeout(ctx, timeout)` to line 297?

peterlimg: No, actually we moved it into the `func()` block because the timeout context starts to count down immediately when it is created, we don't want the http request be canceled early due to the time waste on other processes such as `receiver.Grab()`.

cnlangzi: yes. I see. because `defer cancel()` was in `func()`. 
```suggestion

// move here, it is easier to understanding than inside `func()`
cctx, cancel := context.WithTimeout(ctx, timeout)

func() {
  receiver.Grab()
  defer receiver.Release()
  selfNode = Self.Underlying()
  selfNode.SetLastActiveTime(ts)
  selfNode.InduceDelay(receiver)

  req = req.WithContext(cctx)
  resp, err = httpClient.Do(req)
}()

	
defer cancel()
```

peterlimg: Moving it outside of the `func()` block would waste the context time on `receiver.Grab()`, especially when we have lot of nodes, part of the requests will be canceled before they are actually send out.

dabasov (commented):
lgtm

dabasov (approved):

PR #834, opened 2021-12-28
- Miner(event db)

ssardana96:
- [ ]  Do any new system tests need added to test this change? do any existing system tests need updated? If so create a branch at [0chain/system_test](https://github.com/0chain/system_test) then add and merge your test changes to master BEFORE merging this PR

### DIFF
--- a/code/go/0chain.net/smartcontract/minersc/miner.go
+++ b/code/go/0chain.net/smartcontract/minersc/miner.go
@@ -361,7 +361,7 @@ func (msc *MinerSmartContract) UpdateMinerSettings(t *transaction.Transaction,
 		return "", common.NewErrorf("update_miner_settings", "saving: %v", err)
 	}
 
-	if err = emitUpdateMiner(mn, balances); err != nil {
+	if err = emitUpdateMiner(mn, balances, false); err != nil {
 		return "", common.NewErrorf("update_miner_settings", "saving: %v", err)
 	}
 
--- a/code/go/0chain.net/smartcontract/minersc/miner.go
+++ b/code/go/0chain.net/smartcontract/minersc/miner.go
@@ -133,10 +133,10 @@ func (msc *MinerSmartContract) AddMiner(t *transaction.Transaction,
 	if _, ok := allMap[newMiner.GetKey()]; !ok {
 		allMiners.Nodes = append(allMiners.Nodes, newMiner)
 
-		//if err = updateMinersList(balances, allMiners); err != nil {
-		//	return "", common.NewErrorf("add_miner",
-		//		"saving all miners list: %v", err)
-		//}
+		if err = updateMinersList(balances, allMiners); err != nil {
+			return "", common.NewErrorf("add_miner",
+				"saving all miners list: %v", err)
+		}
 
 		err = emitAddMiner(newMiner, balances)
 		if err != nil {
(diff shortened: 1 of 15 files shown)


code/go/0chain.net/smartcontract/minersc/miner.go
```diff
@@ -133,23 +133,30 @@ func (msc *MinerSmartContract) AddMiner(t *transaction.Transaction,
 	if _, ok := allMap[newMiner.GetKey()]; !ok {
 		allMiners.Nodes = append(allMiners.Nodes, newMiner)
 
-		if err = updateMinersList(balances, allMiners); err != nil {
-			return "", common.NewErrorf("add_miner",
-				"saving all miners list: %v", err)
-		}
-		update = true
-	}
+		//if err = updateMinersList(balances, allMiners); err != nil {
```
Sriep: Why remove this block? We are only emitting events in smartcontracts, not changing how miners work.

code/go/0chain.net/smartcontract/minersc/miner.go
```diff
@@ -133,23 +133,30 @@ func (msc *MinerSmartContract) AddMiner(t *transaction.Transaction,
...
+			return "", common.NewErrorf("add_miner",
+				"insert new miner: %v", err)
 		}
 
-		msc.verifyMinerState(balances, "add_miner: Checking all miners list afterInsert")
-
 		update = true
 	}
 
+	//if !msc.doesMinerExist(newMiner.GetKey(), balances) {
```
Sriep: Again leave this block?

code/go/0chain.net/smartcontract/minersc/models.go
```diff
@@ -1218,13 +1221,34 @@ func (dmn *DKGMinerNodes) GetHashBytes() []byte {
 
 // getMinersList returns miners list
 func getMinersList(state cstate.StateContextI) (*MinerNodes, error) {
+
```
Sriep: The only changes to the smart contract code should be to emit events. 

Smartcontracts should have no access to the eventDb object, this is the reason for https://github.com/0chain/0chain/issues/657, to prevent people trying to use methods like GetEventDb in smart contracts.

code/go/0chain.net/smartcontract/minersc/models.go
```diff
@@ -821,6 +821,9 @@ type SimpleNode struct {
 
 	// LastHealthCheck used to check for active node
 	LastHealthCheck common.Timestamp `json:"last_health_check"`
+
```
Sriep, line 824: I don't think we need this. Instead emit en event in [createMagicBlock](https://github.com/0chain/0chain/blob/1e8779f639d24bcd832baaf07adbd2e31915b066/code/go/0chain.net/smartcontract/minersc/dkg.go#L798) with the `Status` update converting to a boolean Active.

Sriep (changes requested):
Expect the only change made to smartcontracts is to emit events. We should not be changing how smartcontracts work.

code/go/0chain.net/smartcontract/minersc/miner.go
ssardana96: done

code/go/0chain.net/smartcontract/minersc/miner.go
ssardana96: done

code/go/0chain.net/smartcontract/minersc/models.go
ssardana96: noted. So in this case is someone does GetEventDb() (even through not should be done), it should return nil, right?
That's the expected behavior?

ssardana96:
Closing as it is from forked

Issue #757, opened 2021-12-08
Event db: Ensure events processed in correct order.


Sriep:
[Events are emitted](https://github.com/0chain/0chain/blob/staging/code/go/0chain.net/chaincore/block/entity.go#L920)
```
go c.GetEventDb().AddEvents(b.Events)
```
and in [AddEvents](https://github.com/0chain/0chain/blob/staging/code/go/0chain.net/smartcontract/dbs/event/process.go#L31)
```
func (edb *EventDb) AddEvents(events []Event) {
    for loop that process events.
}
```
We need to ensure that events are processed in the correct order. That database tables are updated in the correct order.

We should protect AddEvents with a mutex or channels. So events from one block cannot be processed until events from the previous block are finished. We also do not want to slow down the chain by waiting for events to be processed. So, add an intermediate layer between the block calling AddEvents and events being processed; where execution can wait while events from earlier blocks are handled.

1. Start new event processing worker thread in [NewEventDb](https://github.com/0chain/0chain/blob/staging/code/go/0chain.net/smartcontract/dbs/event/eventdb.go#L8). 
2. Have the 0chain.Block thread pass the events for a block to an `event processing channel`, before continuing with the chain.
3. The event database worker thread from (1) continually polls the `event processing channel` for the next event, then processes it.

samtholiya:
I am not using channels but using a mutex.
The reason is using a channel still does not guarantee the processing of blocks in order if its buffer is overloaded.
```golang
package main

import (
	"fmt"
	"time"
)

func Do(a chan int) {
	more := true
	for more {
		s, more := <-a
		if !more {
			return
		}
		time.Sleep(1 * time.Second)

		// Printed after sleep is over
		fmt.Println(s)
	}
}

func main() {
	send := make(chan int, 2)
	go Do(send)

	for i := 0; i < 1000000; i++ {
		go func(j int) {
			send <- j
		}(i)
	}
	time.Sleep(1000 * time.Second)
}

```
Now mutex does not solve this problem but helps process the events for one block at a time.
Yes channels could achieve the same but. it would be adding one goroutine that is not necessary and a channel memory that is not necessary as well. A mutex can achieve the same functionality and still keep the code simple

kushthedude:
Wont the mutex block the queue for all events which are coming to be processed @samtholiya ?

Sriep:
[...]

PR #836, opened 2021-12-29
removeDuplicate query optimized to partially fix #799

### LINKED ISSUE #799: Event db. Improve remove duplicates.
We want to avoid acting on the same event twice. The current method, [removeDuplicate](https://github.com/0chain/0chain/blob/staging/code/go/0chain.net/smartcontract/dbs/event/event.go#L75), is inefficient as it queries all events. 

Especially in conjunction with https://github.com/0chain/0chain/issues/756 we can assume that blocks will not be repeated. So we should only need to search events for one block for repeats.

samtholiya:
- [x]  Do any new system tests need added to test this change? do any existing system tests need updated? If so create a branch at [0chain/system_test](https://github.com/0chain/system_test) then add and merge your test changes to master BEFORE merging this PR


samtholiya:
```
zchain_unit_test_1  | ?         0chain.net/smartcontract/benchmark/main/cmd/log [no test files]
zchain_unit_test_1  | ?         0chain.net/smartcontract/dbs    [no test files]
zchain_unit_test_1  | ok        0chain.net/smartcontract/dbs/event      0.480s  coverage: 29.4% of statements
zchain_unit_test_1  | ?         0chain.net/smartcontract/dbs/mocket     [no test files]
zchain_unit_test_1  | ?         0chain.net/smartcontract/dbs/postgresql [no test files]
zchain_unit_test_1  | ?         0chain.net/smartcontract/faucetsc       [no test files]
```

Event query passed

kushthedude (commented):
The query is faster but I dont see if there would be any changes in the result expected of it 🤔 ?

Sriep (approved):
.

Sriep (changes requested):
This still searches a potentially large table looking for a duplicate. Has this been benchmarked for extremely large events tables? I am not convinced this will solve the reported problem of the event database taking too long on mature chains.

Can't we just keep a local list of events from the current block and search that? Events are received in order https://github.com/0chain/0chain/issues/757, so if an event for another block is processed we can drop the list for the previously current block and start with the new block.

This way we only search hundreds of events from one block rather than millions or billions of events across the chain's history.

code/go/0chain.net/smartcontract/dbs/event/event.go
```diff
@@ -61,16 +62,16 @@ func (edb *EventDb) GetEvents(ctx context.Context, block int64) ([]Event, error)
 }
 
 func (edb *EventDb) exists(ctx context.Context, event Event) (bool, error) {
-	var count int64
-	result := edb.Store.Get().WithContext(ctx).
-		Model(&Event{}).
-		Where("tx_hash = ? AND index = ?", event.TxHash, event.Index).
-		Count(&count)
+	var ev Event
+	result := edb.Store.Get().WithContext(ctx).Model(&Event{}).Where(&Event{TxHash: event.TxHash, Index: event.Index}).Take(&ev)
```
ssardana96: shouldn't we be checking type and tag here as well @Sriep

ssardana96 (changes requested):

code/go/0chain.net/smartcontract/dbs/event/event.go
samtholiya: done

peterlimg:
`removeDuplicate` should not be needed once https://github.com/0chain/0chain/issues/993 is done.

samtholiya:
as @peterlimg told this pr is not needed therefore closing this pr

PR #839, opened 2021-12-29
Fix for #757 added channels to process one block at a time

### LINKED ISSUE #757: Event db: Ensure events processed in correct order.

samtholiya:
  
...


kushthedude (commented):
Wont the mutex block the queue for all events which are coming to be processed @samtholiya ?

samtholiya:
> Wont the mutex block the queue for all events which are coming to be processed @samtholiya ?
image

@kushthedude 
The places where addevents is used all execute it as a goroutine. ( We can improve the code so that AddEvents itself create a goroutine)
So the parent function will not be blocked due to this...

Sriep:
Could you check https://github.com/0chain/0chain/issues/757#issuecomment-1004040827 and see if this would work?

kushthedude:
> Could you check [#757 (comment)](https://github.com/0chain/0chain/issues/757#issuecomment-1004040827) and see if this would work?

Mutex is better preferred to provide locking mechanism to synchronise the access to any resources, the worker thread process seems more like a semaphore & signalling mechanism which can also fix this but require more resource to keep track of each wait, signal. Why do we need to proceed with signalling mechanism @Sriep, do you see any problem here which is better solved by worker thread ?

Sriep:
Ok maybe mutex is simpler, how would either solution work if the event database went down for a few seconds? Can mutex guarantee first in first out? https://stackoverflow.com/questions/51140111/is-sync-mutex-lock-a-fifo

samtholiya:
@Sriep if the database goes down the channel buffer can go full so in that case even the channel acts like mutex and cannot guarantee first in first out. Refer this: https://github.com/0chain/0chain/issues/757#issuecomment-1002809781

Sriep:
@samtholiya  

#### Channel https://github.com/0chain/0chain/issues/757#issuecomment-1004040827

If the event database is held up, then the intermediate worker thread will maintain a growing slice of events read from the ChainToWorker channel. Once the WorkerToEventD channel is no longer full, the backlog can be sent to the now with free space WorkerToEventD channel. This should ensure that events are processed in the correct order, `which is the point of` https://github.com/0chain/0chain/issues/757.

#### Mutex 

I do not see at the moment how using a mutex ensures events are processed in `first in first out` order.

#### Related changes.

New issue (coming) where we will also send an event hash for each block. Events for the latest block will need to be kept locally so they can be hashed and compared with the block event hash. As the events are coming in order it should be easy enough to compare the hash values.

samtholiya:
@Sriep yes I do agree mutex does not handle FIFO
And I like your solution.

I would like to know what should be the breaking point.

I have once faced an issue previously where we did something similar and we found the ram consumption to be increasing exponentially in case the Database is not available for a bit longer in comparison to the number of events generated.
So there should be a breakpoint and post that something should happen... no idea what. Maybe you can help me here?

code/go/0chain.net/chaincore/block/entity.go
```diff
@@ -995,9 +993,7 @@ func (b *Block) ComputeStateLocal(ctx context.Context, c Chainer) error {
 	}
 
 	if len(b.Events) > 0 && c.GetEventDb() != nil {
-		go func() {
-			c.GetEventDb().AddEvents(ctx, b.Events)
-		}()
+		c.GetEventDb().AddEvents(ctx, b.Events)
 		b.Events = nil
```
Sriep, line 1031: Since AddEvents is now returning an error we should log it. 
`logging.Logger.Error("..."), zap.Error(err))`

code/go/0chain.net/chaincore/block/entity.go
```diff
@@ -905,9 +905,7 @@ func (b *Block) ComputeState(ctx context.Context, c Chainer) error {
 	}
 
 	if len(b.Events) > 0 && c.GetEventDb() != nil {
-		go func() {
-			c.GetEventDb().AddEvents(ctx, b.Events)
-		}()
+		c.GetEventDb().AddEvents(ctx, b.Events)
```
Sriep, line 923: Since AddEvents is now returning an error we should log it.

code/go/0chain.net/smartcontract/dbs/event/eventdb.go
```diff
@@ -14,17 +20,79 @@ func NewEventDb(config dbs.DbAccess) (*EventDb, error) {
...
+}
+
 type EventDb struct {
 	dbs.Store
+	eChannel       chan eventCtx
+	eBufferChannel chan eventCtx
+}
+
+// helps maitaining dynamic buffer for addEventWorker
+func (edb EventDb) channelBufferIntermediate() {
```
Sriep: A unit test would be nice. You could generate some semi-random events. Start channelBufferIntermediate. Send your test events to eBufferChannel. Monitor eChannel and confirm it receives events in the correct order.

Sriep (commented):
For manual integartion testing you could add a temporty log to ComputeStateLocal and addEventWorker. And check events are appearing correctly.

code/go/0chain.net/smartcontract/dbs/event/eventdb.go
samtholiya: UT brought out an edge case where the goroutine is not scheduled and the channel buffer is full
in this case, the order is still not guaranteed. 
Also, why cannot we use a queue service like kafka or rabbitmq to achieve this? Yes, it will increase our requirement but if that is what we want to achieve this is the only solution.
If this was a Kubernetes-based solution. I would have recommended using new-relic to log all the events.
Please read the following golang issue where someone tried to do this
https://github.com/golang/go/issues/20352

Sriep: I don't follow the edge case you refer to, maybe you could commit a unit test that exposes the problem so we can see what you mean.

dabasov: Semaphores, mutexes, monitors, etc, in general knows nothing about data structure they serialize access to. channels is something different, it's more like data structure and semaphore together.

As I got your problem right, you can serialize access to a slice with mutext and implement FIFO logic with this slice easily, you can do queue with two slices (copy on empty) it will be more performant in some cases.
What operations you expect to prevail, reads or writes?

samtholiya: @dabasov no the issue is if the goroutine (that appends to the slice) does not get scheduled and the buffer for channel is full
Then the events are not serialized
cc: @Sriep

Sriep:
```go
	eventDb := &EventDb{
		Store:          db,
		eBufferChannel: make(chan eventCtx, 100),
		eChannel:       make(chan eventCtx, 100),
	}
```
Why is eBufferChannel buffered. sould it not read `eBufferChannel: make(chan eventCtx),`
```go
	edb.eBufferChannel <- eventCtx{
		ctx:    ctx,
		events: events,
	}
}
```
Don't we want this to be synchronous? We are only adding the events to `buf` so this should not hold up the chain. 

Making it `eBufferChannel` buffered raises the issue if the buffer gets full. Also, we should probably change the name as we don't want it buffered.

For the moment I think we can skip the issue with running out of memory with an oversized `buf` object if the event database gets stuck. We can relook if it becomes an issue.

code/go/0chain.net/chaincore/block/entity.go
dabasov: why is it done in blocking manner now without goroutine?

code/go/0chain.net/smartcontract/dbs/event/eventdb.go
```diff
@@ -15,17 +20,81 @@ func NewEventDb(config dbs.DbAccess) (*EventDb, error) {
...
-		Store: db,
+		Store:          db,
+		eBufferChannel: make(chan eventCtx, 100),
+		eChannel:       make(chan eventCtx, 100),
 	}
 
 	if err := eventDb.AutoMigrate(); err != nil {
 		return nil, err
 	}
+	go eventDb.channelBufferIntermediate()
```
dabasov: this should be run with common.GetRootContext()

code/go/0chain.net/smartcontract/dbs/event/eventdb.go
```diff
@@ -15,17 +20,81 @@ func NewEventDb(config dbs.DbAccess) (*EventDb, error) {
...
+	go eventDb.addEventWorker()
```
dabasov, line 30: this should be run with common.GetRootContext()

code/go/0chain.net/smartcontract/dbs/event/eventdb.go
```diff
@@ -15,17 +20,81 @@ func NewEventDb(config dbs.DbAccess) (*EventDb, error) {
...
+	buf := make([]eventCtx, 0)
+	for {
+		events, ok := <-edb.eBufferChannel
+		for _, e := range buf {
+			select {
+			case edb.eChannel <- e:
+			default:
+				break
+			}
+			buf = buf[1:]
```
dabasov: I don't think that it is a good idea to drop events without any notification here, at least logging

code/go/0chain.net/smartcontract/dbs/event/eventdb.go
```diff
@@ -15,17 +20,81 @@ func NewEventDb(config dbs.DbAccess) (*EventDb, error) {
 		return nil, err
 	}
 	eventDb := &EventDb{
-		Store: db,
+		Store:          db,
+		eBufferChannel: make(chan eventCtx, 100),
+		eChannel:       make(chan eventCtx, 100),
```
dabasov: what is an idea behind this second channel?

code/go/0chain.net/smartcontract/dbs/event/eventdb.go
```diff
@@ -15,17 +20,81 @@ func NewEventDb(config dbs.DbAccess) (*EventDb, error) {
...
+	}
+}
+
+// addEventWorker this worker will try to add events unless thery are not added.
+func (edb EventDb) addEventWorker() {
+	for {
+		events := <-edb.eChannel
+		for {
+			newEvents := edb.removeDuplicate(events.ctx, events.events)
+			if err := edb.addEvents(events.ctx, newEvents); err != nil && !strings.Contains(err.Error(), "len(events):0") {
```
dabasov, line 50: it might be not a good idea to pass context via event, should we write events to db if this context passed in for example computeState() got cancelled or timeout? I think yes, we better use some new context with timeout here, just to prevent long db connections

code/go/0chain.net/smartcontract/dbs/event/eventdb.go
samtholiya: This was an experiment I and @Sriep were trying ... changing that.. 😓

code/go/0chain.net/smartcontract/dbs/event/eventdb.go
samtholiya: I don't know where that event comes from ... but there are some questions
Is timeout the only context cancel reason? If yes then that can be done,
But if there are context cancellations due to other reasons how should we handle that?

dabasov: is there any reason for this context cancellation other than timeout?

samtholiya: I did not trace but we do use context.WithCancel
image
at some places 
So what if those cancellations are encountered.
Having a disconnected context will make us miss those cancellations

dabasov: unfortunately context is not equal to transaction. we wish to have single transactional scope for block and for event, I mean when block computation is cancelled than to cancel event writing and if block is committed (there is no such a concept yet) then commit transaction in event db. 

I agree that to ensure data consistency we have to implement something similar to distributed transaction, to be honest we don't want to use one since it will be overkill, we might:
- implement some compensation logic for cases when block state computation is cancelled, for example to generate appropriate event and to mark all generated events previously as cancelled
- we can replace event in your solution with some channel (semaphore) to start event writing after it 

the problem with your solution is that we *can* cancel the event writing if context will be cancelled before the writing, but it is not strictly

code/go/0chain.net/chaincore/block/entity.go
samtholiya: errors logged. 
>why is it done in a blocking manner now without goroutine?

We are actually using the buffered channel. Which if full could block I agree but we have a buffer of 10000 as of now. Are the events going to exceed that in a short time?

code/go/0chain.net/smartcontract/dbs/event/eventdb.go
samtholiya: done

samtholiya:
Old branch please refer new https://github.com/0chain/0chain/pull/968

PR #846, opened 2022-01-03
#771: error table events database

### LINKED ISSUE #771: Event db: Error table
A new Error table for the event database, for events of type `TypeError`. This stores the error for any failed transaction.

| Error table   | type   | notes                              |
|---------------|--------|------------------------------------|
| TransactionId | uint   | foreign key into Transaction table |
| Error         | string |                                    |

Blocker for https://github.com/0chain/gosdk/issues/326

samtholiya:
  
...


code/go/0chain.net/smartcontract/dbs/event/process_test.go
```diff
@@ -0,0 +1,42 @@
...
+	"context"
+	"os"
+	"testing"
+	"time"
+
+	"0chain.net/smartcontract/dbs"
+)
+
+func TestAddEvents(t *testing.T) {
+	access := dbs.DbAccess{
```
Sriep, line 13: Maybe we should stop using active databases in unit tests. Instead, use mocket to confirm events lead to the correct SQL querries.

Sriep (dismissed):
Looks ok. Can merge as no backward compatibility issues.

Consider adding API method in conjunction with GoSDK issue https://github.com/0chain/gosdk/issues/326. (could add a new error API endpoint here or as part of Gosdk326.)

code/go/0chain.net/smartcontract/dbs/event/process_test.go
samtholiya: This test will be skipped as soon as it realizes there is no database to connect.
We need a better pattern for unit and integration testing which will help us running unit test vs integration test. 

I feel integration testing are equally important
Creating a unit test case using mocket is just like we are testing gorm which we do not intend to do here.
We intend to understand if the query works as per our expectations.

kushthedude: Unit test is going to ensure the query is being called, and the code execution goes smooth if Gorm works as expected which is why you might think we are testing gorm, but we actually want to ensure that the queries & in the order which they are being written should be the order they invoked in when running the code. 
    
What do you think, can we also add unit test?

samtholiya: Order of the query 🤔
We are just executing one query.

code/go/0chain.net/smartcontract/dbs/event/process.go
```diff
@@ -45,6 +45,11 @@ func (edb *EventDb) AddEvents(ctx context.Context, events []Event) {
 		switch EventType(event.Type) {
 		case TypeStats:
 			err = edb.addStat(event)
+		case TypeError:
```
ssardana96, line 60: Where is this being used?

ssardana96 (changes requested):

code/go/0chain.net/smartcontract/dbs/event/process.go
samtholiya: EmitError is called in the following lines
https://github.com/0chain/0chain/blob/staging/code/go/0chain.net/chaincore/chain/state.go#L202
https://github.com/0chain/0chain/blob/staging/code/go/0chain.net/chaincore/chain/state.go#L213

Which Eventually adds this to events array
https://github.com/0chain/0chain/blob/staging/code/go/0chain.net/chaincore/chain/state/state_context.go#L198

code/go/0chain.net/smartcontract/dbs/event/error.go
```diff
@@ -0,0 +1,10 @@
+package event
+
+type Error struct {
```
ssardana96, line 5: Gorm.model

code/go/0chain.net/smartcontract/dbs/event/process.go
ssardana96: Did we test changes on local chain post integration?

ssardana96 (changes requested):

code/go/0chain.net/smartcontract/dbs/event/error.go
```diff
@@ -0,0 +1,10 @@
...
+	TransactionID string
+	Error         string
+}
+
+func (edb *EventDb) addError(err Error) error {
```
ssardana96, line 11: @Sriep @kushthedude Don’t we need logic to remove duplicates here?
Or overwrite records ?

ssardana96 (changes requested):

code/go/0chain.net/smartcontract/dbs/event/process_test.go
Sriep: The unit test should be testing our code, not gorm or PostgreSQL which can be assumed to work. Hence we use mocket to test that we send the SQL statement we expect to. 

There is no need to test everything every test, just our added code.

code/go/0chain.net/smartcontract/dbs/event/error.go
Sriep: We want to avoid duplicates, however, error rows should be immutable so no overwriting. Removing duplicates should be handled elsewhere https://github.com/0chain/0chain/issues/799.

code/go/0chain.net/smartcontract/dbs/event/process.go
samtholiya: image
Yep ... 
But not sure how the errors happened.... 😅. Just started the chain to encounter these error

Will add gorm.Model
But as the ticket mentioned just two columns did not add it.

code/go/0chain.net/smartcontract/dbs/event/process_test.go
samtholiya: @Sriep 
I tried to find there is no place where we are using mocket (Except we created a struct and did not use it).
Also, the mocket documentation did not show how to assert a query. I tried to use the mocket struct in the repo but did not find a way to use it correctly... 

Can you please share with me how to do that?

Because 
>  to test that we send the SQL statement we expect to.

We need to assert the query somewhere but i cannot find that.

code/go/0chain.net/smartcontract/dbs/event/process.go
samtholiya: image
this is after adding gorm.Model.

code/go/0chain.net/smartcontract/dbs/event/error.go
samtholiya: done

code/go/0chain.net/smartcontract/dbs/event/eventdb.go
```diff
@@ -37,7 +37,8 @@ func (edb *EventDb) AutoMigrate() error {
 		&Transaction{},
 		&Validator{},
 		&ReadMarker{},
-    &Block{}); err != nil {
+    &Block{},
```
kushthedude: Can we fix indents here?

dabasov:
be aware that we will include transactions that cause error into the block, for more details check #887

code/go/0chain.net/smartcontract/dbs/event/eventdb.go
samtholiya: done

code/go/0chain.net/smartcontract/dbs/event/process_test.go
Sriep: Yes, we need to start using [mocket](https://github.com/Selvatico/go-mocket/blob/master/DOCUMENTATION.md).

Here is an example I did for blobber https://github.com/0chain/blobber/blob/staging/code/go/0chain.net/blobbercore/handler/object_operation_handler_test.go

There is also sqlmocket, which has more features but does not seem to be fully compatible with gorm https://github.com/DATA-DOG/go-sqlmock/issues/118, https://stackoverflow.com/questions/50164399/sqlmock-with-gorm-insert.

Sriep:
```
be aware that we will include transactions that cause error into the block, for more details check #887
```
We should not keep any events emitted from a failed transaction. We just need to ensure the transaction is added to the transaction table (not necessarily here) and the error to the error table.

samtholiya:
image
added endpoint for errors.

code/go/0chain.net/smartcontract/dbs/event/error.go
```diff
@@ -0,0 +1,18 @@
...
+	return edb.Store.Get().Create(&err).Error
+}
+
+func (edb *EventDb) GetErrorByTransactionHash(transactionID string) ([]Error, error) {
+	var err []Error
```
ssardana96: probably a better variable name for error data from db
and not err. err is generally used for go based errors.

code/go/0chain.net/smartcontract/storagesc/sc.go
```diff
@@ -120,6 +120,7 @@ func (ssc *StorageSmartContract) setSC(sc *sci.SmartContract, bcContext sci.BCCo
 	ssc.SmartContractExecutionStats["/get_transaction"] = ssc.GetTransactionByHashHandler
 	ssc.SmartContractExecutionStats["/get_block_by_hash"] = ssc.GetBlockByHashHandler
 	ssc.SmartContractExecutionStats["/get_blocks"] = ssc.GetBlocksHandler
+	ssc.SmartContract.RestHandlers["/errors_by_transaction_hash"] = ssc.GetErrorByTransactionHash
```
ssardana96: just have an API /get_errors
as you're already taking `transaction_hash` from params. 
That is self descriptive

ssardana96 (dismissed):

dabasov:
> ```
> be aware that we will include transactions that cause error into the block, for more details check #887
> ```
> 
> We should not keep any events emitted from a failed transaction. We just need to ensure the transaction is added to the transaction table (not necessarily here) and the error to the error table.

this transaction is not failed, it is successful, since it is added to the block and fee is charged

code/go/0chain.net/smartcontract/dbs/event/error.go
samtholiya: done

code/go/0chain.net/smartcontract/storagesc/sc.go
samtholiya: done
made inline with recent transaction api discussion

peterlimg (approved):
LGTM

PR #853, opened 2022-01-04
Fix #759 added transaction table

### LINKED ISSUE #759: Event db: Transaction table
We need to add a Transaction table to the Event db. This is shown below with the location of the data.

Currently, we send events to the event database in [ComputeStateLocal](https://github.com/0chain/0chain/blob/staging/code/go/0chain.net/chaincore/block/entity.go#L920). At this point we have the array of transactions for the block in the block object, so it should be easy to emit an event for each of the blocks transactions.

As well as a new `Transaction` table as below we will want an addTransaction method to be called when calling [addStat](https://github.com/0chain/0chain/blob/staging/code/go/0chain.net/smartcontract/dbs/event/process.go#L52). Plus a new  `TagAddTransaction`.

A new REST API method that returns details of a  transaction from a transaction hash. Needs to work with [Explorer](https://github.com/0chain/block-explorer) which will be using this endpoint. Explorer might also want a join with other tables, so we might need to wait here until those tables are available.

| Event database Transaction  | type      | 0chain Transaction.  | type   |
|---------------------|-----------|----------------------|--------|
| Hash                | string    | HashIDField.Hash     | Key    |
| BlockHash           | string    | parent object        |        |
| Version             | string    | VersionField,Version | string |
| ClientId            | string    | ClientID             | Key    |
| ToClientId          | string    | ToClientID           | Key    |
| TransactionData     | string    | TransactionData      | string |
| Value               | int64     | Value                | int64  |
| Signature           | string    | Signature            | string |
| CreationDate        | int64     | CreationDate         | int64  |
| Fee                 | int64     | Fee                  | int64  |
| TransactionType     | int       | TransactionType      | int    |
| TransactionOutput   | string    | TransactionOutput    | string |
| OutputHash          | string    | OutputHash           | string |
| Status              | string    | Status               | string |
| CreatedAt           | time.Time | CreationDate         | int64  |

A new `Nonce int64` field will be added in https://github.com/0chain/0chain/issues/765.

samtholiya:
  
...


kushthedude (commented):
Where is event being emmitted in this PR?

samtholiya:
@kushthedude where should be the event emitted. I thought this https://github.com/0chain/0chain/pull/853/files#diff-c976011a57a7753982a71bbc00699fea26b903d5176c9434c844a0fc882e9275R917 is the place where @Sriep mentioned

code/go/0chain.net/smartcontract/dbs/event/process.go
```diff
@@ -27,6 +27,7 @@ const (
 	TagAddOrOverwriteBlobber
 	TagUpdateBlobber
 	TagDeleteBlobber
+	TagAddTransaction
```
kushthedude, line 30: Where is this being used? According to current structure, there is no TagAddTransaction being emitted

code/go/0chain.net/chaincore/block/entity.go
```diff
@@ -944,6 +951,29 @@ func (b *Block) ComputeState(ctx context.Context, c Chainer) error {
 	return nil
 }
 
+func (b *Block) addTransactions(c Chainer) {
```
kushthedude: You need to remove this & add something like emitAddTransaction, see this code block by @ssardana08 for miner https://github.com/0chain/0chain/blob/ffcf5a425aa7d917a68fee18b7363bd7ae0d7d26/code/go/0chain.net/smartcontract/minersc/sharder_eventdb.go#L76 , something similar needs to be done for transactions.

code/go/0chain.net/smartcontract/dbs/event/process.go
```diff
@@ -68,6 +69,13 @@ func (edb *EventDb) addStat(event Event) error {
 		return edb.updateBlobber(updates)
 	case TagDeleteBlobber:
 		return edb.deleteBlobber(event.Data)
+	case TagAddTransaction:
+		var transaction Transaction
+		err := json.Unmarshal([]byte(event.Data), &transaction)
+		if err != nil {
+			return err
+		}
+		return edb.AddTransaction(transaction)
```
kushthedude: This should be called from the above block where I have commented which shall add the transaction in the db

code/go/0chain.net/smartcontract/dbs/event/transaction.go
```diff
@@ -0,0 +1,37 @@
...
+	CreationDate      int64
+	Fee               int64
+	TransactionType   int
+	TransactionOutput string
+	OutputHash        string
+	Status            int
+}
+
+// AddTransaction Saves the transaction to the event db
+func (edb *EventDb) AddTransaction(transaction Transaction) error {
```
ssardana96: it should be non-exportable function

ssardana96 (changes requested):

code/go/0chain.net/smartcontract/dbs/event/transaction.go
samtholiya: done

code/go/0chain.net/smartcontract/dbs/event/process.go
samtholiya: done

code/go/0chain.net/chaincore/block/entity.go
samtholiya: done... discussed and a different approach was decided and implemented

code/go/0chain.net/smartcontract/dbs/event/process.go
samtholiya: done... discussed and a different approach was decided and implemented

code/go/0chain.net/chaincore/block/entity.go
```diff
@@ -1033,6 +1047,26 @@ func (b *Block) ComputeStateLocal(ctx context.Context, c Chainer) error {
 	return nil
 }
 
+func transactionNodeToEventTransaction(tr *transaction.Transaction, blockHash string) event.Transaction {
```
kushthedude, line 1050: Can we move this helper function to `event/trasnaction.go` ?

kushthedude (commented):
A small neat

code/go/0chain.net/chaincore/block/entity.go
samtholiya: No because doing that we are doing this in dependency tree
chiancore imports  smartcontract
smartcontract imports chaincore


This is not recommended

kushthedude: Cyclic imports, got it.

Sriep (approved):
It looks fine. There are no backward compatibility issues. 

These changes need to be integration tested with upcomming explorer changes. Check with @shaktals.

ssardana96:
lgtm

ssardana96 (approved):

PR #874, opened 2022-01-12
Fix #857 Make versioned api

### LINKED ISSUE #857: Choose REST API endpoint by genesis event database version
For backward compatibility, we will need to choose between different endpoints depending on the build history of the chain.

Add a new immutable genesisEventDbVersion to global settings. Initialise this to a hardcoded value during genesis or the first block. Mayne in [miner.main](https://github.com/0chain/0chain/blob/staging/code/go/0chain.net/miner/miner/main.go#L47) and [sharder.main](https://github.com/0chain/0chain/blob/staging/code/go/0chain.net/sharder/sharder/main.go#L94).

Somewhere around [ExecuteRestAPI](https://github.com/0chain/0chain/blob/staging/code/go/0chain.net/chaincore/smartcontract/handler.go#L24) we will need to get the start version of the chain. A StateContextI object is one of the EcecuteRestAPI input parameters so should be simple to get the EventDbVersion

[SmartContractInterface.GetRestPoints](https://github.com/0chain/0chain/blob/staging/code/go/0chain.net/chaincore/smartcontractinterface/interface.go#L38) will need to be modified to take a version number parameter and return the endpoint list suitable for that version number. 

This will require modification to all smart contracts`GetRestPoint` implementations to support this change.
```go
type SmartContractInterface interface {
	Execute(t *transaction.Transaction, funcName string, input []byte, balances c_state.StateContextI) (string, error)
	GetRestPoints(version int) map[string]SmartContractRestHandler
	...
}
```

This can be considered a blocker for merging most Event Db changes.

We might also want to consider https://github.com/0chain/0chain/issues/796 in conjunction with this change. Which would mean splitting the `SmartContractInterface` into a `SmartContractInterface` and a `RestApiInterface`.

samtholiya:
  
...

### DIFF
--- a/code/go/0chain.net/chaincore/chain/state_handler.go
+++ b/code/go/0chain.net/chaincore/chain/state_handler.go
@@ -162,7 +162,7 @@ func (c *Chain) GetSCRestPoints(w http.ResponseWriter, r *http.Request) {
 	PrintCSS(w)
 	fmt.Fprintf(w, "<table class='menu' style='border-collapse: collapse;'>")
 	fmt.Fprintf(w, "<tr class='header'><td>Function</td><td>Link</td></tr>")
-	restPoints := scInt.GetRestPoints()
+	restPoints := scInt.GetRestPoints(0)
 	names := make([]string, 0, len(restPoints))
 	for funcName := range restPoints {
 		names = append(names, funcName)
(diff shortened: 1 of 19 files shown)


kushthedude (changes requested):
https://0chain.slack.com/archives/C02QDMYFQJG/p1641999908099800?thread_ts=1641947468.094200&cid=C02QDMYFQJG

Changes are more complicated than adding versionaning to SC REST APIs

samtholiya:
This pr is old and this is not the write way to do this exactly need to do it in a better way

PR #875, opened 2022-01-12
Fix emitted events were not getting added

ssardana96:
  
...

### DIFF
--- a/code/go/0chain.net/chaincore/block/entity.go
+++ b/code/go/0chain.net/chaincore/block/entity.go
@@ -907,9 +907,9 @@ func (b *Block) ComputeState(ctx context.Context, c Chainer) error {
 	}
 
 	if len(b.Events) > 0 && c.GetEventDb() != nil {
-		go func() {
-			c.GetEventDb().AddEvents(ctx, b.Events)
-		}()
+		go func(events []event.Event) {
+			c.GetEventDb().AddEvents(ctx, events)
+		}(b.Events)
 		b.Events = nil
 	}
 
@@ -1009,9 +1009,9 @@ func (b *Block) ComputeStateLocal(ctx context.Context, c Chainer) error {
 	}
 
 	if len(b.Events) > 0 && c.GetEventDb() != nil {
-		go func() {
-			c.GetEventDb().AddEvents(ctx, b.Events)
-		}()
+		go func(events []event.Event) {
+			c.GetEventDb().AddEvents(ctx, events)
+		}(b.Events)
 		b.Events = nil
 	}
 


code/go/0chain.net/chaincore/block/entity.go
```diff
@@ -907,9 +907,9 @@ func (b *Block) ComputeState(ctx context.Context, c Chainer) error {
 	}
 
 	if len(b.Events) > 0 && c.GetEventDb() != nil {
-		go func() {
-			c.GetEventDb().AddEvents(ctx, b.Events)
-		}()
+		go func(events []event.Event) {
```
kushthedude, line 910: How does this change the code execution process?

kushthedude (changes requested):

code/go/0chain.net/chaincore/block/entity.go
ssardana96: we're passing the global context of b.Events to anonymous function
and just after that `b.Events = nil` happens
that's where the order of execution comes into picture

which is hence calling AddEvents will nil slice.

ssardana96: @kushthedude so I tested on local chain for the same

```
2022-01-12T06:45:13.554Z        INFO    block/entity.go:937     compute state successful        {"round": 17, "block": "433ff599bf7153d101c08
8c5c2c5d115aa7306e9afd4e5ddb8902757c2f025c8", "block_size": 2, "changes": 18, "begin_client_state": "06aef01990456649b022c09d150f04cb26bb8f9d
af5d16ad96f044a61c3cd5c9", "computed_state_hash": "1cb9ab798e36dadbf5e210a134e1e7df6ff575916b07b9f4b69bb151c467802c", "block_state_hash": "1c
b9ab798e36dadbf5e210a134e1e7df6ff575916b07b9f4b69bb151c467802c", "prev_block": "3b3af3a399a028b5cf6c4e6e39eea1e67fd9ce498d349ce2b120b90bfa0a0
357", "prev_block_client_state": "06aef01990456649b022c09d150f04cb26bb8f9daf5d16ad96f044a61c3cd5c9"}
2022-01-12T06:45:13.554Z        INFO    event/process.go:34     inside add_events       {"events": null}
```

that's how the flow was working.
ComputeState was getting events emiited
but when AddEvents was reading it, it got nil.
Because of which blobbers and other tables were not getting data.

Hence we should pass a local copy of that variable to anonymous goroutine function before it is nullified.

code/go/0chain.net/chaincore/block/entity.go
```diff
@@ -907,9 +907,9 @@ func (b *Block) ComputeState(ctx context.Context, c Chainer) error {
...
+			c.GetEventDb().AddEvents(ctx, events)
+		}(b.Events)
 		b.Events = nil
```
ssardana96, line 913: @kushthedude that's where it's getting nil :)

ssardana96:
@kushthedude to understand what's happening here.
Check this: https://go.dev/play/p/SAtP9-Foa1Y

code/go/0chain.net/chaincore/block/entity.go
dabasov: yeah, the next line it is assigned to nil, nice data race, good catch @ssardana08!

dabasov (approved):
lgtm

kushthedude (approved):

PR #888, opened 2022-01-14
Hotfix to remove DbAccess settings from MPT

Sriep:
  
...

### DIFF
--- a/code/go/0chain.net/chaincore/chain/config.go
+++ b/code/go/0chain.net/chaincore/chain/config.go
@@ -1,11 +1,12 @@
 package chain
 
 import (
+	"sync"
+	"time"
+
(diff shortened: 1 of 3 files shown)


kushthedude (commented):
Does this mean we will get the DB access creds from the sc.yaml config file passed during deployment itself?

Sriep:
This should prevent saving and loading EventDb logging details from MPT. You also cannot change anymore from an `update_global_settings` transaction only by editing the `0chain.yaml` and restarting the miner. The logging details are therefore defined on a sharder by sharder case rather than forced to the same for the whole chain.

kushthedude (approved):
Looks good to me 👍🏽

peterlimg (approved):
LGTM

PR #889, opened 2022-01-14
Add block eventsdb

### LINKED ISSUE #758: Event db Block table
We need to add a block table to the Event database. Data is available from the 0chain [Block](https://github.com/0chain/0block/blob/staging/code/go/0block/zblockcore/models/block.go#L22) object.

Currently, events are sent to the event database by the sharder block object in [ComputeStateLocal](https://github.com/0chain/0chain/blob/staging/code/go/0chain.net/chaincore/block/entity.go#L920). We will want to ensure the block object is emitted after the last time it can change. So we will want to confirm this is a suitable place.

As well as a new `Block` table as below we will want an addBlock method to be called when calling [addStat](https://github.com/0chain/0chain/blob/staging/code/go/0chain.net/smartcontract/dbs/event/process.go#L52). Plus a new  `TagAddBlock`.

A new REST API method that returns the block table. Needs to work with [Explorer](https://github.com/0chain/block-explorer) which will be using this endpoint.

| Event Db Block fields         | 0block type | 0chain Block.?                                     | type    |
|-----------------------|-------------|----------------------------------------------------|---------|
| Hash                  | string      | HashIDField.Hash                                   | Key     |
| Version               | string      | UnverifiedBlockBody.VersionField.Version           | string  |
| CreationDate          | int64       | UnverifiedBlockBody.CreationDateField.CreationDate | int64   |
| Round                 | int64       | UnverifiedBlockBody.Round                          | int64   |
| MinerId               | string      | UnverifiedBlockBody.MinerId                        | Key     |
| RoundRandomSeed       | int64       | UnverifiedBlockBody.RoundRandomSeed                | int64   |
| MerkleTreeRoot        | string      |  ClientState.GetRoot()                        | Key     |
| StateHash             | string      |  UnverifiedBlockBody.ClientStateHash          | Key     |
| ReceiptMerkleTreeRoot | string      |       block.GetReceiptsMerkleTree().GetRoot()         |    string     |
| NumTxns               | int       |  len(UnverifiedBlockBody.Txn)                 | int     |
| MagicBlockHash        | string      | MagicBlock.HashIDField.Hash                        | Key     |
| PrevHash              | string      | UnverifiedBlockBody.PrevHash                       | string  |
| Signature             | string      | Signature                                          | string  |
| ChainId               | string      | ChainID                                            | Key     |
| ChainWeight           | string      | ChainWeight                                        | float64 |
| RunningTxnCount       | string      | RunningTxnCount                                    | int64   |
| RoundTimeoutCount     | int         | UnverifiedBlockBody.RoundTimeoutCount              | int     |
| CreatedAt             | time.Time   | UnverifiedBlockBody.CreationDateField.CreationDate | int64   |

amanjain97:
  
...

### DIFF
--- a/code/go/0chain.net/chaincore/block/block_eventdb.go
+++ b/code/go/0chain.net/chaincore/block/block_eventdb.go
@@ -11,7 +11,7 @@ func blockToBlockEvent(block *Block) *event.Block {
 	return &event.Block{
 		Hash:                  block.Hash,
 		Version:               block.Version,
-		CreationDate:          block.CreationDate.Duration().Nanoseconds(),
+		CreationDate:          int64(block.CreationDate.Duration()),
 		Round:                 block.Round,
 		MinerID:               block.MinerID,
 		RoundRandomSeed:       block.RoundRandomSeed,
--- a/code/go/0chain.net/chaincore/block/block_eventdb.go
+++ b/code/go/0chain.net/chaincore/block/block_eventdb.go
@@ -31,22 +31,17 @@ func blockToBlockEvent(block *Block) *event.Block {
 }
 
 func emitBlockEvent(block *Block) error {
-	if len(block.Txns) == 0 {
-		return fmt.Errorf("no transaction for the block")
-	}
-
 	data, err := json.Marshal(blockToBlockEvent(block))
 	if err != nil {
 		return fmt.Errorf("error marshalling block: %v", err)
 	}
 
-	t := block.Txns[len(block.Txns)-1]
 	block.Events = append(block.Events, event.Event{
 		BlockNumber: block.Round,
-		TxHash:      t.Hash,
+		TxHash:      "",
 		Type:        int(event.TypeStats),
 		Tag:         int(event.TagAddBlock),
-		Index:       t.Hash,
+		Index:       block.Hash,
 		Data:        string(data),
 	})
 
--- a/code/go/0chain.net/chaincore/block/block_eventdb.go
+++ b/code/go/0chain.net/chaincore/block/block_eventdb.go
@@ -1,0 +1,54 @@
+package block
+
+import (
+	"encoding/json"
+	"fmt"
+
+	"0chain.net/smartcontract/dbs/event"
+)
+
+func blockToBlockEvent(block *Block) *event.Block {
+	return &event.Block{
+		Hash:                  block.Hash,
+		Version:               block.Version,
+		CreationDate:          block.CreationDate.Duration().Nanoseconds(),
+		Round:                 block.Round,
+		MinerID:               block.MinerID,
+		RoundRandomSeed:       block.RoundRandomSeed,
+		MerkleTreeRoot:        block.GetMerkleTree().GetRoot(),
+		StateHash:             string(block.ClientStateHash),
+		ReceiptMerkleTreeRoot: block.GetReceiptsMerkleTree().GetRoot(),
+		NumTxns:               len(block.Txns),
+		MagicBlockHash:        block.MagicBlock.Hash,
+		PrevHash:              block.PrevHash,
+		Signature:             block.Signature,
+		ChainId:               block.ChainID,
+		ChainWeight:           fmt.Sprintf("%f", block.ChainWeight),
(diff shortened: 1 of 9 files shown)


code/go/0chain.net/chaincore/block/entity.go
```diff
@@ -1008,6 +1008,11 @@ func (b *Block) ComputeStateLocal(ctx context.Context, c Chainer) error {
 		}
 	}
 
+	err := emitBlockEvent(b)
```
amanjain97: Please check if it is the correct place to emit the event.

code/go/0chain.net/chaincore/block/block_eventdb.go
```diff
@@ -0,0 +1,54 @@
...
+		return fmt.Errorf("error marshalling block: %v", err)
+	}
+
+	t := block.Txns[len(block.Txns)-1]
+	block.Events = append(block.Events, event.Event{
+		BlockNumber: block.Round,
+		TxHash:      t.Hash,
+		Type:        int(event.TypeStats),
+		Tag:         int(event.TagAddBlock),
+		Index:       t.Hash,
```
amanjain97: I am passing the transaction hash of last transaction. Is it okay or I am missing anything ?

code/go/0chain.net/chaincore/block/block_eventdb.go
```diff
@@ -0,0 +1,54 @@
...
+		ChainWeight:           fmt.Sprintf("%f", block.ChainWeight),
+		RunningTxnCount:       fmt.Sprintf("%d", block.RunningTxnCount),
+		RoundTimeoutCount:     block.RoundTimeoutCount,
+		CreatedAt:             block.CreationDateField.ToTime(),
+	}
+}
+
+func emitBlockEvent(block *Block) error {
+	if len(block.Txns) == 0 {
+		return fmt.Errorf("no transaction for the block")
```
Sriep: We still need to save the block even if there are no transactions. So we can remove this check.

code/go/0chain.net/chaincore/block/entity.go
Sriep: Yes. However, I think to be safe we should emit the previous block here:; `b.PrevBlock` rather than `b`.

code/go/0chain.net/chaincore/block/entity.go
```diff
@@ -1008,6 +1008,11 @@ func (b *Block) ComputeStateLocal(ctx context.Context, c Chainer) error {
...
+	if err != nil {
+		return err
```
Sriep: Don't return an error, log it and continue, `logger.Logger.Error`. In practice, this should never happen as an `event.Block` should always marshal.

code/go/0chain.net/chaincore/block/block_eventdb.go
Sriep: Use the block's hash as an index. b.Hash. 

There is no meaning for the `Event.TxHash` field, so set to `""`, we can revisit if that breaks anything. However, it seems a reasonable way to indicate an event that is not linked to a particular transaction.

Sriep (changes requested):
I think it is safest if we emit the previous block, rather than the current one. We cannot be certain that some key part of the block won't change. 

If we were to emit the current block here we would need to call `b.GetHash()`and put a log in `Block.GetHash()`, to confirm that the block's hash does not change after this point. But even if this is ok, it leaves a GOTCHA lying around for later.

code/go/0chain.net/chaincore/block/entity.go
kushthedude: Is there any function which makes sure that the block object has all the fields populated which we need, or is it by design that block will have each & every field populated? Should we keep this check for a fail safe, what repurcussions can we have if we return an error here?

amanjain97: I think it is correct to log the error rather than throwing back the error as it can stop other events and subsequent process just because of events. Also, even if there are some fields that are not present, then they would just be `null` after marshal.

amanjain97:
@Sriep I have updated the PR based on comments. Please check.

kushthedude (dismissed):

kushthedude (approved):

PR #891, opened 2022-01-14
Fix #890 added transaction parameter clientID

### LINKED ISSUE #890: Transactions in event database should be searchable with client_id
https://0chain.slack.com/archives/C02QDMYFQJG/p1642022754110100?thread_ts=1641664353.058800&cid=C02QDMYFQJG

As per @shaktals comment, we need to add more capabilities to the transaction api from event database that will help them to search transactions in the event database by client_id

samtholiya:
  
...


code/go/0chain.net/smartcontract/storagesc/handler.go
```diff
@@ -111,14 +111,22 @@ func (msc *StorageSmartContract) GetTransactionByHashHandler(
...
-	transaction, err := balances.GetEventDB().GetTransactionByHash(transactionHash)
-	return &transaction, err
+	if len(transactionHash) > 0 {
+		transaction, err := balances.GetEventDB().GetTransactionByHash(transactionHash)
+		return []event.Transaction{transaction}, err
+	}
+	if len(clientID) > 0 {
+		transactions, err := balances.GetEventDB().GetTransactionByClientId(clientID)
+		return transactions, err
+	}
```
kushthedude: Lets not clutter the existing handler to handle the clientID too

code/go/0chain.net/smartcontract/storagesc/sc.go
```diff
@@ -113,7 +113,7 @@ func (ssc *StorageSmartContract) setSC(sc *sci.SmartContract, bcContext sci.BCCo
 	// challenge pool
 	ssc.SmartContract.RestHandlers["/getChallengePoolStat"] = ssc.getChallengePoolStatHandler
 	// events db
-	ssc.SmartContractExecutionStats["/get_transaction"] = ssc.GetTransactionByHashHandler
+	ssc.SmartContract.RestHandlers["/get_transaction"] = ssc.GetTransactionByHashHandler
```
kushthedude: Add a new API in addition to the which queries the transaction by Hash, and add one api `/transaction/search` where I can provide multiple query params as per my need.
This is in conjunction to what explorer uses as of now 
```
/transaction/:hash: To query a transaction by its hash
/transactions/search: API to search transactions by providing multiple query params
/transactions/block-hash: API to get transactions by a block's hash.
```

kushthedude (commented):
Let us be sure that all the new APIs which are being added to aid the explorer are in conjunction to what explorer uses as of now.


@shaktals Can you tell what all other params does the `search` endpoint expects apart from the clientID?
```
/transactions/search: API to search transactions by providing multiple query params
```

code/go/0chain.net/smartcontract/storagesc/sc.go
samtholiya: 1. `/transaction/:hash:` is not possible because we are using map (I hate this approach) as key to the handler which makes it impossible for us to implement URL parameters naturally. (We can still improve that)
2. Why do we need two APIs (`/transactions/search`, `/transactions/block-hash`) doing the same thing. According to API standards
a.  `/transaction/:transaction_id` for single transaction
b. `/transaction?client_id=xyz&metadata=2332` for multiple transactions with filters

Now `a` is not possible because of the map structure 
The query is possible which I would recommend.

shaktals:
As per @samtholiya comment, the search params we are using today on explorer are:
- hash
- client_id
- metadata

Sriep:
I think we should leave GetTransactionByHashHandler unchanged and add a new endpoint GetTransactionByClientIdHandler.

The two endpoints seem too different to combine. For example, their return types are different, one returns a transaction and the other a list of transactions.

code/go/0chain.net/smartcontract/storagesc/sc.go
Sriep: I don't understand the point here? You have two endpoints
```go

ssc.SmartContractExecutionStats["/get_transaction_by_hash"] = ssc.GetTransactionByHashHandler
ssc.SmartContractExecutionStats["/get_transaction_by_client"] = ssc.GetTransactionByClientHandler
```
Where is the problem?

shaktals: I agree with @samtholiya that having two endpoints would be more semantic:

a. `/transaction/:id` - single txn
b. `/transactions` - multiple txns, searched by query params

To reiterate, now we use `client_id` and `metadata` (or maybe we could use just filename? which is our current use case).

samtholiya: @shaktals unfortunately a. is not possible due to map implementation so we would be having
a. `/get_transaction_by_hash?transaction_hash=:id`
b. `/get_transaction_by_client?client=:id`
and so on...
cc: @kushthedude

shaktals: Really? Aren't we building this as new endpoints? The above looks kinda of ugly :sweat_smile:
Can we do something like:
a. `/transaction?hash=:id`
b. `/transactions?client_id=:id`
?

Or at the very least we should remove the `get_` part...

shaktals: Here is a [nice article](https://www.freecodecamp.org/news/rest-api-best-practices-rest-endpoint-design-examples/) about semantic REST endpoints

samtholiya: @shaktals .... https://github.com/0chain/0chain/pull/891#discussion_r786959938 this was the comment why I said that would be the endpoint.
I would say @Sriep @shaktals @kushthedude finalize what you guys want... and I will change that signature accordingly 😅 

I am in favor of https://github.com/0chain/0chain/pull/891#discussion_r787235621 as of now... (given the code conditions we have)
But we can improve the API to https://github.com/0chain/0chain/pull/891#discussion_r787129651 if we work on https://github.com/0chain/0chain/issues/926 in a better way (changing the logic to not use a map)

dabasov:
I like @shaktals proposal this get stuff is really ugly, my understanding is that URL is like client interface, we can make some adapter handler that will call different hadlers or we can rewrite urls, I'm not a fan of url-rewriting, but it might be an option though

code/go/0chain.net/smartcontract/storagesc/sc.go
shaktals: For me, both your last suggestions are ok.
The last one still being the cleanest.
If it's not too much work to remove the `map` implementation, I say go for that.

shaktals: For me, both options you mention are ok.
The last one still being the cleanest.
If it's not too much work to remove the map implementation, I say go for that.

samtholiya:
transaction hash:
image
client_id:
image
block_hash:
image

kushthedude (approved):
@shaktals We have configured pagination using limit & offset param, let us know if these works for you or UI needs to use pageSize & pageNumber

PR #913, opened 2022-01-17
Fix #910 Added mint to eventdb

### LINKED ISSUE #910: New mints table in the event databse
A new mint database table. For each block, this will hold the total number of tokens minted during that block.

Each time tokens are [minted](https://github.com/0chain/0chain/blob/staging/code/go/0chain.net/chaincore/chain/state/state_context.go#L145) emit an event. Total up the `mint events` each block and add a row to the mint table.

We might expand this later.

We will also want an endpoint to return the sum of the mints for a given range of blocks.

samtholiya:
  
...


kushthedude (commented):
Looks fine, but lets hold off the merge until the blocks issue is solved so that we may check if mints are getting populated or not

code/go/0chain.net/smartcontract/dbs/event/block.go
```diff
@@ -25,6 +25,7 @@ type Block struct {
 	ChainId               string    `json:"chain_id"`
 	RunningTxnCount       string    `json:"running_txn_count"`
 	RoundTimeoutCount     int       `json:"round_timeout_count"`
+	MintTotalAmount       int64     `json:"mint_total_amount"`
```
Sriep: Add a amount minted this block field.

code/go/0chain.net/smartcontract/dbs/event/mint.go
```diff
@@ -0,0 +1,10 @@
+package event
+
+type Mint struct {
```
Sriep, line 3: I think we can make do with replacing this table with a field in the `block` table. We only want the total amount minted each block, not a list of mints per block.

code/go/0chain.net/chaincore/chain/state/state_context.go
```diff
@@ -163,6 +164,22 @@ func (sc *StateContext) AddSignedTransfer(st *state.SignedTransfer) {
...
 	sc.mutex.Lock()
 	defer sc.mutex.Unlock()
+	data, err := json.Marshal(event.Mint{
+		BlockHash: sc.block.Hash,
+		Amount:    int64(m.Amount),
+	})
+	if err != nil {
+		return fmt.Errorf("marshalling mint: %v", err)
+	}
+	sc.events = append(sc.events, event.Event{
```
Sriep: What we actually want to send to the event database is the total amount minted in a block, not each individual  mint.

I suggest replacing this with emitting a total minted event in [updateState](https://github.com/0chain/0chain/blob/staging/code/go/0chain.net/chaincore/chain/state.go#L316). Total mints in for loop and emit afterwards.

code/go/0chain.net/smartcontract/dbs/event/block.go
```diff
@@ -34,6 +35,11 @@ func (edb *EventDb) GetBlocksByHash(hash string) (Block, error) {
 	return block, res.Error
 }
 
+func (edb *EventDb) GetBlocksMintTotal(hashes []string) (int64, error) {
```
Sriep: Have a block range as input, and sum the amount minted in each of those blocks. Let the `to` block default to the latest block, .

code/go/0chain.net/smartcontract/storagesc/handler.go
```diff
@@ -179,6 +179,22 @@ func (msc *StorageSmartContract) GetTransactionByHashHandler(
 	return transaction, err
 }
 
+func (msc *StorageSmartContract) GetBlockTotalByHashRange(ctx context.Context, params url.Values, balances cstate.StateContextI) (interface{}, error) {
+	blockHashes, ok := params["blockHashes"]
```
Sriep: Enter a block range rather than hashes.

code/go/0chain.net/smartcontract/storagesc/handler.go
```diff
@@ -179,6 +179,22 @@ func (msc *StorageSmartContract) GetTransactionByHashHandler(
 	return transaction, err
 }
 
+func (msc *StorageSmartContract) GetBlockTotalByHashRange(ctx context.Context, params url.Values, balances cstate.StateContextI) (interface{}, error) {
```
Sriep: GetMintTotalByBlockRange

Sriep (changes requested):
I think we don't need a sepearte mint table. As we are only in the total minted on any block we can keep that in the block table.

code/go/0chain.net/smartcontract/dbs/event/process.go
```diff
@@ -155,6 +156,13 @@ func (edb *EventDb) addStat(event Event) error {
 			return err
 		}
 		return edb.addOrOverwriteValidator(vn)
+	case TagAddMint:
```
Sriep: TagAddMint and TagAddBlock could come in either order, but they will both be populating the same row. Feel free to combine the two if you see a convenient way of doing so. Otherwise, both can be an update or insert.

code/go/0chain.net/smartcontract/dbs/event/mint.go
samtholiya: This is not a table... it is just a mint structure... which we use to add mint to block table

code/go/0chain.net/chaincore/chain/state/state_context.go
samtholiya: done

code/go/0chain.net/smartcontract/dbs/event/block.go
samtholiya: need assistance how can be a string be a range...?
eg: a number can be from 1 to 10
but for a string that is a unique id... how do we use a range?

code/go/0chain.net/smartcontract/storagesc/handler.go
samtholiya: done

code/go/0chain.net/smartcontract/dbs/event/process.go
samtholiya: naming it update confuses with the capability to change other properties... are we going to allow that?

code/go/0chain.net/smartcontract/storagesc/handler.go
```diff
@@ -241,6 +241,28 @@ func (msc *StorageSmartContract) GetTransactionByHashHandler(
...
+	if err != nil {
+		return nil, errors.New("round from not provided")
+	}
+
+	roundTo, err := strconv.Atoi(params.Get("round_to"))
+	if err != nil {
+		return nil, errors.New("round from not provided")
+	}
+
+	total, err := balances.GetEventDB().GetBlocksMintTotal(int64(roundFrom), int64(roundTo))
```
Sriep: For the  moment we should check `balances.GetEventDb() != nil` to avoid a possible panic.

Sriep (changes requested):

code/go/0chain.net/chaincore/chain/state.go
```diff
@@ -359,7 +359,6 @@ func (c *Chain) updateState(ctx context.Context, b *block.Block, bState util.Mer
 			return
 		}
 	}
```
dabasov, line 361: revert change

code/go/0chain.net/smartcontract/dbs/event/event_test.go
```diff
@@ -38,3 +38,42 @@ func TestSetupDatabase(t *testing.T) {
 	err = eventDb.AutoMigrate()
 	require.NoError(t, err)
 }
+
+func (edb *EventDb) drop() error {
```
dabasov: why eventdb.Drop() doesnt work?

code/go/0chain.net/smartcontract/dbs/event/snapshot.go
```diff
@@ -0,0 +1,59 @@
...
+	"sync"
+
+	"0chain.net/core/logging"
+	"go.uber.org/zap"
+	"gorm.io/gorm"
+)
+
+// swagger:model Block
+type Snapshot struct {
+	gorm.Model
```
dabasov: not sure we need model here, it will create couple of columns and index for deletes that we don't use

code/go/0chain.net/smartcontract/dbs/event/snapshot.go
```diff
@@ -0,0 +1,59 @@
...
+	Round           int64  `gorm:"primaryKey;autoIncrement:false" json:"round"`
+	BlockHash       string `json:"block_hash"`
+	MintTotalAmount int64  `json:"mint_total_amount"`
+}
+
+var snapshotLock sync.Mutex
```
dabasov: donot need this lock

code/go/0chain.net/smartcontract/dbs/event/snapshot.go
```diff
@@ -0,0 +1,59 @@
...
+
+func (edb *EventDb) GetRoundsMintTotal(from, to int64) (int64, error) {
+	var total int64
+	return total, edb.Store.Get().Model(&Snapshot{}).Where("round between ? and ?", from, to).Select("sum(mint_total_amount)").Scan(&total).Error
+}
+
+func (edb *EventDb) addOrUpdateTotalMint(mint Mint) error {
+	snapshotLock.Lock()
```
dabasov: the problem with this solution is that we can't protect data from update from other function that does not share this lock, even that we handle data change from one place in go code it is buggy and dev can forget or doesn't know about this lock. The best practice is to use SELECT ... FOR UPDATE construction. look for examplehere https://stackoverflow.com/questions/18879584/postgres-select-for-update-in-functions

code/go/0chain.net/smartcontract/dbs/event/snapshot.go
```diff
@@ -0,0 +1,59 @@
...
+
+	snapshot, err := edb.getRoundSnapshot(mint.Round)
+	if err != nil {
+		if errors.Is(err, gorm.ErrRecordNotFound) {
+			snapshot = &Snapshot{
+				Round:           mint.Round,
+				MintTotalAmount: mint.Amount,
+				BlockHash:       mint.BlockHash,
+			}
+			return edb.addSnapshot(snapshot)
```
dabasov: if I understand correctly you even need upsert, it can be done in postgres as follows INSERT ... ON CONFLICT (key) DO UPDATE https://stackoverflow.com/questions/17267417/how-to-upsert-merge-insert-on-duplicate-update-in-postgresql

code/go/0chain.net/smartcontract/dbs/event/transaction_test.go
```diff
@@ -41,6 +41,26 @@ func TestAddTransaction(t *testing.T) {
...
+	}
+	eventDb, err := NewEventDb(access)
+	if err != nil {
+		t.Skip("only for local debugging, requires local postgresql")
+	}
+	return eventDb
+}
+
 func TestFindTransactionByHash(t *testing.T) {
 	access := config.DbAccess{
```
dabasov: remove duplicated code

dabasov (changes requested):

code/go/0chain.net/chaincore/chain/state.go
mohsenno1: This is not the right place to emit this event. I moved it to [finalizBlock](https://github.com/0chain/0chain/blob/16d00f0f71c213ef88ee03c404f47ca3d6516558/code/go/0chain.net/chaincore/chain/protocol_block.go#L357) 

[updateState](https://github.com/0chain/0chain/blob/16d00f0f71c213ef88ee03c404f47ca3d6516558/code/go/0chain.net/chaincore/chain/state.go#L204) is getting called multiple time during the generation of the block or in verifying received blocks. I guess we only need to emit minted amount for a finalized block. right?

Sriep:
My suggestion would be to only update the snapshot table once per round. So keep summary stats in a data object then add a new row once at the start of a new round. Something like:
```go
type roundData {
  thisRound int64
  mintsThisRound currenty.Coin
  ... other  snapshot data
}

func (edb *EventDb) addEventsWorker(ctx context.Context) {
      var data = new(roundData)
      for {
		events := <-edb.eventsChannel
		edb.addEvents(ctx, events)
		for _, event := range events {
			var err error = nil
			switch EventType(event.Type) {
			case TypeStats:
				err = edb.addStat(event, data)
			case TypeError:
				err = edb.addError(Error{
					TransactionID: event.TxHash,
					Error:         event.Data,
				})
			default:
			}
			if err != nil {
				logging.Logger.Error(
					"event could not be processed",
					zap.Any("event", event),
					zap.Error(err),
				)
			}
		}
	}
}
```
then
```go
func (edb *EventDb) addStat(event Event, data *roundData) error {
        if event.BlockNumber == data.thisRound +1 {
            err := addToSnapshotTable(data); err != nil 
            data = roundData{ thisRound : event.BlockNumber }
            return err
        }
        if event.BlockNumber != data.thisRound {
           panic("should be impossible, the event database is corrupted needs to be rebuilt")
        }
	switch EventTag(event.Tag) {
        ....
           case TagAddMint:
       		var mint Mint
		err := json.Unmarshal([]byte(event.Data), &mint)
		if err != nil {
			return err
		}
		data.mints += mint
       }
```
The snapshot table is then immutable in that after a row gets added that row cannot be changed further..

mohsenno1:
> My suggestion would be to only update the snapshot table once per round. So keep summary stats in a data object then add a new row once at the start of a new round. Something like:
> 
> ```go
> type roundData {
>   thisRound int64
>   mintsThisRound currenty.Coin
>   ... other  snapshot data
> }
> 
> func (edb *EventDb) addEventsWorker(ctx context.Context) {
>       var data = new(roundData)
>       for {
> 		events := <-edb.eventsChannel
> 		edb.addEvents(ctx, events)
> 		for _, event := range events {
> 			var err error = nil
> 			switch EventType(event.Type) {
> 			case TypeStats:
> 				err = edb.addStat(event, data)
> 			case TypeError:
> 				err = edb.addError(Error{
> 					TransactionID: event.TxHash,
> 					Error:         event.Data,
> 				})
> 			default:
> 			}
> 			if err != nil {
> 				logging.Logger.Error(
> 					"event could not be processed",
> 					zap.Any("event", event),
> 					zap.Error(err),
> 				)
> 			}
> 		}
> 	}
> }
> ```
> 
> then
> 
> ```go
> func (edb *EventDb) addStat(event Event, data *roundData) error {
>         if event.BlockNumber == data.thisRound +1 {
>             err := addToSnapshotTable(data); err != nil 
>             data = roundData{ thisRound : event.BlockNumber }
>             return err
>         }
>         if event.BlockNumber != data.thisRound {
>            panic("should be impossible, the event database is corrupted needs to be rebuilt")
>         }
> 	switch EventTag(event.Tag) {
>         ....
>            case TagAddMint:
>        		var mint Mint
> 		err := json.Unmarshal([]byte(event.Data), &mint)
> 		if err != nil {
> 			return err
> 		}
> 		data.mints += mint
>        }
> ```
> 
> The snapshot table is then immutable in that after a row gets added that row cannot be changed further..

mohsenno1:
> My suggestion would be to only update the snapshot table once per round. So keep summary stats in a data object then add a new row once at the start of a new round. Something like:
> 
> ```go
> type roundData {
>   thisRound int64
>   mintsThisRound currenty.Coin
>   ... other  snapshot data
> }
> 
> func (edb *EventDb) addEventsWorker(ctx context.Context) {
>       var data = new(roundData)
>       for {
> 		events := <-edb.eventsChannel
> 		edb.addEvents(ctx, events)
> 		for _, event := range events {
> 			var err error = nil
> 			switch EventType(event.Type) {
> 			case TypeStats:
> 				err = edb.addStat(event, data)
> 			case TypeError:
> 				err = edb.addError(Error{
> 					TransactionID: event.TxHash,
> 					Error:         event.Data,
> 				})
> 			default:
> 			}
> 			if err != nil {
> 				logging.Logger.Error(
> 					"event could not be processed",
> 					zap.Any("event", event),
> 					zap.Error(err),
> 				)
> 			}
> 		}
> 	}
> }
> ```
> 
> then
> 
> ```go
> func (edb *EventDb) addStat(event Event, data *roundData) error {
>         if event.BlockNumber == data.thisRound +1 {
>             err := addToSnapshotTable(data); err != nil 
>             data = roundData{ thisRound : event.BlockNumber }
>             return err
>         }
>         if event.BlockNumber != data.thisRound {
>            panic("should be impossible, the event database is corrupted needs to be rebuilt")
>         }
> 	switch EventTag(event.Tag) {
>         ....
>            case TagAddMint:
>        		var mint Mint
> 		err := json.Unmarshal([]byte(event.Data), &mint)
> 		if err != nil {
> 			return err
> 		}
> 		data.mints += mint
>        }
> ```
> 
> The snapshot table is then immutable in that after a row gets added that row cannot be changed further..

@Sriep This issue takes so long and this looks like an improvement. Let's close this. I'll apply your changes during adding other fields of snapshot.

code/go/0chain.net/chaincore/block/magicBlock/generate_test.go
```diff
@@ -0,0 +1,28 @@
+package main
+
+import (
+	"log"
+	"testing"
+
+	"github.com/herumi/bls/ffi/go/bls"
+)
+
+func TestKeyGeneration(t *testing.T) {
```
dabasov, line 10: why this test is added?

code/go/0chain.net/chaincore/block/magicBlock/main.go
```diff
@@ -129,9 +129,15 @@ func (cmd *cmdMagicBlock) createShareOrSigns() {
 			cmd.summaries[nd.SetIndex].SecretShares[partyId.GetHexString()] = share.GetHexString()
 			if mid != id {
 				var privateKey bls.Key
-				if err := privateKey.SetHexString(cmd.yml.MinersMap[id].PrivateKey); err != nil {
+				privateKeyBytes, err := hex.DecodeString(cmd.yml.MinersMap[id].PrivateKey)
```
dabasov, line 132: why this is changed?

code/go/0chain.net/chaincore/smartcontract/handler.go
```diff
@@ -48,7 +48,7 @@ func GetSmartContract(scAddress string) sci.SmartContractInterface {
 
 func ExecuteWithStats(smcoi sci.SmartContractInterface, t *transaction.Transaction, funcName string, input []byte, balances c_state.StateContextI) (string, error) {
 	ts := time.Now()
-	inter, err := smcoi.Execute(t, funcName, input, balances)
+	inter, err := smcoi.Execute(t.Clone(), funcName, input, balances)
```
dabasov, line 51: why this change here?

code/go/0chain.net/smartcontract/benchmark/main/config/benchmark.yaml
```diff
@@ -108,17 +108,17 @@ smart_contracts:
...
       duration: 50h
       read_price_range:
         min: 0.0
-        max: 0.1
+        max: 2.0
       write_price_range:
         min: 0.0
-        max: 0.1
-      max_challenge_completion_time: 3m
+        max: 2.0
```
dabasov, line 120: why this file is changed?

code/go/0chain.net/smartcontract/dbs/event/allocation.go
```diff
@@ -113,32 +114,13 @@ func (edb EventDb) GetActiveAllocsBlobberCount() (int64, error) {
 	return count, nil
 }
 
-func (edb *EventDb) overwriteAllocation(alloc *Allocation) error {
-	return edb.Store.Get().Model(&Allocation{}).Where("allocation_id = ?", alloc.AllocationID).Updates(alloc).Error
+func (edb *EventDb) updateAllocation(updates *dbs.DbUpdates) error {
```
dabasov, line 117: why this is changed?

code/go/0chain.net/smartcontract/dbs/event/allocation_test.go
```diff
@@ -353,7 +353,7 @@ func TestAllocations(t *testing.T) {
 		BlockNumber: 1,
 		TxHash:      "txn_hash",
 		Type:        int(TypeStats),
-		Tag:         int(TagAddOrOverwriteAllocation),
+		Tag:         int(TagAddAllocation),
```
dabasov, line 356: why do we change it?

dabasov (changes requested):

Sriep:
> 
> @Sriep This issue takes so long and this looks like an improvement. Let's close this. I'll apply your changes during adding other fields of snapshot.

I don't follow? There is no point adding and testing code that is going to immediately be removed. I think it might make more sense to do the snapshot table https://github.com/0chain/0chain/issues/1390 first, then this Pr becomes much easier.

mohsenno1:
branch has merging problems. New PR created: [#1455](https://github.com/0chain/0chain/pull/1455)

PR #930, opened 2022-01-20
Feature/transaction errors

### LINKED ISSUE #887: Transaction: chargeable errors
**Abstract**
Clients should be charged for erroneous transactions. If client's transaction failed with benign error, this transaction should be included in the block and fees should be charged from miner.

**Implementation**
All errors returned by smartcontracts have to be split in two groups: client errors and server errors.
- Client errors. These errors should be included in the block, if fees can be charged from `txn.Client` balance. This group is pretty large and often errors from this group can be created with `errors.New()` or similar. Since these errors won't need any specific logic to handle they can be returned from any smart contract func freely. 
- Server errors. These errors occurs when miner can't process smart contract, since this group is much smaller, these errors are following:

 Server errors:
- context.Deadline
- context.Cancel
- ErrSmartContractContext
- ErrNodeNotFound
- EncodingError
- InternalServerError

It is possible to unite all `ErrNodeNotFound`, `EncodingError`, `InternalServerError` in one `InternalServerError`, can discuss

TODO: link old issue

dabasov:
This PR includes changes needed for chargeable transaction errors support.
It also includes stability changes and fixes to couple of bugs mentioned in #view-change slack

successful test run:
https://github.com/0chain/0chain/runs/4886993726?check_suite_focus=true

- [gosdk](https://github.com/0chain/gosdk/pull/361)
- [zwalletcli](https://github.com/0chain/zwalletcli/pull/88)
- [zboxcli](https://github.com/0chain/zboxcli/pull/158)
- [system_tests](https://github.com/0chain/system_test/pull/167)
...


service-0chain:
| config  | value |
| tests skipped  | FALSE  |
| system-tests  | feature/transaction-errors  |
| gosdk  | NOT OVERRIDDEN  |
| zboxcli  | feature/transaction-errors  |
| zwalletcli  | feature/transaction-errors  |
| 0chain  | feature/transaction-errors  |
| blobber  | staging  |
| authorizer  | staging  |
| 0box  | staging  |
| 0block  | staging  |
| 0dns  | staging  |
| explorer  | staging  |
| 0proxy  | staging  |
| 0search  | staging  |

service-0chain:
| config  | value |
| tests skipped  | FALSE  |
| system-tests  | feature/transaction-errors  |
| gosdk  | NOT OVERRIDDEN  |
| zboxcli  | feature/transaction-errors  |
| zwalletcli  | feature/transaction-errors  |
| 0chain  | feature/transaction-errors  |
| blobber  | staging  |
| authorizer  | staging  |
| 0box  | staging  |
| 0block  | staging  |
| 0dns  | staging  |
| explorer  | staging  |
| 0proxy  | staging  |
| 0search  | staging  |

dabasov:
> I did the same changes before, but the system tests failed. This will change the response data, so the clients need to be adjust to the changes as well.

system tests are now passing, looks like something was fixed there

code/go/0chain.net/chaincore/chain/state.go
```diff
@@ -181,6 +180,11 @@ func (c *Chain) updateState(ctx context.Context, b *block.Block, bState util.Mer
 	)
 	defer func() { events = sctx.GetEvents() }()
 
+	//we should check that client hash enough funds to pay for transaction before heavy computations are executed
```
peterlimg: `hash` -> `has`

code/go/0chain.net/chaincore/chain/state.go
```diff
@@ -198,26 +202,37 @@ func (c *Chain) updateState(ctx context.Context, b *block.Block, bState util.Mer
...
 					zap.String("begin client state", util.ToHex(startRoot)),
 					zap.String("prev block", b.PrevBlock.Hash),
 					zap.Any("txn", txn))
-				return
+
+				//refresh client state context, so all changes made by broken smart contract are rejected, it will be used to add fee
+				clientState = CreateTxnMPT(bState) // begin transaction
+				sctx = c.NewStateContext(b, clientState, txn, nil)
+
+				if e, ok := err.(*common.Error); ok {
```
peterlimg: The `common.Error` has implemented the `Error` interface, we can call the `output = err.Error()` directly here.

code/go/0chain.net/miner/protocol_block_main.go
```diff
@@ -274,22 +289,22 @@ func (mc *Chain) generateBlock(ctx context.Context, b *block.Block,
 	if config.DevConfiguration.IsFeeEnabled {
 		err = mc.processTxn(ctx, mc.createFeeTxn(b), b, blockState, iterInfo.clients)
 		if err != nil {
-			return err
+			logging.Logger.Error("generate block (payFees)", zap.Int64("round", b.Round), zap.Error(err))
```
peterlimg, line 292: What would happen if the block reward/fee transaction were failed, maybe we can not allow the failure of these kind of transactions, as miners/blobbers will lose their profit.

docker.local/config/b0magicBlock_3_miners_3_sharders.json
```diff
@@ -0,0 +1,225 @@
+{
```
peterlimg, line 1: Do we have any changes that require the rebuilding of magic block, if yes, I'll need to rebuild it on hero sever.

peterlimg (changes requested):
The `profile.out` file was uploaded mistakenly perhaps, please remove it. Also left some comments, please have a check.

code/go/0chain.net/miner/protocol_block_main.go
dabasov: yep I thought about it, bit it is much better, than network timeout due to absent proposal, so I decided that network liveness is high priority here,

docker.local/config/b0magicBlock_3_miners_3_sharders.json
dabasov: I think originally there was 4 miners 3 sharders, that's why I regenerated it

code/go/0chain.net/miner/protocol_block_main.go
peterlimg: okay, we can fire an issue to document this behavior in case some day we want to re-consider about this or have better solution for the network liveness.

peterlimg: Hmm, I'm still worried about the failure of reward and fee transactions here, we can ignore failures of the transaction for commiting the config updates, but ignoring the reward transaction errors may need further check. @guruhubb  what do you think.

guruhubb: I think this is fine @peterlimg.  If there is less time to generate the block, the miner needs to discard time-consuming transactions and just add their reward transaction and send it otherwise they won't receive their reward.  

There is no reason for a miner to discard a lot of transactions as they will earn less fees.

peterlimg: @guruhubb it's not about we don't have not enough time to generate the block. The generator has already finished most of the work for packing a block, the last thing it tries to do is create and execute the reward/fee transaction so that miners would get rewarded. The question here is whether we should allow this transaction to fail during the block generation process. 

If yes, then generators and miners will create and verify this block for free, no miners would get rewarded. If no, then this block generator will failed to pack the block, and other generators will try to create the block with rewards. If any of the generator created it successfully, then miners will get rewarded.

 Anyway, I prefer to do `no` as miners may complain for not receiving the rewards.

guruhubb: Ok got it. Yes miners won't be happy if reward/fee transaction is not executed, but if it is a code execution issue due to MPT growth, then the whole chain will stall. I think if it fails, then the miner can take less time for the transactions (say half of what it started out) for the next block it generates. @peterlimg

peterlimg: @guruhubb yes, take less time for the transactions in next block generation might be an option, we can do this later. @dabasov what do you think.

peterlimg (dismissed):

code/go/0chain.net/miner/protocol_block_main.go
dabasov: yep, I agree to do it later

guruhubb (approved):

peterlimg (approved):

PR #936, opened 2022-01-22
Feat/add local debugging support

cnlangzi:
  
...

changes:
- added dev.local/cli.sh to support debugging outside docker
  + install deps
  + install cassandra
  + install postgres
  + install redis/redis_txns
  + build and run sharder/miner
  + install debugger config in vscode


code/go/0chain.net/miner/miner/main.go
```diff
@@ -42,16 +43,30 @@ func main() {
 	delayFile := flag.String("delay_file", "", "delay_file")
 	magicBlockFile := flag.String("magic_block_file", "", "magic_block_file")
 	initialStatesFile := flag.String("initial_states", "", "initial_states")
+
+	workdir := ""
```
ssardana96: for neat we can create all var in beginning
var (
workdir string
redisHost string
redisPort int
.
.
.

)

dev.local/cli.base.sh
```diff
@@ -0,0 +1,14 @@
+#!/bin/bash
+
+#clean all data
+cleanAll() {
+    cd $root
```
ssardana96, line 5: whats the value of $root ?

ssardana96 (changes requested):

ssardana96:
lgtm
just few neats.

dev.local/cli.base.sh
cnlangzi: it is ./dev.local

kushthedude: It is defined in the files referencing this function, it is just the present working directory, https://github.com/0chain/0chain/blob/11d3c0f916ea53c84b0faea15ab1c74ef50bd829/dev.local/cli.sh#L5

kushthedude (commented):
Can we add a CI Step to try running this complete script? If we are going to shift to automated scripts to run check, I recommend we should have a sanity check that if the script is always working or not, which in my opinion was one of the major pitfalls of manually setting up the chain since no-one actually had the bandwidth to try these steps manually everytime with each commit.

cnlangzi:
> Can we add a CI Step to try running this complete script? If we are going to shift to automated scripts to run check, I recommend we should have a sanity check that if the script is always working or not, which in my opinion was one of the major pitfalls of manually setting up the chain since no-one actually had the bandwidth to try these steps manually everytime with each commit.

it is designed for local development. developer runs it everyday like I do in blobber development. it seems be not worth to check it in CI job with each commit. and it is an interactive script only for MacOS.

ssardana96 (approved):

code/go/0chain.net/miner/miner/main.go
```diff
@@ -36,6 +36,14 @@ import (
 	"0chain.net/smartcontract/setupsc"
 )
 
+var (
+	workdir       string
```
ssardana96: @cnlangzi you can keep them local

ssardana96: Local to main

ssardana96 (changes requested):

ssardana96 (approved):
good to merge

Issue #741, opened 2021-12-03
Modify challenge creation algorithm


Sriep:
### Challenges

Challenges are generated by the [generateChallenges](https://github.com/0chain/0chain/blob/staging/code/go/0chain.net/smartcontract/storagesc/challenge.go#L524) method. This method is called when the blobber calls either the `read_redeem` or `commit_connetion` endpoints. 
`generateChallenges` creates up to `sc.yaml.storagesc.max_challenges_per_generation` challenges, which typically defaults to 100.

The existing method for creating a challenge selects an allocation at random, then randomly orders the blobbers and select the first with a file/allocation root. Unfortunately, this method could be subject to a [sybil attack](https://en.wikipedia.org/wiki/Sybil_attack), so we intend to redesign this algorithm.

### Select challenge by blobber algorithm

1. Blobbers are partitioned into even-sized partitions.
2. Select a partition at random.
3. Weight the blobbers in the randomly selected partition.
4. Randomly select a blobber using a weighted probability distribution as defined by (3)
5. Select a random allocation supported by the blobber that has an allocation root - file.
6. Create a challenge out of the (blobber, allocation, allocation root)
7. Save challenge information to MPT and emit challenge event.

Repeat 1-7 a number of times equal to  `max_challenges_per_generation`

(3) amount of data stored on blobber

~We will need to modify the blobber object to hold the active used capacity of the blobber and a rolling total of reads during the last P rounds.~
[commitBlobberRead](https://github.com/0chain/0chain/blob/staging/code/go/0chain.net/smartcontract/storagesc/blobber.go#L344) increment details of each read against the blobber.
[commitBlobberConnection](https://github.com/0chain/0chain/blob/staging/code/go/0chain.net/smartcontract/storagesc/blobber.go#L556) adjuct total data stored, either up or down, agasit the blobber.
Total data stored is an `int64` in StorageNode. 
The number of blobber reads can be an array of ints of size 2*period. Each read is only active for `period length` rounds after the read, so increment the matching `period length` array elements after each read.

Sriep:
As well as emitting to the event database so the blobber can pick it up, challenge data needs to be saved in 0chain for `verifyChalelnge` and ` cancelAllocation` and `finalizeAllocation`

`cancelAllocation` needs to know how many open challenges there are and what proportion of them have expired. In order to calculate a pass rate.

[...]

PR #963, opened 2022-02-01
Blobber Block reward

### LINKED ISSUE #753: Blobber block reward lottery
# Block Rewards

We want to reward blobbers based on their total stake and recent challenge history. So each block we will run a smart contract that will reward split a fixed sum between a random collection of blobbers and their delegates. Weighting the payments by stake and recent successful challenges.

# Smart contract

1. Blobber are partitioned 
2. Select a partition at random
3. Weight the blobbers in the randomly selected partition
4. Calculate reward amounts for randomly selected blobbers. Divide up the blobber's block reward between blobbers and their delegates, taking account of weighting.
5. Rewards added to Blobber's delegate's reward object.

 The weight will be
```
challenges completed in last period * stake * write price
```

New keys to be added to sc.yaml
- `block_reward.challenge_period` - period after which block reward should be computed
- `block_reward.block_reward_change_period`- period after which block reward will reduce
- `block_reward.block_reward_change_ratio` - by how much ratio will the block reward be reduced after every block reward change period
stake is blobber's total stake excluding unstaking tokens.

The block reward amount should be flexible. For example, we will likely want it to reduce by a set proportion year on year.

ssardana96:

Tasks done:
- created a blobber_reward partition
- this blobber_reward partition follows 2 different MPT: 
  + (ongoing_blobber_reward) - this is blobber challenges for current challenge period
  + (active_blobber_reward) - this is blobber challenges for previous challenge period
- change in verifyChallenge to populate these partition
- added blobber block reward functionality
- blobber_block_reward to be triggerred every challenge_period, and not every round
  
...


code/go/0chain.net/smartcontract/storagesc/challenge.go
```diff
@@ -330,6 +368,34 @@ func (sc *StorageSmartContract) verifyChallenge(t *transaction.Transaction,
...
+				"cannot delete active partition: "+err.Error())
+		}
+
+		// move ongoing_passed_blobbers to active_passed_blobbers
+		ongoingList, err := getOngoingPassedBlobbersList(balances)
+		if err != nil {
+			return "", common.NewError("verify_challenge",
+				"cannot get ongoing partition: "+err.Error())
+		}
+		err = ongoingList.Migrate(ACTIVE_PASSED_BLOBBERS_KEY, balances)
```
Sriep: I am not sure what Migrate does but it looks far too complicated. Can't you just do
```go
_, err = balances.InsertTrieNode(ACTIVE_PASSED_BLOBBERS_KEY, ongoingList)
_, err = balances.InsertTrieNode(ONGOING_PASSED_BLOBBERS_KEY, emptyPartition_plus_one_pass_from_this_call)
```

code/go/0chain.net/smartcontract/storagesc/challenge.go
```diff
@@ -406,17 +472,66 @@ func (sc *StorageSmartContract) verifyChallenge(t *transaction.Transaction,
...
 	// verification, or partial verification
 	if pass && fresh {
 
+		blobber, err := sc.getBlobber(t.ClientID, balances)
+		if err != nil {
+			return "", common.NewError("verify_challenge",
+				"can't get blobber"+err.Error())
+		}
+
+		ogList, err := getOngoingPassedBlobbersList(balances)
```
Sriep: Access each MPT key only once. You called `getOngoingPassedBlobbersList` above.

Sriep (commented):
We still need to decide how to integrate time in seconds with time in rounds.

code/go/0chain.net/smartcontract/storagesc/challenge.go
ssardana96: @Sriep what about the list stored in ongoingList, they also needs to be migrated no?
else the keys will start clashing in future rounds
Partition.Key()

code/go/0chain.net/smartcontract/storagesc/challenge.go
ssardana96: the reason I'm calling again is because the older one has been migrated and deleted in .Migrate()
this will fetch the latest partition

code/go/0chain.net/smartcontract/storagesc/challenge.go
Sriep: The partition name is part of the key for all the individual portions MPT keys.

If you ensure that the block number is part of the overall partition's name then that will be inherited by the key to the individual partitions. So there will be no key cllashing.

code/go/0chain.net/smartcontract/storagesc/challenge.go
Sriep: If it has been deleted then you already know that. Use a local variable. 

MPT access takes hundreds of microseconds not nanoseconds. There should be no reason to read or save the same key more than once.

code/go/0chain.net/smartcontract/storagesc/challenge.go
ssardana96: superb. let me do that

code/go/0chain.net/smartcontract/storagesc/block_reward.go
```diff
@@ -15,57 +18,71 @@ func (ssc *StorageSmartContract) payBlobberBlockRewards(
...
+		err = blobber.Decode(b.Encode())
+		if err != nil {
+			return common.NewError("blobber_block_rewards_failed",
+				"can't decode blobber reward node: "+err.Error())
+		}
+		if sp, err = ssc.getStakePool(b.Name(), balances); err != nil {
 			return common.NewError("blobber_block_rewards_failed",
 				"can't get related stake pool: "+err.Error())
 		}
+
```
Sriep, line 135: you can use [sp.stake](https://github.com/0chain/0chain/blob/staging/code/go/0chain.net/smartcontract/storagesc/stakepool.go#L212) to get the total stake in the stake pool.

code/go/0chain.net/smartcontract/storagesc/block_reward.go
```diff
@@ -15,57 +18,71 @@ func (ssc *StorageSmartContract) payBlobberBlockRewards(
...
-		if totalQStake > 0 {
-			ratio = stakeTotals[i] / totalQStake
-		} else {
-			ratio = 1.0 / float64(len(stakePools))
-		}
+		reward := float64(conf.BlockReward.BlockReward) * (weight[i] / totalWeight)
 
-		capacityReward := float64(conf.BlockReward.BlockReward) * conf.BlockReward.BlobberCapacityWeight * ratio
-		if err := mintReward(qsp, capacityReward, balances); err != nil {
+		if err := mintReward(qsp, reward, balances); err != nil {
```
Sriep: This will be replaced. We are not going to mint tokens but save to the stake pool rewards using [DistributeRewards](https://github.com/0chain/0chain/blob/358a22285cef5f6782ef78c925295cf266c435d9/code/go/0chain.net/smartcontract/stakepool/stakepool.go#L190). Something like
```go
err = qsp.DistributeRewards(reward, qualifyingBlobberIds[i], stakepool.Blobber, balances)
```
This change requires https://github.com/0chain/0chain/pull/967/

Sriep (changes requested):

code/go/0chain.net/smartcontract/storagesc/block_reward.go
ssardana96: done

code/go/0chain.net/smartcontract/storagesc/block_reward.go
ssardana96: cool. will make this change once #967 is merged to staging
or should I make this branch merge to #967 ?

service-0chain:
| config  | value |
| tests skipped  | FALSE  |
| system-tests  | master  |
| gosdk  | NOT OVERRIDDEN  |
| zboxcli  | staging  |
| zwalletcli  | staging  |
| 0chain  | bbr  |
| blobber  | staging  |
| authorizer  | staging  |
| 0box  | staging  |
| 0block  | staging  |
| 0dns  | staging  |
| explorer  | staging  |
| 0proxy  | staging  |
| 0search  | staging  |

code/go/0chain.net/miner/protocol_block_main.go
```diff
@@ -293,7 +293,8 @@ func (mc *Chain) generateBlock(ctx context.Context, b *block.Block,
 		}
 	}
 
-	if config.DevConfiguration.IsBlockRewards {
+	if config.DevConfiguration.IsBlockRewards &&
+		b.Round%config.SmartContractConfig.GetInt64("smart_contracts.storagesc.block_reward.challenge_period") == 0 {
```
peterlimg: Instead of getting the setting from the viper config directly, we'd better put the `challenge period` setting in the scConfig and get it from there.

code/go/0chain.net/smartcontract/partitions/blobber_list.go
```diff
@@ -0,0 +1,162 @@
...
+	"0chain.net/core/util"
+	"encoding/json"
+	"errors"
+	"fmt"
+)
+
+//------------------------------------------------------------------------------
+
+type BlobberNode struct {
+	Id  string `json:"id"`
```
peterlimg: Id -> ID, to pass the lint, https://github.com/golang/go/wiki/CodeReviewComments#initialisms

code/go/0chain.net/smartcontract/partitions/blobber_reward.go
```diff
@@ -0,0 +1,176 @@
...
+	"0chain.net/core/util"
+	"encoding/json"
+	"errors"
+	"fmt"
+)
+
+//------------------------------------------------------------------------------
+
+type BlobberRewardNode struct {
+	Id                string         `json:"id"`
```
peterlimg: Id -> ID

code/go/0chain.net/smartcontract/partitions/random.go
```diff
@@ -205,6 +207,14 @@ func (rs *randomSelector) addPartition() PartitionItemList {
 		newPartition = &itemList{
 			Key: rs.partitionKey(rs.NumPartitions),
 		}
+	} else if rs.ItemType == ItemBlobber {
+		newPartition = &blobberItemList{
+			Key: rs.partitionKey(rs.NumPartitions),
+		}
+	} else if rs.ItemType == ItemBlobberReward {
```
peterlimg, line 221: Partition package should not be coupled with the concrete node types.

code/go/0chain.net/smartcontract/storagesc/models.go
```diff
@@ -301,18 +349,28 @@ func (sng StorageNodeGeolocation) validate() error {
...
+	Capacity          int64                  `json:"capacity"`          // total blobber capacity
+	Used              int64                  `json:"used"`              // allocated capacity
+	BytesWritten      int64                  `json:"dataWritten"`       // in bytes
+	DataReadLR        float64                `json:"dataRead"`          // in GB
+	DRUpdateLastRound int64                  `json:"dataReadLastRound"` // last round when data read was updated
+	LastHealthCheck   common.Timestamp       `json:"last_health_check"`
+	PublicKey         string                 `json:"-"`
 	// StakePoolSettings used initially to create and setup stake pool.
 	StakePoolSettings stakePoolSettings `json:"stake_pool_settings"`
+	RewardPartition   partitionLocation `json:"rewardPartition"`
```
peterlimg: Use underscore json tag

code/go/0chain.net/smartcontract/storagesc/challenge.go
```diff
@@ -324,6 +372,56 @@ func (sc *StorageSmartContract) verifyChallenge(t *transaction.Transaction,
...
+
+		if balances.GetBlock().Round != 0 {
+
+			ongoingList, err = getOngoingPassedBlobbersList(balances, startRound-conf.BlockReward.ChallengePeriod)
+			if err != nil {
+				return "", common.NewError("verify_challenge",
+					"cannot get ongoing partition: "+err.Error())
+			}
+
+			_, err = balances.InsertTrieNode(ACTIVE_PASSED_BLOBBERS_KEY, ongoingList)
```
peterlimg: If the active blobbers is a copy of the ongoingList and we can get it base on the  key calculated from start round, please do not save it to MPT again as a new 'active' list.  

So basically the `ongoing` and `active` blobber lists are logic concept, we should not put the the names into the MPT node key, and should avoid saving the duplicate data into MPT. 

What we should instead are having two functions:
func getOngoingPassedBlobbersList() {
   ...
   // the partition of current period
}

func getActivePasssedBlobbersList() {
  // get partition of previous period
}

There's no point to get the previous period partition and save it into a new partition that named `active`.

code/go/0chain.net/smartcontract/storagesc/blobber.go
```diff
@@ -452,6 +485,51 @@ func (sc *StorageSmartContract) commitBlobberRead(t *transaction.Transaction,
 	details.ReadReward += value // stat
 	details.Spent += value      // reduce min lock demand left
 
+	startRound := getStartRound(balances.GetBlock().Round, conf.BlockReward.ChallengePeriod)
+
+	if blobber.DRUpdateLastRound >= startRound {
+		blobber.DataReadLR += sizeRead
+	} else {
+		blobber.DataReadLR = sizeRead
```
peterlimg: If  the DataReadLR will be reset for each period, then we should save it in reward node. And we can remove the DRUpdateLastRound from blobber node as well.

code/go/0chain.net/core/maths/maths.go
```diff
@@ -0,0 +1,26 @@
+package maths
+
+import "math"
+
+// GetGamma gets gamma for blobber block reward
+// A, B, alpha are constants
+// X is total data stored in blobber
+// R is data read by blobber
+func GetGamma(A, B, alpha, X, R float64) float64 {
```
peterlimg, line 9: What's the idea behind these Gama and Zeta, please link to document or paper properly to help us understand it, thanks.

code/go/0chain.net/smartcontract/storagesc/block_reward.go
```diff
@@ -4,68 +4,103 @@ import (
...
+	const (
+		// constants for gamma
+		alpha = 1
+		A     = 1
+		B     = 1
+
+		// constants for zeta
+		I  = 1
+		K  = 1
+		mu = 1
```
peterlimg, line 44: If the above constants are used by functions defined in maths package, please define them in there too. Otherwise, users would have to define these in their own packages everywhere.

code/go/0chain.net/smartcontract/storagesc/block_reward.go
```diff
@@ -4,68 +4,103 @@ import (
...
 			"cannot get all blobbers list: "+err.Error())
 	}
 
-	// filter out blobbers with stake too low to qualify for rewards
-	var qualifyingBlobberIds []string
-	var stakePools []*stakePool
-	var stakeTotals []float64
-	var totalQStake float64
-	for _, blobber := range allBlobbers.Nodes {
+	random := rand.New(rand.NewSource(time.Now().UTC().UnixNano()))
```
peterlimg: This random will make the block reward transaction got different results when it is executed on different miners. All miners should have the same random value so that the same blobbers will be selected for rewards. 

Please check how round random seed is generated and used to select the block generators.

code/go/0chain.net/smartcontract/storagesc/block_reward.go
```diff
@@ -84,3 +119,9 @@ func (ssc *StorageSmartContract) payBlobberBlockRewards(
 
 	return nil
 }
+
+func getBlockReward(br state.Balance, currentRound, brChangePeriod int64, brChangeRatio, blobberWeight float64) float64 {
```
peterlimg: The conf in line 114 is not changed, so we don't have to save into MPT again.

code/go/0chain.net/smartcontract/storagesc/models.go
```diff
@@ -145,6 +148,28 @@ func (sn *BlobberChallenge) addChallenge(challenge *StorageChallenge) bool {
...
+	now common.Timestamp, lastNDays int64, cct time.Duration) (count int) {
+	fromTimestamp := common.Timestamp(common.ToTime(now).Add(time.Duration(lastNDays) * time.Hour * -24).Unix())
+
+	if sn.LatestCompletedChallenge.Created < fromTimestamp {
+		return
+	}
+
+	for i := len(sn.Challenges) - 1; i >= 0; i++ {
+		if sn.Challenges[i].Created < fromTimestamp {
+			return
```
peterlimg: Return or continue?

code/go/0chain.net/smartcontract/storagesc/models.go
```diff
@@ -145,6 +148,28 @@ func (sn *BlobberChallenge) addChallenge(challenge *StorageChallenge) bool {
...
+		}
+		success, failure := sn.Challenges[i].getStats()
+		pass, _, _ := sn.Challenges[i].isChallengePassed(success, failure, cct, now)
+		if pass == isSuccess {
```
peterlimg: Is it possible the `isSuccess` false? otherwise, we don't need this param I think.

code/go/0chain.net/smartcontract/storagesc/models.go
```diff
@@ -157,6 +182,29 @@ type StorageChallenge struct {
...
 }
 
+func (sc *StorageChallenge) isChallengePassed(success, failure int,
+	challengeCompletionTime time.Duration, now common.Timestamp) (pass, fresh bool, threshold int) {
+
+	cct := toSeconds(challengeCompletionTime)
+	threshold = len(sc.Validators) / 2
+	pass = success > threshold ||
+		(success > failure && success+failure < threshold)
+	fresh = sc.Created+cct >= now
```
peterlimg: We should always require success > threshold

code/go/0chain.net/smartcontract/storagesc/models.go
```diff
@@ -301,18 +349,28 @@ func (sng StorageNodeGeolocation) validate() error {
 	return nil
 }
 
+type partitionLocation struct {
```
peterlimg: I guess this is for blobber partition location only, if so, please do not use the common name (partition location) for it, rename it to properly such as `RewardPartitionLocation`.

code/go/0chain.net/smartcontract/storagesc/models.go
```diff
@@ -301,18 +349,28 @@ func (sng StorageNodeGeolocation) validate() error {
...
-	LastHealthCheck common.Timestamp       `json:"last_health_check"`
-	PublicKey       string                 `json:"-"`
+	ID                string                 `json:"id"`
+	BaseURL           string                 `json:"url"`
+	Geolocation       StorageNodeGeolocation `json:"geolocation"`
+	Terms             Terms                  `json:"terms"`             // terms
+	Capacity          int64                  `json:"capacity"`          // total blobber capacity
+	Used              int64                  `json:"used"`              // allocated capacity
+	BytesWritten      int64                  `json:"dataWritten"`       // in bytes
+	DataReadLR        float64                `json:"dataRead"`          // in GB
```
peterlimg: Rename to DataReadLastRound if this is needed, so as the json tag to:  data_read_last_round

code/go/0chain.net/smartcontract/storagesc/models.go
```diff
@@ -301,18 +349,28 @@ func (sng StorageNodeGeolocation) validate() error {
...
+	DRUpdateLastRound int64                  `json:"dataReadLastRound"` // last round when data read was updated
```
peterlimg: Rename to DataReadUpdateLastRound if this is needed, json tag to: data_read_update_last_round

peterlimg (changes requested):
There are three main issues:
- The active and ongoing blobber reward lists, they should be logic concept, the name should not be insert into the MPT as part of the key. Also, we should not save the previous period data into MPT again as the `active` list.
- The random value generated in blobberBlockRewards is incorrect, it would lead to different transaction result get returned when it is executed on different miners/sharders.
- https://0chain.slack.com/archives/C020E8HA8TE/p1644062465226779?thread_ts=1643724926.281819&cid=C020E8HA8TE. The `partition` package should be a common package that we can use it easily to support partition, see the comments . Otherwise, we will have to do the things like the new added `blobber_list.go` and `blobber_reward.go`, basicaily most of the code there are duplicate work, we should avoid doing this. Anyway, this could be refactored later.

code/go/0chain.net/core/maths/maths.go
Sriep: Maybe you could take the pdf from slack, https://0chain.slack.com/archives/C02MT6RV6QP/p1644236441266629, and add it to `0chain/docs` then you can reference it directly.

code/go/0chain.net/smartcontract/storagesc/block_reward.go
Sriep: https://github.com/0chain/0chain/blob/staging/code/go/0chain.net/smartcontract/storagesc/challenge.go#L570
Here the current transaction hash and previous block hash are used to generate the seed. You could turn this code from `generateChallenges` into a common function and use that.

code/go/0chain.net/smartcontract/storagesc/models.go
```diff
@@ -145,6 +148,28 @@ func (sn *BlobberChallenge) addChallenge(challenge *StorageChallenge) bool {
 	return false
 }
 
+func (sn *BlobberChallenge) getSuccessCount(isSuccess bool,
```
Sriep: This method does not seem to be used. If so, it is probably best to remove it.

code/go/0chain.net/smartcontract/storagesc/models.go
```diff
@@ -157,6 +182,29 @@ type StorageChallenge struct {
...
+	cct := toSeconds(challengeCompletionTime)
+	threshold = len(sc.Validators) / 2
+	pass = success > threshold ||
+		(success > failure && success+failure < threshold)
+	fresh = sc.Created+cct >= now
+
+	return
+}
+
+func (sc *StorageChallenge) getStats() (success, fail int) {
```
Sriep: `getStats` is only called from `getSuccessCount` which appears to be never used. So this code is unreachable. If so we should remove it.

code/go/0chain.net/smartcontract/storagesc/models.go
```diff
@@ -157,6 +182,29 @@ type StorageChallenge struct {
 	Response       *ChallengeResponse `json:"challenge_response,omitempty"`
 }
 
+func (sc *StorageChallenge) isChallengePassed(success, failure int,
+	challengeCompletionTime time.Duration, now common.Timestamp) (pass, fresh bool, threshold int) {
```
Sriep: `isChallengePassed` can only be accessed from `verifyChallenge`. It does help to make the code more readable, however, the `BlobberChallege` is too big and will likely be removed as part of the concurrent changes for https://github.com/0chain/0chain/issues/741.  

So this code will be removed shortly. I think it would be best to not make unneeded changes to the `StorageChallenge` or `BlobberChallenge` object to reduce merging issues in the near future.

code/go/0chain.net/smartcontract/storagesc/block_reward.go
Sriep: These six constants need to be added to smart contract settings read from `sc.yaml` Check slack thread. For six parameters it might be worth creating a separate object that can be passed to the maths package.

code/go/0chain.net/smartcontract/storagesc/block_reward_test.go
```diff
@@ -1,286 +1,124 @@
...
-				blobberStakes:        [][]float64{{10}, {5, 5, 10}, {2, 2}},
-				blobberServiceCharge: []float64{0.1, 0.2, 0.3},
-				blockReward:          100,
-				qualifyingStake:      10.0,
-				blobberCapacityRatio: 5,
-				blobberUsageRatio:    10,
-				minerRatio:           2,
-				sharderRatio:         3,
+			name: "1_blobber",
+			params: params{
```
Sriep, line 91: Once we have added the calculation parameters to sc.yaml settings, we should have a more realistic test case. With a full partition of blobbers, and add (gamma, a, b, i, k mu) to the `parmas` object. This could allow a user to experiment with different values.

docker.local/config/sc.yaml
```diff
@@ -149,11 +149,13 @@ smart_contracts:
...
       block_reward: 1
+      block_reward_change_period: 10000
+      block_reward_change_ratio: 0.1
       qualifying_stake: 1
       sharder_ratio: 10
       miner_ratio: 40
-      blobber_capacity_ratio: 10
-      blobber_usage_ratio: 40
+      blobber_ratio: 50
+      challenge_period: 30
```
Sriep: Add calculation parameters (gamma, a, b, i, k, mu). Check slack thread https://0chain.slack.com/archives/C02MT6RV6QP/p1644840052762349?thread_ts=1644236441.266629&cid=C02MT6RV6QP for labels.

Sriep (changes requested):
As [verifyChallenge](https://github.com/0chain/0chain/blob/staging/code/go/0chain.net/smartcontract/storagesc/challenge.go#L323) will be undergoing changes as part of the concurrent issue https://github.com/0chain/0chain/issues/741, it might be best to keep changes here minimal and focused on only what is needed for block rewards. This is to reduce any short term merging issues.

code/go/0chain.net/smartcontract/partitions/blobber_list.go
Sriep: It seems [contentious](https://stackoverflow.com/questions/15526107/acronyms-in-camelcase) but we should be consistent. Also, it's the first time I have seen [CodeReviewComments](https://github.com/golang/go/wiki/CodeReviewComments#go-code-review-comments) thanks for the reference.

ssardana96: fixed. this is helpful @peterlimg

code/go/0chain.net/smartcontract/storagesc/models.go
ssardana96: @Sriep this was intended changes made for previous version of BBR. will remove them

code/go/0chain.net/smartcontract/storagesc/models.go
ssardana96: @peterlimg this was existing code, which I moved to a function as this was needed earlier.
reverting it back to older state

code/go/0chain.net/smartcontract/storagesc/models.go
ssardana96: this function is not needed

code/go/0chain.net/smartcontract/storagesc/blobber.go
ssardana96: @peterlimg we add rewardnode to partition only when challenge is passed.
What if read happens first, the rewardNode won't be in partition
but once challenge comes later on, then the reward node would be created and we would need this data

code/go/0chain.net/smartcontract/storagesc/models.go
ssardana96: not required function

code/go/0chain.net/smartcontract/storagesc/models.go
ssardana96: removed

code/go/0chain.net/smartcontract/storagesc/models.go
ssardana96: removed

code/go/0chain.net/smartcontract/storagesc/challenge.go
ssardana96: fixed @peterlimg 
can you check the new changes for this

code/go/0chain.net/core/maths/maths.go
peterlimg: Thanks, I'll have a check.

code/go/0chain.net/smartcontract/storagesc/challenge.go
peterlimg: Thanks, I'll have a check

code/go/0chain.net/smartcontract/storagesc/blobber.go
peterlimg: Oh, yes, that makes scene. Blobber would have to keep the read data before it pass the challenge. 

One question: 
Do we have mechanism to ensure that all blobbers will have the chance to do verifyChallenge in a block reward period. So if a blobber could pass the challenge itself, but it just didn't get the chance, the code about seems would reset the `data read` of the blobber, should we reset it or accumulate the `data read` for next reward period? @Sriep @ssardana08

code/go/0chain.net/smartcontract/storagesc/block_reward.go
peterlimg: Yes, that should work as well and easy to go.

code/go/0chain.net/miner/protocol_block_main.go
```diff
@@ -293,7 +293,8 @@ func (mc *Chain) generateBlock(ctx context.Context, b *block.Block,
 		}
 	}
 
-	if config.DevConfiguration.IsBlockRewards {
+	if config.DevConfiguration.IsBlockRewards &&
+		b.Round%config.SmartContractConfig.GetInt64("smart_contracts.storagesc.block_reward.trigger_period") == 0 {
```
peterlimg: block_reward.period seems better, also get it from storagesc config variable instead of reading from viper.

code/go/0chain.net/chaincore/chain/state/state_context.go
```diff
@@ -58,6 +58,8 @@ type StateContextI interface {
 	EmitError(error)
 	GetEvents() []event.Event   // cannot use in smart contracts or REST endpoints
 	GetEventDB() *event.EventDb // do not use in smart contracts can use in REST endpoints
+	GetCurrentRewardRound(period int64) int64
+	GetPreviousRewardRound(period int64) int64
```
Sriep: I cannot see why these two need to be in StateContextI. Cannot you just have a local function in stroagesc?
```go
func  GetCurrentRewardRound(period int64, balances cstate.StateContextI) int64 {
	extra := balances.GetBlock().Round % period
	return balances.GetBlock().Round - extra
}
```
Adding to StateContextI creates a maintenance overhead as it is used in so many places.

Sriep (changes requested):

code/go/0chain.net/smartcontract/storagesc/block_reward.go
```diff
@@ -4,68 +4,111 @@ import (
...
 
-	// filter out blobbers with stake too low to qualify for rewards
-	var qualifyingBlobberIds []string
-	var stakePools []*stakePool
-	var stakeTotals []float64
-	var totalQStake float64
-	for _, blobber := range allBlobbers.Nodes {
+	hashString := encryption.Hash(balances.GetTransaction().Hash + balances.GetBlock().PrevHash)
+	var randomSeed uint64
+	randomSeed, err = strconv.ParseUint(hashString[0:16], 16, 64)
```
peterlimg: Use `ParseInt` so we don't  have to convert it to int64 below

code/go/0chain.net/smartcontract/utils/utils.go
```diff
@@ -0,0 +1,11 @@
+package utils
```
peterlimg: These functions are blobber reward related, so it's better to put them in the package where blobber or reward structs are defined . `utils` is not a good package name actually, check the "Bad package names" section on https://go.dev/blog/package-names

peterlimg (changes requested):

code/go/0chain.net/smartcontract/storagesc/block_reward.go
Sriep: https://github.com/0chain/0chain/pull/967 has been merged, so we can now make the change; replace `mintReward` with `DistrbuteRewards`.

code/go/0chain.net/miner/protocol_block_main.go
ssardana96: but @peterlimg we don't have access to storagesc.scConfig from here.

code/go/0chain.net/smartcontract/storagesc/block_reward_test.go
ssardana96: to keep as constant for now

docker.local/config/sc.yaml
ssardana96: to be kept as it is for now

code/go/0chain.net/miner/protocol_block_main.go
ssardana96: I've changed it to block_reward.trigger_period @peterlimg 
on storagesc config we don't have access from here. Please suggest!

code/go/0chain.net/smartcontract/utils/utils.go
ssardana96: done

code/go/0chain.net/smartcontract/storagesc/block_reward.go
```diff
@@ -2,10 +2,149 @@ package storagesc
...
+		stakeTotals          []float64
+		totalQStake          float64
+		weight               []float64
+		totalWeight          float64
+		conf                 *scConfig
+	)
+
+	const (
+		// constants for gamma
+		alpha = 1
```
ssardana08, line 31: @Sriep for now it's okay to keep these constant?

code/go/0chain.net/smartcontract/storagesc/challenge.go
ssardana96: @peterlimg I have already made these changes long back.
if these are looking as expected can you mark this resolved?

code/go/0chain.net/smartcontract/storagesc/blobber.go
ssardana96: I believe we care only for what happened in the last period and hence reset.
@Sriep can comment more..

code/go/0chain.net/smartcontract/storagesc/block_reward.go
ssardana96: for now we're keeping these as it is, right @Sriep ?

code/go/0chain.net/smartcontract/partitions/random.go
ssardana96: but @peterlimg partitions are of different for different itemtype
hence using it

peterlimg: NVM, we can refactor this later. Anyway, I think each type of partition should has it's own RandomSelector instance, therefore, each RandomSelector can run related interface to generate its own type of partition item list instead of using the `if else..` here.

code/go/0chain.net/miner/protocol_block_main.go
peterlimg: Add method to the StateContextI to get the config, the config should be read from MPT, it could be different from the value read from the viper. We should always use the loaded config instead of reading it from viper directly.

code/go/0chain.net/smartcontract/storagesc/block_reward.go
peterlimg: Maybe add a TODO comment or create an issue to track this

code/go/0chain.net/smartcontract/partitions/blobber_list.go
```diff
@@ -0,0 +1,143 @@
...
+		}
+	}
+	if err := il.Decode(val.Encode()); err != nil {
+		return fmt.Errorf("%w: %s", common.ErrDecoding, err)
+	}
+	il.Key = key
+	return nil
+}
+
+func (il *blobberItemList) add(it PartitionItem) {
```
peterlimg: We should check if the item does exist in the list before appending it, otherwise we could have duplicate items.

code/go/0chain.net/smartcontract/storagesc/block_reward.go
```diff
@@ -2,10 +2,126 @@ package storagesc
...
+		weight = append(weight, blobberWeight)
+		totalweight += blobberWeight
+	}
+
+	if totalweight == 0 {
+		totalweight = 1
+	}
+
+	for i, qsp := range stakePools {
+		reward := float64(conf.BlockReward.BlockReward) * (weight[i] / totalweight)
```
peterlimg: Check if  (weight[i] / totalWeight) > 0 and < 1

code/go/0chain.net/smartcontract/storagesc/block_reward.go
```diff
@@ -2,10 +2,126 @@ package storagesc
...
+		totalDelegateRewards := reward
+		for id, delegate := range qsp.Pools {
+			var ratio float64
+			if stakeTotals[i] > 0 {
+				ratio = float64(delegate.Balance) / stakeTotals[i]
+			} else {
+				ratio = 0
+			}
+			dReward := float64(conf.BlockReward.BlockReward) * weight[i] * ratio
```
peterlimg: explicitly check if ratio > 0 and < 1

code/go/0chain.net/smartcontract/storagesc/block_reward.go
```diff
@@ -2,10 +2,106 @@ package storagesc
...
+		return nil
+	}
+
+	allBlobbers, err := getBlobbersList(balances)
+	if err != nil {
+		return common.NewError("blobber_block_rewards_failed",
+			"cannot get all blobbers list: "+err.Error())
+	}
+
+	r := rand.New(rand.NewSource(time.Now().UTC().UnixNano()))
```
peterlimg: Any reason for changing back to use  time.Now() as random source?

code/go/0chain.net/smartcontract/partitions/blobber_list.go
```diff
@@ -0,0 +1,162 @@
...
+	il.Items = append(il.Items, BlobberNode{
+		Id:  it.Name(),
+		Url: it.Data(),
+	})
+	il.Changed = true
+}
+
+func (il *blobberItemList) update(it PartitionItem) error {
+	var found bool
+	for i := range il.itemRange(0, il.length()) {
```
peterlimg: Why not just iterate through the item list, the itemRange will copy the list of item pointers, which seems like is not necessary

code/go/0chain.net/smartcontract/partitions/itemlist.go
```diff
@@ -55,6 +56,21 @@ func (il *itemList) add(it PartitionItem) {
 	il.Changed = true
 }
 
+func (il *itemList) update(it PartitionItem) error {
+	var found bool
+	for i := range il.itemRange(0, il.length()) {
+		if il.Items[i].Name() == it.Name() {
+			found = true
+			il.Items[i] = StringItem{it.Name()}
```
peterlimg, line 66: we can set Changed=true and return nil here. And return the item not found error out the loop. No need to use the `found` variable, and should always return error early.

code/go/0chain.net/smartcontract/partitions/random.go
```diff
@@ -320,3 +335,84 @@ func (rs *randomSelector) Decode(b []byte) error {
...
+
+			_, err = balances.DeleteTrieNode(rs.Name)
+			if err != nil {
+				return err
+			}
+		} else {
+			return err
+		}
+	} else {
+		return errors.New("partition to migrate to is already populated")
```
peterlimg: Check and returns error early to avoid this if else ... else mode

code/go/0chain.net/smartcontract/partitions/random.go
```diff
@@ -320,3 +335,84 @@ func (rs *randomSelector) Decode(b []byte) error {
...
+			if err != nil {
+				return fmt.Errorf("error inserting node: %v", err)
+			}
+
+			_, err = balances.DeleteTrieNode(rs.Name)
+			if err != nil {
+				return err
+			}
+		} else {
+			return err
```
peterlimg: Check and returns error early to avoid this if else ... else mode

code/go/0chain.net/smartcontract/partitions/validator_list.go
```diff
@@ -120,6 +121,24 @@ func (il *validatorItemList) add(it PartitionItem) {
...
 }
 
+func (il *validatorItemList) update(it PartitionItem) error {
+	var found bool
+	for i := range il.itemRange(0, il.length()) {
+		if il.Items[i].Name() == it.Name() {
+			found = true
+			il.Items[i] = ValidationNode{
+				Id:  it.Name(),
+				Url: it.Data(),
```
peterlimg: Return here with Changed=true, remove found variable

code/go/0chain.net/smartcontract/storagesc/block_reward.go
```diff
@@ -2,10 +2,102 @@ package storagesc
...
+
+	blobberPartition, err := allBlobbers.GetRandomSlice(r, balances)
+	if err != nil {
+		return common.NewError("blobber_block_rewards_failed",
+			"Error getting random partition: "+err.Error())
+	}
+
+	// filter out blobbers with stake too low to qualify for rewards
+	var qualifyingBlobberIds []string
```
peterlimg: create the qulifyingBlobberIds slice with capacity of blobberPartition length

code/go/0chain.net/smartcontract/storagesc/block_reward.go
```diff
@@ -2,10 +2,119 @@ package storagesc
...
+		weight = append(weight, blobberWeight)
+		totalWeight += blobberWeight
+	}
+
+	if totalWeight == 0 {
+		totalWeight = 1
+	}
+
+	for i, qsp := range stakePools {
+		reward := bbr * (weight[i] / totalWeight)
```
peterlimg: check explicitly to ensure the weight[i] / totalWeight is > 0 and < 1

code/go/0chain.net/smartcontract/storagesc/block_reward.go
```diff
@@ -2,10 +2,119 @@ package storagesc
...
+	if err != nil {
+		return common.NewError("blobber_block_rewards_failed",
+			"saving configurations: "+err.Error())
+	}
+
 	return nil
 }
+
+func getBlockReward(br state.Balance, currentRound, brChangePeriod int64, brChangeRatio, blobberWeight float64) float64 {
+	changeBalance := 1 - brChangeRatio
```
peterlimg, line 191: assert the changeBalance is > 0 and < 1

code/go/0chain.net/smartcontract/storagesc/block_reward.go
```diff
@@ -2,10 +2,149 @@ package storagesc
...
+	changePeriods := currentRound % brChangePeriod
+	return float64(br) * math.Pow(changeBalance, float64(changePeriods)) * blobberWeight
+}
+
+func GetCurrentRewardRound(currentRound, period int64) int64 {
+	if period > 0 {
+		extra := currentRound % period
+		if currentRound >= extra {
```
peterlimg: Is it possible currentRound < extra? it could happen only when it is < 0, but we should avoid it.

code/go/0chain.net/smartcontract/storagesc/blobber.go
```diff
@@ -461,6 +494,51 @@ func (sc *StorageSmartContract) commitBlobberRead(t *transaction.Transaction,
 	details.ReadReward += value // stat
 	details.Spent += value      // reduce min lock demand left
 
+	rewardRound := GetCurrentRewardRound(balances.GetBlock().Round, conf.BlockReward.TriggerPeriod)
+
+	if blobber.LastRewardRoundOfDataRead >= rewardRound {
+		blobber.DataReadLastRewardRound += sizeRead
+	} else {
+		blobber.DataReadLastRewardRound = sizeRead
```
peterlimg, line 421: It's a little confuse to read the DataReadLastRewardRound, not sure if it is a round number or the data size from the name. So maybe we can change it to LastRewardDataReadSize for size. And change the LastRewardRoundOfDataRead to LastRewardDataReadRound for round.

code/go/0chain.net/miner/protocol_block.go
```diff
@@ -776,7 +776,9 @@ func (mc *Chain) generateBlock(ctx context.Context, b *block.Block,
 		}
 	}
 
-	if config.DevConfiguration.IsBlockRewards {
+	if config.DevConfiguration.IsBlockRewards &&
+		b.Round%config.SmartContractConfig.GetInt64("smart_contracts.storagesc.block_reward.trigger_period") == 0 {
```
peterlimg, line 844: Get config from MPT, anyway, should avoid read from file directly.

peterlimg (changes requested):

code/go/0chain.net/smartcontract/partitions/random.go
ssardana96: got it.
@peterlimg can you create an issue on the exact thought you have in mind. We can pick it up later

code/go/0chain.net/smartcontract/storagesc/block_reward.go
ssardana96: added a TODO and created an issue: https://github.com/0chain/0chain/issues/1035

code/go/0chain.net/smartcontract/storagesc/block_reward.go
ssardana96: done

code/go/0chain.net/smartcontract/storagesc/block_reward.go
ssardana96: done

code/go/0chain.net/smartcontract/storagesc/block_reward.go
ssardana96: done

code/go/0chain.net/smartcontract/storagesc/blobber.go
ssardana96: yeah I know. I was myself thriving to get a better name. and hence added proper comments in model
LastRewardDataReadSize -  looks like the data size read in last reward, but as we're paying rewards in arrear, what we want to capture here is: what is the size read by blobber in last arrear reward round.

changed LastRewardDataReadRound

code/go/0chain.net/smartcontract/storagesc/block_reward.go
```diff
@@ -106,16 +113,19 @@ func (ssc *StorageSmartContract) blobberBlockRewards(
...
 	for i, qsp := range stakePools {
-		reward := bbr * (weight[i] / totalWeight)
-		logging.Logger.Info("blobber_block_rewards_pass",
-			zap.Float64("reward", reward),
-			zap.String("blobber id", qualifyingBlobberIds[i]))
-
-		if err := qsp.DistributeRewards(reward, qualifyingBlobberIds[i], stakepool.Blobber, balances); err != nil {
-			return common.NewError("blobber_block_rewards_failed", "minting capacity reward"+err.Error())
+		weightRatio := weight[i] / totalWeight
+		if weightRatio > 0 && weightRatio <= 1 {
```
peterlimg, line 153: Invalid ratio should be reported or return as error if necessary.

code/go/0chain.net/smartcontract/storagesc/block_reward.go
```diff
@@ -106,16 +113,19 @@ func (ssc *StorageSmartContract) blobberBlockRewards(
...
+			reward := bbr * (weight[i] / totalWeight)
```
peterlimg: Use the `weightRatio` instead of do it again with `weight[i] / totalWeight`

peterlimg (changes requested):

code/go/0chain.net/smartcontract/partitions/blobber_list.go
ssardana96: done

code/go/0chain.net/smartcontract/partitions/itemlist.go
ssardana96: done

code/go/0chain.net/smartcontract/partitions/blobber_list.go
```diff
@@ -0,0 +1,159 @@
...
+		Url: it.Data(),
+	})
+	il.Changed = true
+}
+
+func (il *blobberItemList) update(it PartitionItem) error {
+	for i := range il.itemRange(0, il.length()) {
+		if il.Items[i].Name() == it.Name() {
+			var newItem BlobberNode
+			err := newItem.Decode(it.Encode())
```
peterlimg: We can do a type assertion and then do a copy to avoid unnecessary data encoding/decoding, such as:

bn, ok :=  (*it.BlobberNode)
if !ok{
  return err 
}

newItem :=  *bn
// add the newItem to list.

peterlimg (changes requested):

code/go/0chain.net/smartcontract/partitions/blobber_list.go
ssardana96: done

code/go/0chain.net/smartcontract/partitions/blobber_list.go
ssardana96: done

code/go/0chain.net/smartcontract/storagesc/block_reward.go
ssardana96: done

code/go/0chain.net/smartcontract/storagesc/block_reward.go
ssardana96: done

code/go/0chain.net/smartcontract/partitions/blobber_list.go
```diff
@@ -0,0 +1,165 @@
...
+		return fmt.Errorf("%w: %s", common.ErrDecoding, err)
+	}
+	il.Key = key
+	return nil
+}
+
+func (il *blobberItemList) add(it PartitionItem) {
+	for _, bi := range il.Items {
+		if bi.ID == it.Name() {
+			return
```
peterlimg: We'd better return error if the item already exist so that the caller could know what to do next. Imagine what could happen if we want to add a new item but the same name item does already exist in the list, then new item would not be inserted or updated to the exist one, then all following action against this new item will break. 

Same as other lists

code/go/0chain.net/smartcontract/storagesc/block_reward.go
```diff
@@ -2,10 +2,172 @@ package storagesc
...
+				zap.String("blobber id", qualifyingBlobberIds[i]))
+
+			if err := qsp.DistributeRewards(reward, qualifyingBlobberIds[i], stakepool.Blobber, balances); err != nil {
+				return common.NewError("blobber_block_rewards_failed", "minting capacity reward"+err.Error())
+			}
+
+			qsp.Reward += state.Balance(reward) // do we need to do this?
+		} else {
+			logging.Logger.Error("blobber_bloc_rewards - error in weight ratio",
+				zap.Any("stake pool", qsp))
```
peterlimg, line 167: We need to return error instead of continue for loop of reward calucation to avoid potential reward calculation issue.

peterlimg (changes requested):

code/go/0chain.net/smartcontract/storagesc/block_reward.go
ssardana96: done

code/go/0chain.net/smartcontract/partitions/blobber_list.go
ssardana96: done

code/go/0chain.net/smartcontract/storagesc/block_reward.go
```diff
@@ -2,10 +2,173 @@ package storagesc
...
+			reward := bbr * weightRatio
+			logging.Logger.Info("blobber_block_rewards_pass",
+				zap.Float64("reward", reward),
+				zap.String("blobber id", qualifyingBlobberIds[i]))
+
+			if err := qsp.DistributeRewards(reward, qualifyingBlobberIds[i], spenum.Blobber, balances); err != nil {
+				return common.NewError("blobber_block_rewards_failed", "minting capacity reward"+err.Error())
+			}
+
+			qsp.Reward += state.Balance(reward) // do we need to do this?
```
peterlimg: I found the stake pool service reward has already been handled in the qsp.DistributeRewards, and it is the  serviceChargeRatio*reward, should we give all the reward to it again, this could be wrong.

code/go/0chain.net/smartcontract/storagesc/challenge.go
```diff
@@ -407,6 +474,64 @@ func (sc *StorageSmartContract) verifyChallenge(t *transaction.Transaction,
...
+		}
+
+		// this expiry of blobber needs to be corrected once logic is finalized
+
+		if blobber.RewardPartition.StartRound != rewardRound ||
+			balances.GetBlock().Round == 0 {
+
+			var dataRead float64 = 0
+			if blobber.LastRewardDataReadRound >= rewardRound {
+				dataRead = blobber.DataReadLastRewardRound
```
peterlimg, line 477: Hmm, in what situation the blobber.ReadPatition.StartRound != rewardRound but the blobber.LastRewardDataReadRound  >= rewardRound could happen?

Perhaps we need to make sure the blobber.ReadPartition.StartRound and the LastRewardDataReadRound are updated correctly.

code/go/0chain.net/smartcontract/storagesc/models.go
```diff
@@ -303,19 +317,30 @@ func (sng StorageNodeGeolocation) validate() error {
...
-	LastHealthCheck common.Timestamp       `json:"last_health_check"`
-	PublicKey       string                 `json:"-"`
-	SavedData       int64                  `json:"saved_data"`
+	ID                      string                 `json:"id"`
+	BaseURL                 string                 `json:"url"`
+	Geolocation             StorageNodeGeolocation `json:"geolocation"`
+	Terms                   Terms                  `json:"terms"`       // terms
+	Capacity                int64                  `json:"capacity"`    // total blobber capacity
+	Used                    int64                  `json:"used"`        // allocated capacity
+	BytesWritten            int64                  `json:"dataWritten"` // in bytes
```
peterlimg: json tag dataWritten -> data_written

peterlimg (dismissed):

code/go/0chain.net/smartcontract/storagesc/block_reward.go
Sriep: The reward has already been paid by  `qsp.DistributeRewards`, so we should remove this line.

code/go/0chain.net/miner/protocol_block.go
ssardana96: to be done later

code/go/0chain.net/smartcontract/storagesc/challenge.go
ssardana96: when in this reward round period, data has been read by blobber, but verify_challenge has not been called yet for that blobber.

which means- in current reward round period, data has been accessed, but blobber has not yet been added to partition.

code/go/0chain.net/smartcontract/storagesc/block_reward.go
```diff
@@ -1,11 +1,180 @@
...
+		mu = 1
+	)
+
+	conf, err := ssc.getConfig(balances, true)
+	if err != nil {
+		return common.NewError("blobber_block_rewards_failed",
+			"cannot get smart contract configurations: "+err.Error())
+	}
+
+	if conf.BlockReward.BlockReward == 0 {
```
Sriep, line 53: When I run the benchmark program, block rewards exited here. We need to setup the block reward settings in [SetMockConfig](https://github.com/0chain/0chain/blob/staging/code/go/0chain.net/smartcontract/storagesc/benchmark_setup.go#L642)

Sriep (changes requested):

code/go/0chain.net/smartcontract/storagesc/block_reward.go
```diff
@@ -1,11 +1,178 @@
...
+	for i, b := range blobberPartition {
+		var sp *stakePool
+		var blobber partitions.BlobberRewardNode
+
+		err = blobber.Decode(b.Encode())
+		if err != nil {
+			return common.NewError("blobber_block_rewards_failed",
+				"can't decode blobber reward node: "+err.Error())
+		}
+		if sp, err = ssc.getStakePool(b.Name(), balances); err != nil {
```
Sriep: My benchmark of block rewards gave ~20ms. How much do we reduce this if we get stakepools first? Something like
```go
wg sync.WaitGroup
c := make(chan error)
for i, b := range blobberPartitions {
  wg.add(1)
  go func {
    defer wg. Done
    if sp, err = ssc.getStakePool(b.Name(), balances); err != nil { c <- err }
    mu.Lock()
    qsp = append(qsp, sp)
    mu.UnLock()
  } ()
}
wg.Wait()
  for err := range c {
        if err != nil {
            return err
        }
  }
```

ssardana96: it brought down a bit by few ms.
for me it came down from 24 -> 19. are you benchmarking on some linux bare metal server. I guess there it'll be a bit better timing.
I have pushed changes in this branch

Sriep (dismissed):

dabasov (approved):
lgtm

PR #968, opened 2022-02-03
fix #757 channels for eventdb

### LINKED ISSUE #757: Event db: Ensure events processed in correct order.

samtholiya:
  
...

### DIFF
--- a/code/go/0chain.net/smartcontract/dbs/event/process.go
+++ b/code/go/0chain.net/smartcontract/dbs/event/process.go
@@ -74,6 +74,10 @@ func (edb *EventDb) AddEvents(ctx context.Context, events []Event) {
 	}
 }
 
+func (edb *EventDb) addEventsWorker(ctx context.Context) {
+
+}
+
 func (edb *EventDb) addStat(event Event) error {
 	switch EventTag(event.Tag) {
 	// blobber
--- a/code/go/0chain.net/smartcontract/dbs/event/process.go
+++ b/code/go/0chain.net/smartcontract/dbs/event/process.go
@@ -49,33 +49,36 @@ const (
 )
 
 func (edb *EventDb) AddEvents(ctx context.Context, events []Event) {
-	newEvents := edb.removeDuplicate(ctx, events)
-
-	edb.addEvents(ctx, newEvents)
-	for _, event := range newEvents {
-		var err error = nil
-		switch EventType(event.Type) {
-		case TypeStats:
-			err = edb.addStat(event)
-		case TypeError:
-			err = edb.addError(Error{
-				TransactionID: event.TxHash,
-				Error:         event.Data,
-			})
-		default:
-		}
-		if err != nil {
-			logging.Logger.Error(
-				"event could not be processed",
-				zap.Any("event", event),
-				zap.Error(err),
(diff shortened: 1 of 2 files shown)


code/go/0chain.net/smartcontract/dbs/event/process.go
```diff
@@ -49,27 +49,34 @@ const (
 )
 
 func (edb *EventDb) AddEvents(ctx context.Context, events []Event) {
-	newEvents := edb.removeDuplicate(ctx, events)
+	edb.eventsChannel <- events
+}
+
+func (edb *EventDb) addEventsWorker(ctx context.Context) {
```
dabasov, line 55: I'd suggested using `WithContextFunc` from here https://github.com/0chain/0chain/blob/staging/code/go/0chain.net/core/common/context.go
to drain event pool

code/go/0chain.net/smartcontract/dbs/event/process.go
```diff
@@ -49,27 +49,34 @@ const (
...
+	for {
+		events := <-edb.eventsChannel
+		newEvents := edb.removeDuplicate(ctx, events)
```
kushthedude, line 58: Were you able to do a performance benchmark of this query? As of now we are traversing all the events to check for duplicates, also it since in earlier query too we had this removeDuplicate check but then again in the logs we saw a lot of errors for unique-constraint violation.

kushthedude (approved):
Looks good to me as a workaround by increasing the channel buffer, @samtholiya have you confirmed the order of the events with this change by some unit-tests or local sanity tests?

ssardana96 (commented):
lgtm.
Post benchmark and test we should be good.

ssardana96 (approved):

code/go/0chain.net/smartcontract/dbs/event/process.go
samtholiya: This optimization is not a part of this pr The query pr is https://github.com/0chain/0chain/pull/836

PR #1001, opened 2022-02-15
Fix #915 add miner/sharder get api for eventdb

### LINKED ISSUE #915: Existing miner sharder API : EventsDb
The Miner and Sharder tables can be used to provide data for the [/getMinerList](https://github.com/0chain/0chain/blob/staging/code/go/0chain.net/smartcontract/minersc/handler.go#L81) and [getSharderList](https://github.com/0chain/0chain/blob/staging/code/go/0chain.net/smartcontract/minersc/handler.go#L91) and [nodeStat](https://github.com/0chain/0chain/blob/staging/code/go/0chain.net/smartcontract/minersc/handler.go#L222) endpoints.

reference #790 
to be done post #857

samtholiya:
  
...


code/go/0chain.net/smartcontract/dbs/event/miner.go
```diff
@@ -52,6 +54,43 @@ func (edb *EventDb) GetMiner(id string) (*Miner, error) {
...
+	Rewards           null.Int
+	Fees              null.Int
+	Active            null.Bool
+	Longitude         null.Int
+	Latitude          null.Int
+}
+
+func (edb *EventDb) GetMinersWithFiltersAndPagination(filter MinerQuery, offset, limit int) ([]Miner, error) {
+	var miners []Miner
+	query := edb.Get().Debug().Model(&Miner{}).Where(&filter)
```
ssardana96: remove debug

code/go/0chain.net/smartcontract/dbs/event/miner.go
```diff
@@ -52,6 +54,43 @@ func (edb *EventDb) GetMiner(id string) (*Miner, error) {
...
+	if offset != -1 {
```
ssardana96: Change >=0

code/go/0chain.net/smartcontract/dbs/event/miner.go
```diff
@@ -52,6 +54,43 @@ func (edb *EventDb) GetMiner(id string) (*Miner, error) {
...
+		query = query.Offset(offset)
+	}
+	if limit != -1 {
```
ssardana96: change >=0

code/go/0chain.net/smartcontract/dbs/event/sharder.go
```diff
@@ -180,6 +181,43 @@ func (sh *Sharder) exists(edb *EventDb) (bool, error) {
...
+	ServiceCharge     null.Float
+	NumberOfDelegates null.Int
+	MinStake          null.Int
+	MaxStake          null.Int
+	LastHealthCheck   null.Int
+	Rewards           null.Int
+	Fees              null.Int
+	Active            null.Bool
+	Longitude         null.Int
+	Latitude          null.Int
```
ssardana96, line 198: have never used null
what benefit does it bring?

code/go/0chain.net/smartcontract/dbs/event/sharder.go
```diff
@@ -180,6 +181,43 @@ func (sh *Sharder) exists(edb *EventDb) (bool, error) {
...
+}
+
+func (edb *EventDb) GetShardersWithFilterAndPagination(filter SharderQuery, offset, limit int) ([]Sharder, error) {
+	var sharders []Sharder
+	query := edb.Get().Model(&Sharder{}).Where(&filter)
+	if offset != -1 {
```
ssardana96: change to >=0

code/go/0chain.net/smartcontract/dbs/event/sharder.go
```diff
@@ -180,6 +181,43 @@ func (sh *Sharder) exists(edb *EventDb) (bool, error) {
...
+		query = query.Offset(offset)
+	}
+	if limit != -1 {
```
ssardana96: change to >= 0

code/go/0chain.net/smartcontract/minersc/handler.go
```diff
@@ -81,11 +82,74 @@ func (msc *MinerSmartContract) GetNodepoolHandler(ctx context.Context, params ur
...
+		limitString  = params.Get("limit")
+		activeString = params.Get("active")
+		offset       = -1
+		limit        = -1
+		err          error
+	)
+	if offsetString != "" {
+		offset, err = strconv.Atoi(offsetString)
+		if err != nil {
+			return nil, err
```
ssardana96: please wrao it under common.NewError with function context

code/go/0chain.net/smartcontract/minersc/handler.go
```diff
@@ -81,11 +82,74 @@ func (msc *MinerSmartContract) GetNodepoolHandler(ctx context.Context, params ur
...
+		}
+	}
+	if limitString != "" {
+		limit, err = strconv.Atoi(limitString)
+		if err != nil {
+			return nil, err
```
ssardana96: common.NewError

code/go/0chain.net/smartcontract/minersc/handler.go
```diff
@@ -81,11 +82,74 @@ func (msc *MinerSmartContract) GetNodepoolHandler(ctx context.Context, params ur
...
+		}
+	}
+	filter := event.MinerQuery{}
+	if activeString != "" {
+		active, err := strconv.ParseBool(activeString)
+		if err != nil {
+			return nil, err
```
ssardana96: error wrapper

code/go/0chain.net/smartcontract/minersc/handler.go
```diff
@@ -81,11 +82,74 @@ func (msc *MinerSmartContract) GetNodepoolHandler(ctx context.Context, params ur
...
+		return nil, common.NewErrInternal("can't get miners. no db connection")
+	}
+
+	miners, err := balances.GetEventDB().GetMinersWithFiltersAndPagination(filter, offset, limit)
 	if err != nil {
 		return "", common.NewErrInternal("can't get miners list", err.Error())
 	}
-	return allMinersList, nil
+	minersArr := make([]interface{}, len(miners))
+	for i, miner := range miners {
```
ssardana96, line 123: please use existng struct MinerNodes
we should not duplicate structs

ssardana96: also we have function in minersc.minerTableToMinerNode() use transformers functions

code/go/0chain.net/smartcontract/minersc/handler.go
```diff
@@ -131,11 +195,74 @@ func (msc *MinerSmartContract) GetMinersStateHandler(ctx context.Context, params
...
+		limitString  = params.Get("limit")
+		activeString = params.Get("active")
+		offset       = -1
+		limit        = -1
+		err          error
+	)
+	if offsetString != "" {
+		offset, err = strconv.Atoi(offsetString)
+		if err != nil {
+			return nil, err
```
ssardana96: error wrapper

code/go/0chain.net/smartcontract/minersc/handler.go
```diff
@@ -131,11 +195,74 @@ func (msc *MinerSmartContract) GetMinersStateHandler(ctx context.Context, params
...
+		}
+	}
+	if limitString != "" {
+		limit, err = strconv.Atoi(limitString)
+		if err != nil {
+			return nil, err
```
ssardana96: error wrapper

code/go/0chain.net/smartcontract/minersc/handler.go
```diff
@@ -131,11 +195,74 @@ func (msc *MinerSmartContract) GetMinersStateHandler(ctx context.Context, params
...
+		}
+	}
+	filter := event.SharderQuery{}
+	if activeString != "" {
+		active, err := strconv.ParseBool(activeString)
+		if err != nil {
+			return nil, err
```
ssardana96: error wrapper

code/go/0chain.net/smartcontract/minersc/handler.go
```diff
@@ -131,11 +195,74 @@ func (msc *MinerSmartContract) GetMinersStateHandler(ctx context.Context, params
...
+
+	sharders, err := balances.GetEventDB().GetShardersWithFilterAndPagination(filter, offset, limit)
 	if err != nil {
-		return "", common.NewErrInternal(cantGetShardersListMsg, err.Error())
+		return "", common.NewErrInternal("can't get miners list", err.Error())
 	}
-	return allShardersList, nil
+	shardersArr := make([]interface{}, len(sharders))
+	for i, sharder := range sharders {
+		shardersArr[i] = map[string]interface{}{
```
ssardana96: please don't add new structs
use existing

code/go/0chain.net/smartcontract/minersc/miner.go
```diff
@@ -398,3 +398,27 @@ func validateNodeSettings(node *MinerNode, gn *GlobalNode, opcode string) error
 
 	return nil
 }
+
+type MinerView struct {
```
ssardana96: we should not create duplicate structs

code/go/0chain.net/smartcontract/minersc/sharder.go
```diff
@@ -349,3 +350,27 @@ func (msc *MinerSmartContract) sharderKeep(_ *transaction.Transaction,
 	buff := newSharder.Encode()
 	return string(buff), nil
 }
+
+type SharderView struct {
```
ssardana96: duplicate struct

ssardana96 (changes requested):

ssardana96:
@samtholiya I'm not seeing changes for [nodeStat](https://github.com/0chain/0chain/blob/staging/code/go/0chain.net/smartcontract/minersc/handler.go#L222) endpoint

code/go/0chain.net/smartcontract/dbs/event/sharder.go
kushthedude: `Int is an nullable int64. It does not consider zero values to be null. It will decode to null, not zero, if null.` 
Same for Bool too, by default bool is assigned false

ssardana96: @kushthedude still unclear
can you state with an example.

code/go/0chain.net/smartcontract/dbs/event/miner.go
samtholiya: keeping consistent default value with offset

ssardana96: Have a check >= 0 and not != -1
The checks should be logical and not restricted to a value.
Not sure why all comments starting with > are changed to quote
But everywhere add check >= 0

code/go/0chain.net/smartcontract/dbs/event/sharder.go
samtholiya: @ssardana08 while filtering we want values or we don't. 
So for example if we need Longitude = 0 and we do not use null.Int the form by default will not include `where Longitude=0` in query. So we use nullable value so the gorm is sure about whether the value is assigned intentionally or is just some default value. So if you now set `longitude=0` gorm will include the value in the query.

ssardana96: ok

code/go/0chain.net/smartcontract/minersc/handler.go
samtholiya: The reason we are not using the existing struct is that the existing struct creates an error in json unmarshalling because there are 2 `id` in json `ID` and  `id` one is by the gorm and other by our event struct.

code/go/0chain.net/smartcontract/minersc/sharder.go
samtholiya: needed for compatibility

code/go/0chain.net/smartcontract/minersc/handler.go
samtholiya: needed for compatibility (old json format)

code/go/0chain.net/smartcontract/minersc/handler.go
ssardana96: thats why I said transform
[minersc.minerTableToMinerNode()](https://github.com/0chain/0chain/blob/0c0e7894394e469e150e29786aeced3119107d70/code/go/0chain.net/smartcontract/storagesc/eventdb.go#L19)

code/go/0chain.net/smartcontract/minersc/handler.go
ssardana96: use existing transformer and existing structs
no duplications of struct

code/go/0chain.net/smartcontract/minersc/sharder.go
ssardana96: use existing transformer and existing structs
no duplications of struct
[minersc.minerTableToMinerNode()](https://github.com/0chain/0chain/blob/0c0e7894394e469e150e29786aeced3119107d70/code/go/0chain.net/smartcontract/storagesc/eventdb.go#L19)

code/go/0chain.net/smartcontract/dbs/event/miner.go
samtholiya: done

code/go/0chain.net/smartcontract/dbs/event/miner.go
samtholiya: done

code/go/0chain.net/smartcontract/dbs/event/miner.go
samtholiya: done

code/go/0chain.net/smartcontract/dbs/event/sharder.go
samtholiya: done

code/go/0chain.net/smartcontract/minersc/handler.go
samtholiya: done

code/go/0chain.net/smartcontract/minersc/handler.go
samtholiya: done

code/go/0chain.net/smartcontract/minersc/handler.go
samtholiya: done

code/go/0chain.net/smartcontract/minersc/miner.go
samtholiya: done

code/go/0chain.net/smartcontract/minersc/sharder.go
samtholiya: done

code/go/0chain.net/smartcontract/dbs/event/miner.go
```diff
@@ -30,26 +32,54 @@ type Miner struct {
...
 	Fees              state.Balance
 	Active            bool
-	Longitude         int64
-	Latitude          int64
+	Longitude         float64
+	Latitude          float64
 }
 
-func (edb *EventDb) GetMiner(id string) (*Miner, error) {
+func (edb *EventDb) GetMiner(id string) (Miner, error) {
```
ssardana96, line 39: any reason to not use reference?

code/go/0chain.net/smartcontract/minersc/handler.go
```diff
@@ -307,14 +386,20 @@ func (msc *MinerSmartContract) nodeStatHandler(ctx context.Context,
...
+	if balances.GetEventDB() == nil {
+		return nil, common.NewErrInternal("cannot connect to eventdb database")
+	}
+	if miner, err := balances.GetEventDB().GetMiner(id); err == nil {
+		return minerTableToMinerNode(miner), nil
+	}
+	if sharder, err := balances.GetEventDB().GetSharder(id); err == nil {
+		return sharderTableToSharderNode(sharder), nil
+	}
+	return nil, common.NewErrBadRequest("id was not found")
```
ssardana96: miner/sharder not found

ssardana96 (approved):

code/go/0chain.net/smartcontract/minersc/handler.go
ssardana96: message update

ssardana96 (changes requested):

ssardana96:
few comments
good to merge post integration testing

code/go/0chain.net/smartcontract/dbs/event/miner.go
samtholiya: returning a pointer sends the data to heap instead of the stack. 
There was no reason that we want to return a pointer here.

The less we send objects to stack the faster the program because less work for the garbage collector and in turn less garbage collector pauses in golang.

samtholiya: Note: garbage collector gets triggered after the heap allocation doubles or after a specific time (don't remember the interval time) whichever is earlier.

samtholiya:
Avoid the id in sharderlist and minerlist was just lazy to remove that 😅 
image
image
image
image

code/go/0chain.net/smartcontract/dbs/event/miner.go
ssardana96: @samtholiya we should not have a hard and fast rule here.
These structs are big structs and doesn't make sense to return copy every time.

We should be smart in when to return by value and pointer.
If returning pointer would be so bad, it would've not been given by nature of language to us.
the gc optimisation you'll be getting from this won't be too much, but creating copies of these structs every time is expensive. So we have to weigh the trade-offs

For big structs, I would not recommend to return value, but pointer.

samtholiya: okay... but this is not a big struct... And slices/maps that can be considered as bigger structs are by default reference driven.
Pointer return is allowed by golang for `mutex` or `struct containing mutex` because if you are returning a mutex it should be a pointer. 
or 
The pointer should be used in places where you want to achieve something similar to `flag` 

But otherwise using a pointer like the json.Unmarshal or marshal uses actually is better

ssardana96: @samtholiya it is a big struct, quite evidently:
```
type Miner struct {
	gorm.Model
	MinerID           string `gorm:"uniqueIndex"`
	N2NHost           string `gorm:"column:n2n_host"`
	Host              string
	Port              int
	Path              string
	PublicKey         string
	ShortName         string
	BuildTag          string
	TotalStaked       state.Balance
	Delete            bool
	DelegateWallet    string
	ServiceCharge     float64
	NumberOfDelegates int
	MinStake          state.Balance
	MaxStake          state.Balance
	LastHealthCheck   common.Timestamp
	Rewards           state.Balance
	Fees              state.Balance
	Active            bool
	Longitude         int64
	Latitude          int64
}
```
there is not even a single slice and map if you check carefully.
please don't have hard coded rules for return types, methods and arguments.
you should apply what best fits the code, this will not have much impact from the optimisation you're thinking.
we should not have hard and fast rules @samtholiya. please start thinking from what fits best in a given scenario.

it wouldn't really matter much if you use return as value or pointer much from gc POV
but copies of these big structs would be impactful. and moreover we have no point making copies..
Please go ahead with profiling and benchmark to show the improvement coming from your suggestions.

samtholiya: image

More processing is required because of the garbage collectors...

ssardana96: you're doing benchmarking wrongly to what we're discussing.
try doing benchmark for func calling these.
Take below a very simple example:
```
package benchmark

type X struct {
	a string
	b int
	c int64
	d float64
	e string
	f int
	g string
	h float64
}

func returnStruct() X {
	return X{
		a: "a",
		b: 0,
		c: 0,
		d: 0,
		e: "e",
		f: 0,
		g: "g",
		h: 0,
	}
}

func returnStructPointer() *X {
	return &X{
		a: "a",
		b: 0,
		c: 0,
		d: 0,
		e: "e",
		f: 0,
		g: "g",
		h: 0,
	}
}

func fetchStruct() {
	x := returnStruct()
	transformStruct(x)
}

func fetchStructPointer() {
	x := returnStructPointer()
	transformStructPointer(x)
}

func transformStruct(x X)         {}
func transformStructPointer(x *X) {}
```

I benchmarked `fetchStruct()` and `fetchStructPointer` and these are results:
```
~/GolandProjects/testProject/benchmark > go test -benchmem -run ^$ -bench .                                                                                                       4s
goos: darwin
goarch: arm64
pkg: testProject/benchmark
BenchmarkReturnStruct-8                 286885963                4.155 ns/op           0 B/op          0 allocs/op
BenchmarkReturnStructPointer-8          1000000000               0.3195 ns/op          0 B/op          0 allocs/op
PASS
ok      testProject/benchmark   2.707s
```
creating copies are expensive @samtholiya.
way more expensive than gc triggers that you're concerned about. and FYI gc trigger can be controlled by you also at any point using GOGC

samtholiya: No that is actually the wrong way because:
1. you did not add benchtime. The reason I added bechtime of 10m is so that I can get garbage collector metrics considered in the benchmark. That is the main concern. (heap allocation and garbage collection). Also the data your struct have is not equal to miner struct because more data on struct the faster the heap will be filled.
2. with your code you can not add benchtime because there is no io and therefore the benchmark will freeze if you added benchtime of 10 minutes. because the goroutine will not give up the process (at least happens on my dumb intel mac). 
3. Refer to this go conference that was conducted in Singapore: https://www.youtube.com/watch?v=ZMZpH4yT7M0 and Israel https://www.youtube.com/watch?v=gPxFOMuhnUU
4. Refer to this article if you don't have time for the above conference: https://medium.com/@vCabbage/go-are-pointers-a-performance-optimization-a95840d3ef85 and https://www.reddit.com/r/golang/comments/kfd5rr/when_should_i_return_a_pointer_vs_value_when/ 
(not the best explanation)
5. Just like in C++ you can write code that can lead to memory leakage even though it allows for that syntax. The language syntax is not the reason you code in any way to just compile. Golang != Java where you depend on garbage collector only and Golang != C++ where weak pointers exist.
6. In my code the entire logic was common the only change was the one we wanted to inspect. I understand not the best way to benchmark but it was the best way to benchmark considering the fact how golang behaves with goroutine.
so common algo requires common compute cycle
only uncommon algo requires different compute cycle 
so you can see the benchmark showing the diff

ssardana96: not sure why ur benchmark is hanging, i was able to run
but again not sure what u r benchmarking, but I'm discussing the benchmarking of return types and the impact it brings and that's what I have written.

Please take my code, add as much data to struct you want.
that is the bare minimum we want to test, right?

I hope that'll answer the return type(ref vs value) changes that is concerning?

ssardana96: All apart!
anyhow I'm not able to run your benchmark as some db connection issue, but I don't see a drastic difference from what your both benchmarks are resulting. and the benchmark written is completely wrong. For you to do deeper dive, try profiling allocation and cpu and see where ur benchmarking is making few ns difference. but clearly bad benchmarking @samtholiya 

Question we're discussing is clearly shown in my benchmarks, of return type pointer vs value.
Please let me know if it doesn't answer the basic question on where performance is getting impacted.

ssardana96: having external connection factor the way you have is not the way to benchmark these cases.
you can't say for sure, the time impact is due to which reasons.

to have a concrete benchmarking on what we're discussing a bare minimum setup with simple functionality which I have used should be used.
I'm not getting why we don't stick to the basics and keep trying things beyond that first to GC, etc. Our basics are the ones that'll make an impact hugely.
my benchmarks are clearly stating 14X difference between both. Definitely you should've a reasoning to deny that factor @samtholiya

ssardana96: I'm doing benchmarking with different struct size. Something interesting to see here. Posting results for better clarity:

struct size 10:
```
BenchmarkStructReturnValue
BenchmarkStructReturnValue-8     	149987712	         7.982 ns/op	       0 B/op	       0 allocs/op
BenchmarkStructReturnPointer
BenchmarkStructReturnPointer-8   	268468488	         4.469 ns/op	       0 B/op	       0 allocs/op
```

struct size 100
```
BenchmarkStructReturnValue
BenchmarkStructReturnValue-8     	24166323	        49.78 ns/op	       0 B/op	       0 allocs/op
BenchmarkStructReturnPointer
BenchmarkStructReturnPointer-8   	35991136	        33.19 ns/op	       0 B/op	       0 allocs/op
```

struct size 1000
```
BenchmarkStructReturnValue
BenchmarkStructReturnValue-8     	 2371482	       503.5 ns/op	       0 B/op	       0 allocs/op
BenchmarkStructReturnPointer
BenchmarkStructReturnPointer-8   	 3650842	       328.8 ns/op	       0 B/op	       0 allocs/op
PASS
```

ssardana96: To what u r saying, there are functions where returning by value would make sense too. Yes.
But all these theories are for some scenarios, which you've taken as a rule by heart.

And what you're saying, are not yet proven by any of ur benchmarks 😅.
I've shared with you an explaination - a very simple and given 5 benchmarks to show a very simple, not so complicated, comparison.

Let's take this further if any concrete evidence to your points by benchmarking or profiling.
Else please revert to references :)

samtholiya: If you watch the video they provide live examples...

samtholiya: that is valid proof and a golang thing
and a software engineer fundamental "HEAP is not good"
Give me one reference where they say heap is better than stack allocation in long term for performance.
I think the entire golang community is talking from experience and not something as a theory.

I think you need to check the references.
They are from golang community people who work on golang 

Also if you leave golang heap is never a good thing
I hope you read college books as well. And there are many examples

Your program works if I run benchmark directly but that does not account for garbage collection. For that to happen. you have to run a benchmark for enough time so that the heap doubles.

The world is not insane to say that.

samtholiya: Onefootball is a company whose entire infrastructure is based on golang. The entire codebase never returns a pointer
Also if you see all the functions in golang they have avoided things going to heap.
Even gorm does the same. 
So I don't know whom you are following.|
Knowing syntax != knowing language

ssardana96: That's great @samtholiya 
Let me go through the above references.

@samtholiya meanwhile can u try having a benchmark written without external db connections and gc triggers as u want. That would help understand our discussion of return types (pointer vs value) for big struct.
there are so many things and optimisations compiler do under the hood.
It's great to have this discussion. Let's use go tools with so much capabilities today to discuss out our point.

Our point of discussion: "for big structs to be returned - is it better to have return pointer or value?"
let's right benchmark to get this out.
Also let's benchmark if the result changes if we do the same for int or some small data types?

It's time to use all reference knowledge to practical. Use benchmarks and profiling to state the facts (if any) 

Thanks

ssardana96: An interesting point to note in ur benchmark screenshot which I'm not able to see difference is.
The benchmark benchmem just captures heap allocation. If we are comparing the heap on both benchmark, then ur 1 benchmark should have more heap allocation than other. But I see equal in both cases.

I need 1 right benchmarking to state your point:
- db connections should not be part of the benchmark we're discussing
- I'm expecting to see difference in heap allocation of both benchmark
- would be great to also see gc triggers comparison for both scenarios

samtholiya: Had to use log otherwise the benchmark gets hanged.(The way goroutine and benchmarks works)
The benchmarks you did ... did not include garbage collection. The way I included them is by setting the benchtime which forces the benchmark to run for a specified time. And because garbage collector runs as soon as the heap doubles you can see the bellow output
Hope you don't mind the log ... because if that is an issue I don't know how to convince you.
image

samtholiya: Also benchmark did not hang for you because you did not use benchtime

samtholiya: And about the diff being small... It increases with time and the more such code which returns pointer fills the heap.
If you want more diff try running it with 10m instead of 2m which I did

ssardana96: @samtholiya I was able to run even with benchtime, not sure why benchtime will block something.

That being said, don't try benchmark of miner. We don't need db and all to benchmark.

What you need to do to bechmark is:
- benchmark taking just general functions. Big structs and try returning pointer and value.
- please note important aspect here is big struct.
- make sure you don't have any db or external connections in functions. That is not needed for our discussion.

What I'm expecting is:
- time difference.
- allocation differences
Please feel free to use benchtime and other args that makes it helpful.
Also what would be great is to profile gc triggers for both the return types.
Also using go tools you can find out what goes to heap and stack. I guess -m. You can try that also.

All the above points will help us answer what is needed.

Let me know if any doubts.

samtholiya: if you see i have changed the code for the benchmark. It does not include DB anymore.

samtholiya: also, the reason for bench time is more of a goroutine thing. it is because of the way goroutines are scheduled

ssardana96: cool! this works @samtholiya 
thanks for the discussion.

ssardana96:
good to be merged, post unit test fix

ssardana96 (approved):

service-0chain:
| config  | value |
| tests skipped  | FALSE  |
| system-tests  | master  |
| gosdk  | NOT OVERRIDDEN  |
| zboxcli  | staging  |
| zwalletcli  | staging  |
| 0chain  | feat/minerSharderList  |
| blobber  | staging  |
| authorizer  | staging  |
| 0box  | staging  |
| 0block  | staging  |
| 0dns  | staging  |
| explorer  | staging  |
| 0proxy  | staging  |
| 0search  | staging  |

Issue #769, opened 2021-12-10
Event db: Challenge table.


Sriep:
We need a new Challenge table for the Event Db.

Code for this already exists in https://github.com/0chain/0chain/pull/679 [challenge.go](https://github.com/0chain/0chain/blob/challenge-events/code/go/0chain.net/smartcontract/dbs/event/challenge.go#L10).

This needs to be reworked as part of https://github.com/0chain/0chain/issues/741. The event database code should be altered to use the blobber table rather than the blobberChallenge table.

ma2b0043:
@Sriep when can we expect to close this issue?

PR #679, opened 2021-11-18
WIP: Challenge events

Sriep:
Emits events when a new challenge is created or resolved.
Adds a blobber challenge table to the events database, emitted events used to update the table.
Modify the 0chain storagesc endpoints `/openchallenges` and `/getchallenge` to get its information form the event SQL database rather than the MPT.

Some of the fields that are not being removed.
Validators - `stake_pool_settings`
Blobber - `geolocation`, `terms`, `capacity`, `used`, `last_health_check`, and `stake_pool_settings`.
`lastest_completed_challenge`

PR #1055, opened 2022-03-10
Feat/generate challenge

### LINKED ISSUE #741: Modify challenge creation algorithm

ssardana96:
## Tests 

- blobber:
- gosdk:
- system_test: https://github.com/0chain/system_test/pull/224
- zboxcli:
- zwalletcli:
- Other: ...


code/go/0chain.net/smartcontract/storagesc/alloation2_test.go
```diff
@@ -363,17 +363,30 @@ func testCancelAllocation(
...
-			Challenges: []*StorageChallenge{},
+			BlobberID: strconv.Itoa(i),
+			//Challenges: []*StorageChallenge{},
+		}
+		//for _, created := range blobberChallenges {
+		//	bc.Challenges = append(bc.Challenges, &StorageChallenge{
+		//		AllocationID: sAllocation.ID,
+		//		Created:      created,
+		//	})
+		//}
```
dabasov: can we delete this dead code?

code/go/0chain.net/smartcontract/storagesc/challenge.go
```diff
@@ -581,27 +689,21 @@ func (sc *StorageSmartContract) generateChallenges(t *transaction.Transaction,
 			"error getting the validators list: %v", err)
 	}
 
-	var all *Allocations
```
Sriep, line 723: At statt of function.
```go
	var tp = time.Now()
	defer sc.addGenerateChallengesStat(tp, &err)
```
Could we move this out of the smart-contract. It should not have access to non-deterministic data.

code/go/0chain.net/smartcontract/storagesc/challenge.go
```diff
@@ -581,27 +689,21 @@ func (sc *StorageSmartContract) generateChallenges(t *transaction.Transaction,
...
 	}
 
-	if len(all.List) == 0 {
-		return common.NewError("adding_challenge_error",
-			"no allocations at this time")
-	}
-
-	var selectAlloc = func(i int) (alloc *StorageAllocation, err error) {
-		alloc, err = sc.getAllocation(all.List[i], balances)
+	var selectAlloc = func(allocID string) (alloc *StorageAllocation, err error) {
```
Sriep: Why use an Anonymous function to validate an allocation, rather than a regular one?

code/go/0chain.net/smartcontract/partitions/blobber_challenge_list.go
```diff
@@ -0,0 +1,172 @@
...
+	"0chain.net/core/util"
+	"encoding/json"
+	"errors"
+	"fmt"
+)
+
+//------------------------------------------------------------------------------
+
+type BlobberChallengeNode struct {
+	ID           string `json:"id"`
```
Sriep: Maybe `BlobberID` rather than `ID`. This is not the id of the `BlobberChallengeNode` but the id of some blobber somewhere else.

code/go/0chain.net/smartcontract/storagesc/challenge.go
```diff
@@ -617,17 +719,35 @@ func (sc *StorageSmartContract) generateChallenges(t *transaction.Transaction,
...
-		alloc, err = selectAlloc(r.Intn(len(all.List)))
+		bcPartition, err := blobberChallengeList.GetRandomSlice(cr, balances)
+		if err != nil {
+			return common.NewError("generate_challenges",
+				"error getting random slice from blobber challenge partition")
+		}
+
+		randomIndex := rand.Intn(len(bcPartition))
+		bcItem := bcPartition[randomIndex]
+		val, ok := bcItem.(*partitions.BlobberChallengeNode)
```
Sriep: Ugh! could we hide this bit of code in the Partitions package. Maybe add a global  `GetRandomAllocationBlobberPair` `functions to blobber_callenge_list.go` for the previous twenty or so lines of code.

code/go/0chain.net/smartcontract/storagesc/challenge.go
```diff
@@ -617,17 +719,35 @@ func (sc *StorageSmartContract) generateChallenges(t *transaction.Transaction,
...
 
-	//
-	//
-	//
-
 	var alloc *StorageAllocation
 
 	for i := int64(0); i < numChallenges; i++ {
 
-		// looking for allocation with NumWrites > 0
```
Sriep, line 767: Could we split this `for` loop into two `for` loops? 
1. Call all the `GetTrieNode` methods. Make this a parallel `for` loop using wait groups or such.
2. Use a synchronous `for` loop for all the `InsertTrieNode` calls.

We are getting a few hundred items from the MPT, this could save a significant amount of time.

Probably best to do most of the work in the first for loop. Create a slice of `allocChallengeObj`, `storageChallenge`, `blobberChallengeObj` and `alloc` objects to be inserted into the MPT in the second for loop.

code/go/0chain.net/smartcontract/storagesc/challenge.go
```diff
@@ -667,48 +777,43 @@ func (sc *StorageSmartContract) generateChallenges(t *transaction.Transaction,
...
-		blobberAllocation = alloc.BlobberMap[selectedBlobberObj.ID]
-		if blobberAllocation.AllocationRoot != "" {
-			break // found
+	for _, b := range alloc.Blobbers {
+		if b.ID == blobberID {
+			blobber = b
+			break
 		}
 	}
+	if blobber == nil {
```
Sriep: blobber is never nil as set to `&StorageNode{}` above. Check `blobber.ID == ""` or similar instead.

code/go/0chain.net/smartcontract/storagesc/challenge.go
```diff
@@ -722,39 +827,77 @@ func (sc *StorageSmartContract) addChallenge(alloc *StorageAllocation,
...
-	addedChallege := blobberChallengeObj.addChallenge(&storageChallenge)
-	if !addedChallege {
+	addedAllocChallenge := allocChallengeObj.addChallenge(&storageChallenge)
+	if !addedAllocChallenge {
 		challengeBytes, err := json.Marshal(storageChallenge)
 		return string(challengeBytes), err
 	}
 
-	balances.InsertTrieNode(blobberChallengeObj.GetKey(sc.ID), blobberChallengeObj)
+	_, err = balances.InsertTrieNode(allocChallengeObj.GetKey(sc.ID), allocChallengeObj)
```
Sriep, line 1094: Separate out the following block of code with the  `InsertTrieNode` calls, and put in a separate synchronous second for loop.

code/go/0chain.net/smartcontract/storagesc/models.go
```diff
@@ -117,6 +121,61 @@ func (sn *BlobberChallenge) GetHashBytes() []byte {
...
+		sn.ChallengeIDs = append(sn.ChallengeIDs, challenge.ID)
+		sn.ChallengeIDMap[challenge.ID] = true
+		return true
+	}
+	return false
+}
+
+type AllocationChallenge struct {
+	AllocationID             string                       `json:"allocation_id"`
+	Challenges               []*StorageChallenge          `json:"challenges"`
```
Sriep, line 178: This slice of `StorageChallenge` objects is too big. I think we need to come up with a way of avoiding needing it. In the meanwhile, the StoreaChallenger object only needs to know its Id and Created.

code/go/0chain.net/smartcontract/storagesc/models.go
```diff
@@ -144,19 +204,47 @@ func (sn *BlobberChallenge) addChallenge(challenge *StorageChallenge) bool {
...
 
 type StorageChallenge struct {
-	Created        common.Timestamp   `json:"created"`
-	ID             string             `json:"id"`
-	PrevID         string             `json:"prev_id"`
-	Validators     []*ValidationNode  `json:"validators"`
-	RandomNumber   int64              `json:"seed"`
-	AllocationID   string             `json:"allocation_id"`
-	Blobber        *StorageNode       `json:"blobber"`
+	Created common.Timestamp `json:"created"` // needed
```
Sriep: Why do we need anything other than ID and Created here *(for `canceledPassRates`)? 

This needs to be as slim as possible or ideally removed altogether. Most of this data is to be stored in the event database not the MPT. Only keep what is absolutely essential in the MPT.

code/go/0chain.net/smartcontract/storagesc/challenge.go
```diff
@@ -722,39 +827,77 @@ func (sc *StorageSmartContract) addChallenge(alloc *StorageAllocation,
...
+	if err != nil {
+		return "", common.NewError("add_challenge",
+			"error storing alloc challenge: "+err.Error())
+	}
+
+	_, err = balances.InsertTrieNode(storageChallenge.GetKey(sc.ID), &storageChallenge)
```
Sriep: The storageChallenge object should be emitted to the event database and stored in the Challenge table, not stored in the MPT. For example see https://github.com/0chain/0chain/pull/679 
[balances.EmitEvent(event.TypeStats, event.TagAddChallenge, challengeID, string(data))](https://github.com/0chain/0chain/blob/challenge-events/code/go/0chain.net/smartcontract/storagesc/challenge.go#L809)

This way we can remove everything from StorageChallenge except what we actually need for smart contracts. Ideally, it can be replaced altogether with summary stats.

code/go/0chain.net/smartcontract/storagesc/challenge.go
```diff
@@ -722,39 +827,77 @@ func (sc *StorageSmartContract) addChallenge(alloc *StorageAllocation,
...
-	balances.InsertTrieNode(alloc.GetKey(sc.ID), alloc)
+
+	_, err = balances.InsertTrieNode(alloc.GetKey(sc.ID), alloc)
+	if err != nil {
+		return "", common.NewError("add_challenge",
+			"error storing allocation: "+err.Error())
+	}
 	//Logger.Info("Adding a new challenge", zap.Any("blobberChallengeObj", blobberChallengeObj), zap.Any("challenge", storageChallenge.ID))
 	challengeBytes, err := json.Marshal(storageChallenge)
 	sc.newChallenge(balances, storageChallenge.Created)
```
Sriep: `sc.newChallenge(balances, storageChallenge.Created)` I am not convinced we need this object, but it does not need to be updated once per loop, we can do this once at the end outside the for loops.

Sriep (changes requested):
We need a new Challenge table https://github.com/0chain/0chain/issues/769 for the event database if this PR is to work  For example https://github.com/0chain/0chain/pull/679. This can either be done first or included as part of this PR?

code/go/0chain.net/smartcontract/storagesc/blobber.go
```diff
@@ -1,19 +1,20 @@
 package storagesc
 
 import (
+	"0chain.net/core/logging"
+	"0chain.net/smartcontract/partitions"
```
peterlimg: run goimports to format

code/go/0chain.net/smartcontract/partitions/blobber_challenge.go
```diff
@@ -0,0 +1,166 @@
...
+	if err := il.Decode(val.Encode()); err != nil {
+		return fmt.Errorf("%w: %s", common.ErrDecoding, err)
+	}
+	il.Key = key
+	return nil
+}
+
+func (il *blobberChallengeItemList) add(it PartitionItem) error {
+
+	il.Items = append(il.Items, BlobberChallengeNode{
```
peterlimg, line 85: Check duplicate and return error like bbr

code/go/0chain.net/smartcontract/storagesc/challenge.go
```diff
@@ -28,23 +28,87 @@ import (
...
+	if len(blobberChallengeObj.ChallengeIDs) > 0 {
+		latestOpenChallengeID := blobberChallengeObj.ChallengeIDs[0]
+		if latestOpenChallengeID == challengeCompleted.ID {
 			found = true
 		}
 	}
 	idx := 0
-	if found && idx >= 0 && idx < len(blobberChallengeObj.Challenges) {
-		blobberChallengeObj.Challenges = append(blobberChallengeObj.Challenges[:idx], blobberChallengeObj.Challenges[idx+1:]...)
+	if found && idx < len(blobberChallengeObj.ChallengeIDs) {
```
peterlimg, line 127: If found is true, the idx is always 0 I think, so perhaps we can use ChallengeIDs = ChallengeIDs[1:] to remove it and make this code cleaner

code/go/0chain.net/smartcontract/storagesc/challenge.go
```diff
@@ -513,6 +640,104 @@ func (sc *StorageSmartContract) addGenerateChallengesStat(tp time.Time,
...
+	alloc, err = sc.getAllocation(allocID, balances)
+	if err != nil && err != util.ErrValueNotPresent {
+		return nil, common.NewErrorf("adding_challenge_error",
+			"unexpected error getting allocation: %v", err)
+	}
+	if err == util.ErrValueNotPresent {
+		Logger.Error("client state has invalid allocations",
+			zap.Any("selected_allocation", allocID))
+		return nil, common.NewErrorf("invalid_allocation",
+			"client state has invalid allocations")
```
peterlimg, line 731: Use switch case to make this error handling easier to read, e.g:
switch err {
case nil:
case util.ErrValueNotPresent:
default:

code/go/0chain.net/smartcontract/storagesc/challenge.go
```diff
@@ -513,6 +640,104 @@ func (sc *StorageSmartContract) addGenerateChallengesStat(tp time.Time,
 	}
 }
 
+func (sc *StorageSmartContract) selectAllocationForChallenge(
```
peterlimg: function name as getAllocationForChallenge would be more reasonable, as we don't have any 'select' action here, we just get it from mpt and validate it.

code/go/0chain.net/smartcontract/storagesc/challenge.go
```diff
@@ -513,6 +640,104 @@ func (sc *StorageSmartContract) addGenerateChallengesStat(tp time.Time,
...
+	defer wg.Done()
+
+	for d := range data {
+		bcPartition, err := blobberChallengeList.GetRandomSlice(d.cr, balances)
+		if err != nil {
+			errChan <- common.NewError("generate_challenges",
+				"error getting random slice from blobber challenge partition")
+		}
+
+		randomIndex := rand.Intn(len(bcPartition))
```
peterlimg: The randomIndex will make the state inconsistent when the block is execute on different miners/sharders, and lead  to state hash mismatch error.  We need to run it with the same random seed.

code/go/0chain.net/smartcontract/storagesc/challenge.go
```diff
@@ -513,6 +640,104 @@ func (sc *StorageSmartContract) addGenerateChallengesStat(tp time.Time,
...
+				"error getting blobber_challenge_allocation list: "+err.Error())
+		}
+
+		// maybe we should use another random seed
+		bcAllocPartition, err := bcAllocList.GetRandomSlice(d.cr, balances)
+		if err != nil {
+			errChan <- common.NewError("generate_challenges",
+				"error getting random slice from blobber challenge allocation partition")
+		}
+		randomIndex = rand.Intn(len(bcAllocPartition))
```
peterlimg: same issue as above

code/go/0chain.net/smartcontract/storagesc/challenge.go
```diff
@@ -28,23 +28,87 @@ import (
...
+			nil,
+			partitions.ItemBlobberChallenge,
+		)
+	}
+	all.SetCallback(nil)
+	return all, nil
+}
+
+func getBlobbersChallengeAllocationList(blobberID string, balances c_state.StateContextI) (partitions.RandPartition, error) {
+	all, err := partitions.GetRandomSelector(getBlobberChallengeAllocationKey(blobberID), balances)
```
peterlimg, line 99: When will this partition be updated?

peterlimg (changes requested):

code/go/0chain.net/smartcontract/partitions/itemlist.go
```diff
@@ -50,9 +51,10 @@ func (il *itemList) get(key datastore.Key, balances state.StateContextI) error {
 	return nil
 }
 
-func (il *itemList) add(it PartitionItem) {
+func (il *itemList) add(it PartitionItem) error {
 	il.Items = append(il.Items, StringItem{it.Name()})
```
peterlimg: Check duplicate item

code/go/0chain.net/smartcontract/storagesc/models.go
```diff
@@ -93,10 +99,12 @@ type ChallengeResponse struct {
 }
 
 type BlobberChallenge struct {
-	BlobberID                string                       `json:"blobber_id"`
-	Challenges               []*StorageChallenge          `json:"challenges"`
-	ChallengeMap             map[string]*StorageChallenge `json:"-"`
-	LatestCompletedChallenge *StorageChallenge            `json:"lastest_completed_challenge"`
+	BlobberID string `json:"blobber_id"`
+	//Challenges               []*StorageChallenge          `json:"challenges"`                  //to be removed
+	//ChallengeMap             map[string]*StorageChallenge `json:"-"`                           // to be removed
```
peterlimg: Remove if these commented code is not used anymore

docker.local/bin/unit_test.sh
```diff
@@ -43,5 +43,5 @@ if [[ -n "$PACKAGE" ]]; then
 else
     # Run all tests.
     echo "Running general unit tests:"
-    docker run "$INTERACTIVE" zchain_unit_test sh -c "cd 0chain.net; go test -tags bn256 -cover ./..."
+    docker run "$INTERACTIVE" zchain_unit_test sh -c "cd 0chain.neti/smartcontract/storagesc; go test -tags bn256 -cover ./..."
```
peterlimg: Seems like local changes for testing

peterlimg (changes requested):

code/go/0chain.net/smartcontract/storagesc/alloation2_test.go
```diff
@@ -363,17 +363,30 @@ func testCancelAllocation(
...
 				Created:      created,
 			})
 		}
 		_, err := ctx.InsertTrieNode(bc.GetKey(ssc.ID), &bc)
 		require.NoError(t, err)
+		_, err = ctx.InsertTrieNode(ac.GetKey(ssc.ID), &ac)
+		require.NoError(t, err)
 	}
 
 	allAllocationsBefore, err := ssc.getAllAllocationsList(ctx)
```
Sriep: The only user for the `ALL_ALLOCATIONS_KEY` and the all allcoatinos list is in genearteChallengers. As we are not going to use it anymore all references to `ALL_ALLOCATIONS_KEY` can be removed. 

This will save a lot of time across several endpoints.

Sriep (changes requested):

code/go/0chain.net/smartcontract/storagesc/challenge.go
ssardana96: done

code/go/0chain.net/smartcontract/storagesc/blobber.go
ssardana96: running goimport formatted imports
but how did you get to know we should run it?

code/go/0chain.net/smartcontract/storagesc/challenge.go
ssardana96: peter can you check this now.

code/go/0chain.net/smartcontract/storagesc/challenge.go
ssardana96: can u check this now

code/go/0chain.net/smartcontract/storagesc/challenge.go
ssardana96: in commitBlobberConnection

service-0chain:
| config  | value |
| tests skipped  | FALSE  |
| system-tests  | feat/generate_challenge  |
| gosdk  | NOT OVERRIDDEN  |
| zboxcli  | staging  |
| zwalletcli  | staging  |
| 0chain  | feat/generate_challenge  |
| blobber  | staging  |
| authorizer  | staging  |
| 0box  | staging  |
| 0block  | staging  |
| 0dns  | staging  |
| explorer  | staging  |
| 0proxy  | staging  |
| 0search  | staging  |

code/go/0chain.net/smartcontract/storagesc/challenge.go
```diff
@@ -498,43 +591,290 @@ func (sc *StorageSmartContract) addGenerateChallengesStat(tp time.Time,
...
+			"error getting the blobber challenge list: %v", err)
+	}
+
+	var (
+		data   = make(chan challengeInput)
+		output = make(chan challengeOutput)
+	)
+	defer close(data)
+
+	go sc.asyncGenerateChallenges(blobberChallengeList, validators, r, t.CreationDate, data, output, balances)
```
Sriep: Why are we running this  in a goroutine? We are only generating one challenge so there is no need for this.

code/go/0chain.net/smartcontract/storagesc/challenge.go
```diff
@@ -498,43 +591,290 @@ func (sc *StorageSmartContract) addGenerateChallengesStat(tp time.Time,
...
+	validators partitions.RandPartition,
+	r *rand.Rand,
+	creationDate common.Timestamp,
+	data <-chan challengeInput,
+	result chan<- challengeOutput,
+	balances c_state.StateContextI) {
+
+	defer close(result)
+
+	for d := range data {
```
Sriep: When only creating one challenge we want to create it synchronously. In which case, there will be only one entry in data.

I suggest putting the content of this for loop in a separate function that can be called directly from generateChallenge instead of `asyncGenerateChallenges` and also from for loop.

code/go/0chain.net/smartcontract/storagesc/challenge.go
```diff
@@ -498,43 +591,290 @@ func (sc *StorageSmartContract) addGenerateChallengesStat(tp time.Time,
...
+
+	blobberChallengeList, err := getBlobbersChallengeList(balances)
+	if err != nil {
+		return common.NewErrorf("adding_challenge_error",
+			"error getting the blobber challenge list: %v", err)
+	}
+
+	var (
+		data   = make(chan challengeInput)
+		output = make(chan challengeOutput)
```
Sriep: We are only creating one challenge so there is no need to use channels.

code/go/0chain.net/smartcontract/storagesc/challenge.go
```diff
@@ -498,43 +591,290 @@ func (sc *StorageSmartContract) addGenerateChallengesStat(tp time.Time,
...
 
-	lastChallengeTime := stats.LastChallengedTime
-	if lastChallengeTime == 0 {
-		lastChallengeTime = t.CreationDate
+	cr := rand.New(rand.NewSource(int64(challengeSeed)))
+
+	data <- challengeInput{
+		cr:          cr,
+		t:           t,
+		challengeID: challengeID,
```
Sriep: All this code involving channels can be removed.

Sriep (changes requested):

code/go/0chain.net/smartcontract/storagesc/challenge.go
ssardana96: this has channels. hence goroutine

code/go/0chain.net/smartcontract/storagesc/benchmark_tests.go
```diff
@@ -748,7 +748,7 @@ func BenchmarkTests(
 			) (string, error) {
 				challengesEnabled := viper.GetBool(bk.StorageChallengeEnabled)
 				if challengesEnabled {
-					err := ssc.generateChallenges(txn, balances.GetBlock(), nil, balances)
+					err := ssc.generateChallenge(txn, balances.GetBlock(), nil, balances)
```
Sriep, line 791: Change the name as well, ~s~ 
```go
			name: "storage.generate_challenge",
```

service-0chain:
| config  | value |
| tests skipped  | FALSE  |
| system-tests  | feat/generate_challenge  |
| gosdk  | NOT OVERRIDDEN  |
| zboxcli  | staging  |
| zwalletcli  | staging  |
| 0chain  | feat/generate_challenge  |
| blobber  | staging  |
| authorizer  | staging  |
| 0box  | staging  |
| 0block  | staging  |
| 0dns  | staging  |
| explorer  | staging  |
| 0proxy  | staging  |
| 0search  | staging  |

code/go/0chain.net/smartcontract/storagesc/sc.go
```diff
@@ -217,29 +218,29 @@ func (sc *StorageSmartContract) Execute(t *transaction.Transaction,
...
 		}
-		challengesEnabled := config.SmartContractConfig.GetBool(
-			"smart_contracts.storagesc.challenge_enabled")
-		if challengesEnabled {
-			err = sc.generateChallenges(t, balances.GetBlock(), input, balances)
-			if err != nil {
-				return "", err
-			}
-		}
+		//challengesEnabled := config.SmartContractConfig.GetBool(
```
Sriep: remove commented out code, same below.

code/go/0chain.net/smartcontract/storagesc/challenge.go
```diff
@@ -489,60 +584,275 @@ func (sc *StorageSmartContract) verifyChallenge(t *transaction.Transaction,
...
+		result.blobberChallenge,
+		result.allocChallenge,
+		result.blobberAlloc,
+		balances,
+	)
+	if err != nil {
+		return common.NewErrorf("adding_challenge_error",
+			"Error in adding challenge: %v", err)
 	}
+	if tm := sc.SmartContractExecutionStats["challenge_request"]; tm != nil {
```
Sriep: What is this bit of code doing? I can't see any reason to keep it.

code/go/0chain.net/smartcontract/storagesc/blobber.go
peterlimg: The packages sequence is different if not formatted.

code/go/0chain.net/smartcontract/storagesc/challenge.go
```diff
@@ -451,7 +541,7 @@ func (sc *StorageSmartContract) verifyChallenge(t *transaction.Transaction,
 	}
 
 	var (
-		threshold = len(challReq.Validators) / 2
+		threshold = challReq.TotalValidators / 2
 		pass      = success > threshold ||
 			(success > failure && success+failure < threshold)
```
peterlimg, line 546: Mark for later discussion about the `pass` calculation

peterlimg, line 546: Mark for later  discussion about the `pass` calculation and the following partial success reward

code/go/0chain.net/smartcontract/storagesc/challenge.go
```diff
@@ -628,85 +716,278 @@ func (sc *StorageSmartContract) verifyChallenge(t *transaction.Transaction,
...
+			"client state has invalid allocations")
+	default:
+		return nil, common.NewErrorf("adding_challenge_error",
+			"unexpected error getting allocation: %v", err)
 	}
 
-	var tm = sc.SmartContractExecutionStats["generate_challenges"]
-	if tm == nil {
-		return // missing timer (unexpected)
+	if alloc.Expiration < t.CreationDate {
```
peterlimg, line 737: we can return error to indicate the allocation expiration error

code/go/0chain.net/smartcontract/storagesc/challenge.go
```diff
@@ -628,85 +716,278 @@ func (sc *StorageSmartContract) verifyChallenge(t *transaction.Transaction,
...
+		return nil, nil
 	}
-
-	if timer, ok := tm.(metrics.Timer); ok {
-		timer.Update(time.Since(tp))
+	if alloc.Stats == nil {
+		return nil, nil
```
peterlimg: Return error to indicate the empty stats

code/go/0chain.net/smartcontract/storagesc/challenge.go
```diff
@@ -628,85 +716,278 @@ func (sc *StorageSmartContract) verifyChallenge(t *transaction.Transaction,
...
+
+	for _, b := range alloc.Blobbers {
+		if b.ID == blobberID {
+			blobber = b
+			break
+		}
+	}
+
+	blobberAllocation, ok := alloc.BlobberMap[blobber.ID]
+	if !ok {
```
peterlimg, line 817: Move this checking above before the blobber finding to avoid unnecessary for loop.

code/go/0chain.net/smartcontract/storagesc/challenge.go
```diff
@@ -628,85 +716,278 @@ func (sc *StorageSmartContract) verifyChallenge(t *transaction.Transaction,
...
+		return nil, common.NewError("add_challenges",
+			"blobber allocation doesn't exists in allocation")
+	}
+
+	if blobberAllocation.Stats == nil {
+		blobberAllocation.Stats = new(StorageAllocationStats)
+	}
+
+	selectedValidators := make([]*ValidationNode, 0)
+	randSlice, err := validators.GetRandomSlice(challRand, balances)
```
peterlimg: rename randSlice to randValidators would make it meaningful and easy to read

code/go/0chain.net/smartcontract/storagesc/challenge.go
```diff
@@ -628,85 +716,278 @@ func (sc *StorageSmartContract) verifyChallenge(t *transaction.Transaction,
...
+	if err != nil {
+		return nil, common.NewError("add_challenge",
+			"error getting validators random slice: "+err.Error())
+	}
+
+	perm := challRand.Perm(len(randSlice))
+	for i := 0; i < minInt(len(randSlice), alloc.DataShards+1); i++ {
+		if randSlice[perm[i]].Name() != blobber.ID {
```
peterlimg: Assign as `randValidator := randSlice[perm[i]]` and use it below would make the code cleaner,

code/go/0chain.net/smartcontract/storagesc/challenge.go
```diff
@@ -628,85 +716,278 @@ func (sc *StorageSmartContract) verifyChallenge(t *transaction.Transaction,
...
+			"error getting the blobber challenge list: %v", err)
+	}
+	if listSize, err := blobberChallengeList.Size(balances); err == nil && listSize == 0 {
+		Logger.Info("skipping generate challenge: empty blobber challenge partition")
 		return nil
 	}
 
-	if numMins == 0 {
-		numMins = 1
+	challengeID := encryption.Hash(hashString + strconv.FormatInt(int64(1), 10))
```
peterlimg: Unnecessary type convert from `int(1)` to string, is it needed if its value is always `1`?

code/go/0chain.net/smartcontract/storagesc/challenge.go
```diff
@@ -628,85 +716,278 @@ func (sc *StorageSmartContract) verifyChallenge(t *transaction.Transaction,
...
+	if err != nil {
+		return nil, err
+	}
+
+	if alloc == nil {
+		return nil, errors.New("empty allocation")
+	}
+
+	if blobberID == "" {
+		return nil, common.NewError("add_challenges",
```
peterlimg: do this empty checking earlier where it was acquired to avoid unnecessary works.

code/go/0chain.net/smartcontract/storagesc/models.go
```diff
@@ -99,10 +102,10 @@ type ChallengeResponse struct {
...
 
 type BlobberChallenge struct {
-	BlobberID                string                       `json:"blobber_id"`
-	Challenges               []*StorageChallenge          `json:"challenges"`
-	ChallengeMap             map[string]*StorageChallenge `json:"-" msg:"-"`
-	LatestCompletedChallenge *StorageChallenge            `json:"lastest_completed_challenge"`
+	BlobberID                string            `json:"blobber_id"`
+	LatestCompletedChallenge *StorageChallenge `json:"lastest_completed_challenge"`
+	ChallengeIDs             []string          `json:"challenge_ids"`
+	ChallengeIDMap           map[string]bool   `json:"-"`
```
peterlimg: Use struct{} instead of bool as the map value

code/go/0chain.net/smartcontract/storagesc/models.go
```diff
@@ -150,42 +153,109 @@ func (sn *BlobberChallenge) UnmarshalMsg(data []byte) ([]byte, error) {
...
+
+type AllocationChallenge struct {
+	AllocationID             string                       `json:"allocation_id"`
+	Challenges               []*StorageChallenge          `json:"challenges"`
+	ChallengeMap             map[string]*StorageChallenge `json:"-"`
+	LatestCompletedChallenge *StorageChallenge            `json:"lastest_completed_challenge"`
+}
+
+func (sn *AllocationChallenge) GetKey(globalKey string) datastore.Key {
+	return datastore.Key(globalKey + ":allocationchallenge:" + sn.AllocationID)
```
peterlimg: Unnecessary datastore.Key conversion as it's an alias of type `string`.

code/go/0chain.net/smartcontract/storagesc/models.go
```diff
@@ -442,8 +537,9 @@ type StorageAllocationStats struct {
 }
 
 type BlobberAllocation struct {
-	BlobberID       string                  `json:"blobber_id"`
-	AllocationID    string                  `json:"allocation_id"`
+	BlobberID    string `json:"blobber_id"`
+	AllocationID string `json:"allocation_id"`
+	// Size is blobber allocation maximum size
 	Size            int64                   `json:"size"`
```
peterlimg, line 543: If it means the maximum size, we'd better rename it to `Cap` or `MaxSize`, etc. Maybe change it in the future

peterlimg (changes requested):

code/go/0chain.net/smartcontract/storagesc/challenge.go
ssardana96: created an issue for the same: https://github.com/0chain/0chain/issues/1128

code/go/0chain.net/smartcontract/storagesc/challenge.go
ssardana96: this is fixed in: https://github.com/0chain/0chain/pull/1120

ssardana96:
Green check pasts: https://github.com/0chain/0chain/runs/5819707009?check_suite_focus=true

code/go/0chain.net/smartcontract/storagesc/challenge.go
ssardana96: fixed

peterlimg (approved):
lgtm

Sriep (approved):

Issue #993, opened 2022-02-09
Send events to event database when we complete block stake


Sriep:
https://0chain.slack.com/archives/C01GMABG2T1/p1644361635068779?thread_ts=1644296899.482209&cid=C01GMABG2T1

We should not execute insert events to db transaction when we compute block state (we can collect the events instead, like what we do for the MPT state). We should only do the db insert action when this block is finalized, otherwise we will see this error when two blocks for the same round get executed or one block executed twice due to round restarting.

peterlimg:
Closed by #1059

PR #1059, opened 2022-03-14
Add events on block finalizing process

peterlimg:
## Fixes
-  https://github.com/0chain/0chain/issues/993 
    Fix duplicate db events insert error, db events should be insert to db when the block is finalized, otherwise, sharders would continually reporting the duplicate transaction hash get insert error. 

## Tests 

### DIFF
--- a/code/go/0chain.net/chaincore/chain/protocol_block.go
+++ b/code/go/0chain.net/chaincore/chain/protocol_block.go
@@ -6,6 +6,7 @@ import (
 	"time"
 
 	"0chain.net/core/datastore"
+	"0chain.net/smartcontract/dbs/event"
 
 	"0chain.net/chaincore/config"
 	"0chain.net/chaincore/node"
@@ -344,6 +345,13 @@ func (c *Chain) finalizeBlock(ctx context.Context, fb *block.Block, bsh BlockSta
 
 	c.SetLatestOwnFinalizedBlockRound(fb.Round)
 	c.SetLatestFinalizedBlock(fb)
+
+	if len(fb.Events) > 0 && c.GetEventDb() != nil {
+		go func(events []event.Event) {
+			c.GetEventDb().AddEvents(ctx, events)
+		}(fb.Events)
+		fb.Events = nil
+	}
 
 	if fb.MagicBlock != nil {
 		if err := c.UpdateMagicBlock(fb.MagicBlock); err != nil {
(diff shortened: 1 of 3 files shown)


code/go/0chain.net/chaincore/chain/protocol_block.go
```diff
@@ -345,6 +346,13 @@ func (c *Chain) finalizeBlock(ctx context.Context, fb *block.Block, bsh BlockSta
 	c.SetLatestOwnFinalizedBlockRound(fb.Round)
 	c.SetLatestFinalizedBlock(fb)
 
+	if len(fb.Events) > 0 && c.GetEventDb() != nil {
+		go func(events []event.Event) {
+			c.GetEventDb().AddEvents(ctx, events)
+		}(fb.Events)
+		fb.Events = nil
```
Sriep, line 353: I think this might be better synchronous with a timeout. 
```go
eventDb := &EventDb{
  Store:         db,
  eventsChannel: make(chan events, 1000000),
}

	edb.eventsChannel <- events
}
```
If the timeout fails then the event database could be set to offline and can be repaired later. The difference in doing it asynchronously is that it always works even if it doesn't.

At the moment, if the event database slows down, then the events for a block could be entered out of order leading to a silently corrupt database rather than an offline database. 

We could also modify [addEventsWorker](https://github.com/0chain/0chain/blob/staging/code/go/0chain.net/smartcontract/dbs/event/process.go#L59) to confirm blocks of events come in order and do something sensible if not?

peterlimg: Since the blocks are processed in finalizing process, so actually they have already ordered, and because blocks will be queued in the `eventsChannel`, so it should be fine to do so. 

If we do it in synchronous way, I'm afraid the finalize process would be affected a lot(it's already slow, we can make it slower), then the whole chain speed would slow down.

service-0chain:
| config  | value |
| tests skipped  | FALSE  |
| system-tests  | fix/verify-miner-fees-test  |
| gosdk  | NOT OVERRIDDEN  |
| zboxcli  | staging  |
| zwalletcli  | staging  |
| 0chain  | fix/events-db-duplicate-keys  |
| blobber  | staging  |
| authorizer  | staging  |
| 0box  | staging  |
| 0block  | staging  |
| 0dns  | staging  |
| explorer  | staging  |
| 0proxy  | staging  |
| 0search  | staging  |

code/go/0chain.net/chaincore/chain/protocol_block.go
Sriep: If there is plenty of space on the `eventsChannel` there will likely be no problem either way. The scenario I envisage is the `eventsChannel` getting blocked so the AddEvents goroutines are building up,  all trying to add to their events to the channel. When the `eventsChannel` starts to clear the order that blocks of events get added will be undefined.

Maybe we are ok just to assume it won't happen, and if it does add an extra zero to eventsChannle size - which should be a setting not hardcoded.

Sriep (approved):

Issue #994, opened 2022-02-11
Fix linter errors so new commits can be checked correctly


peterlimg:
When linter errors and warnings count is less than 281, we will pass the linter check, this basically makes the linter unable to function. So we need to fix the lint errors as soon as possible so that all new commits could be checked by the linter.

ma2b0043:
@peterlimg when can we expect to close this issue?

peterlimg:
Closed via #1065

Issue #202, opened 2021-04-26
Fix lint issues


Sriep:
PR https://github.com/0chain/0chain/pull/184 gives us a pull request check to ensure that a PR does not increase the total number of lint issues.

To support this functionality we should address all the easy to fix lint issues in [0chain](https://github.com/0chain/0chain). Similar to https://github.com/0chain/blobber/pull/37

I would suggest doing one directory or package at a time rather than all at once.

Sriep:
When this is issue is resolved we should activate PR https://github.com/0chain/0chain/pull/184

Sriep:
Fixed by https://github.com/0chain/0chain/pull/184

ma2b0043:
@peterlimg @Sriep when can we expect to close this issue?

peterlimg:
Fixed via #1065

PR #1065, opened 2022-03-18
Fix all lint errors and warnings

peterlimg:
# Fixes: 
 
- #994 
- #202

# Changes:

---
  
- System test: https://github.com/0chain/system_test/pull/214

### DIFF
--- a/code/go/0chain.net/chaincore/block/magicBlock/main.go
+++ b/code/go/0chain.net/chaincore/block/magicBlock/main.go
@@ -17,15 +17,12 @@ import (
 
 	"0chain.net/chaincore/block"
 	"0chain.net/chaincore/client"
-	"0chain.net/chaincore/config"
 	"0chain.net/chaincore/node"
 	"0chain.net/chaincore/state"
 	"0chain.net/chaincore/threshold/bls"
 	"0chain.net/core/common"
-	"0chain.net/core/datastore"
 	"0chain.net/core/encryption"
 	"0chain.net/core/logging"
-	"go.uber.org/zap"
 	"gopkg.in/yaml.v2"
 )
 
@@ -38,7 +35,6 @@ type cmdMagicBlock struct {
 	dkgs map[string]*bls.DKG
 	// summaries collection
 	summaries       map[int]*bls.DKGSummary
-	states          *state.InitStates //nolint: structcheck,unused
 	originalIndices map[string]int
 }
 
@@ -160,23 +156,6 @@ func verifyKeys(hexSecKey, hexPubKey, hexId string) error {
 	if id != hexId {
 		return errors.New("id is not valid")
 	}
-	return nil
-}
-
-func verifySummaries(cmd *cmdMagicBlock, key datastore.Key, index int) error { //nolint unused
-
-	dkgs := cmd.summaries[index]
-	dkgs.ID = strconv.FormatInt(cmd.block.MagicBlockNumber, 10)
-
-	if err := dkgs.Verify(bls.ComputeIDdkg(key), cmd.block.Mpks.GetMpkMap()); err != nil {
-		if config.DevConfiguration.ViewChange {
-			logging.Logger.Error("Failed to verify genesis dkg", zap.Any("error", err))
-		} else {
-			logging.Logger.Panic(fmt.Sprintf("Failed to verify genesis dkg: ERROR: %v", err.Error()))
-		}
-
-	}
-
 	return nil
 }
 
--- a/code/go/0chain.net/chaincore/block/magicBlock/main.go
+++ b/code/go/0chain.net/chaincore/block/magicBlock/main.go
@@ -38,7 +38,7 @@ type cmdMagicBlock struct {
 	dkgs map[string]*bls.DKG
 	// summaries collection
 	summaries       map[int]*bls.DKGSummary
-	states          *state.InitStates
+	states          *state.InitStates //nolint: structcheck,unused
 	originalIndices map[string]int
 }
 
@@ -125,7 +125,9 @@ func (cmd *cmdMagicBlock) createShareOrSigns() {
 			cmd.summaries[node.SetIndex].SecretShares[partyId.GetHexString()] = share.GetHexString()
 			if mid != id {
 				var privateKey bls.Key
-				privateKey.SetHexString(cmd.yml.MinersMap[id].PrivateKey)
+				if err := privateKey.SetHexString(cmd.yml.MinersMap[id].PrivateKey); err != nil {
+					panic(err)
+				}
 				message := encryption.Hash(share.GetHexString())
 				sos.ShareOrSigns[id] = &bls.DKGKeyShare{
 					Message: message,
@@ -161,7 +163,7 @@ func verifyKeys(hexSecKey, hexPubKey, hexId string) error {
 	return nil
 }
 
-func verifySummaries(cmd *cmdMagicBlock, key datastore.Key, index int) error {
+func verifySummaries(cmd *cmdMagicBlock, key datastore.Key, index int) error { //nolint unused
 
 	dkgs := cmd.summaries[index]
 	dkgs.ID = strconv.FormatInt(cmd.block.MagicBlockNumber, 10)
@@ -197,7 +199,7 @@ func (cmd *cmdMagicBlock) setupBlockHash() {
 }
 
 // jsonMB method return indented json of block
-func (cmd *cmdMagicBlock) jsonMB() ([]byte, error) {
+func (cmd *cmdMagicBlock) jsonMB() ([]byte, error) { //nolint
 	return json.MarshalIndent(cmd.block, "", " ")
 }
 
@@ -411,7 +413,7 @@ func writeNames(names map[string]string) string {
 }
 
 func getNamesFileName() string {
-	return fmt.Sprintf("names.yaml")
+	return "names.yaml"
 }
 
 func getNamesEmailFileName(email string) string {
--- a/code/go/0chain.net/chaincore/block/magicBlock/main.go
+++ b/code/go/0chain.net/chaincore/block/magicBlock/main.go
@@ -116,13 +116,13 @@ func (cmd *cmdMagicBlock) createShareOrSigns() {
 			otherPartyId := bls.ComputeIDdkg(id)
 			share, err := cmd.dkgs[mid].ComputeDKGKeyShare(otherPartyId)
 			if err != nil {
-				panic(err)
+				log.Panic(err)
 			}
 			cmd.summaries[node.SetIndex].SecretShares[partyId.GetHexString()] = share.GetHexString()
 			if mid != id {
 				var privateKey bls.Key
 				if err := privateKey.SetHexString(cmd.yml.MinersMap[id].PrivateKey); err != nil {
-					panic(err)
+					log.Panic(err)
 				}
 				message := encryption.Hash(share.GetHexString())
 				sos.ShareOrSigns[id] = &bls.DKGKeyShare{
--- a/code/go/0chain.net/chaincore/chain/handler.go
+++ b/code/go/0chain.net/chaincore/chain/handler.go
@@ -946,8 +946,10 @@ func DiagnosticsDKGHandler(w http.ResponseWriter, r *http.Request) {
 <title>DKG process informations</title></head><body>
 <h1>DKG process disabled</h1></body></html>`)
 
-		w.Write(ss) //nolint: errcheck
-		return
+		if _, err := w.Write(ss); err != nil {
+			logging.Logger.Error("diagnostics DKG handler - http write failed", zap.Error(err))
+			return
+		}
 	}
 
 	var (
@@ -1716,9 +1718,8 @@ func StateDumpHandler(w http.ResponseWriter, r *http.Request) {
 
 	if contract == "" {
 		contract = "global"
-	} else { //nolint: staticcheck
-		//TODO: get the smart contract as an optional parameter and pick the right state hash
 	}
+
 	mptRootHash := util.ToHex(mpt.GetRoot())
 	fileName := fmt.Sprintf("mpt_%v_%v_%v.txt", contract, lfb.Round, mptRootHash)
 	file, err := ioutil.TempFile("", fileName)
--- a/code/go/0chain.net/chaincore/chain/handler.go
+++ b/code/go/0chain.net/chaincore/chain/handler.go
@@ -942,9 +942,11 @@ func DiagnosticsDKGHandler(w http.ResponseWriter, r *http.Request) {
 
 	if !config.DevConfiguration.ViewChange {
 		w.Header().Set("Content-Type", "text/html;charset=UTF-8")
-		w.Write([]byte(`<doctype html><html><head>
+		ss := []byte(`<doctype html><html><head>
 <title>DKG process informations</title></head><body>
-<h1>DKG process disabled</h1></body></html>`))
+<h1>DKG process disabled</h1></body></html>`)
+
+		w.Write(ss) //nolint: errcheck
 		return
 	}
 
@@ -1714,7 +1716,7 @@ func StateDumpHandler(w http.ResponseWriter, r *http.Request) {
 
 	if contract == "" {
 		contract = "global"
-	} else {
+	} else { //nolint: staticcheck
 		//TODO: get the smart contract as an optional parameter and pick the right state hash
 	}
 	mptRootHash := util.ToHex(mpt.GetRoot())
--- a/code/go/0chain.net/chaincore/chain/json_handler.go
+++ b/code/go/0chain.net/chaincore/chain/json_handler.go
@@ -436,5 +436,10 @@ func JSONHandler(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 	w.Header().Set("Content-Type", "application/json")
-	w.Write(data) //nolint: errcheck
+
+	if _, err = w.Write(data); err != nil {
+		logging.Logger.Error("http write failed",
+			zap.String("url", r.URL.String()),
+			zap.Error(err))
+	}
 }
--- a/code/go/0chain.net/chaincore/chain/json_handler.go
+++ b/code/go/0chain.net/chaincore/chain/json_handler.go
@@ -31,7 +31,7 @@ type home struct {
 	BuildTag          string            `json:"build_tag"`
 	StartTime         time.Time         `json:"start_time"`
 	NodeType          string            `json:"node_type"`
-	isDevMode         bool              `json:"is_dev_mode"`
+	IsDevMode         bool              `json:"is_dev_mode"`
 	CurrentMagicBlock *block.MagicBlock `json:"current_magic_block"`
 	Miners            []nodeInfo        `json:"miners"`
 	Sharders          []nodeInfo        `json:"sharders"`
@@ -52,7 +52,7 @@ func jsonHome(ctx context.Context) home {
 		BuildTag:          build.BuildTag,
 		StartTime:         StartTime,
 		NodeType:          node.Self.Underlying().Type.String(),
-		isDevMode:         config.Development(),
+		IsDevMode:         config.Development(),
 		CurrentMagicBlock: mb,
 		Miners:            miners,
 		Sharders:          sharders,
@@ -160,7 +160,7 @@ type roundHealth struct {
 	Shares        int    `json:"shares"`
 	VRFThreshold  int    `json:"vrf_threshold"`
 	LFBTicket     int64  `json:"lfb_ticket"`
-	isActive      bool   `json:"is_active"`
+	IsActive      bool   `json:"is_active"`
 }
 
 type ChainHealth struct {
@@ -249,7 +249,7 @@ func (c *Chain) getRoundHealth(ctx context.Context) roundHealth {
 	}
 
 	return roundHealth{
-		isActive:      isActive,
+		IsActive:      isActive,
 		Round:         rn,
 		VRFs:          vrfMsg,
 		RRS:           rrs,
@@ -436,5 +436,5 @@ func JSONHandler(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 	w.Header().Set("Content-Type", "application/json")
-	w.Write(data)
+	w.Write(data) //nolint: errcheck
 }
--- a/code/go/0chain.net/chaincore/chain/n2n_handler.go
+++ b/code/go/0chain.net/chaincore/chain/n2n_handler.go
@@ -83,7 +83,10 @@ func SetupX2XResponders(c *Chain) {
 //StateNodesHandler - return a list of state nodes
 func StateNodesHandler(ctx context.Context, r *http.Request) (interface{}, error) {
 	// this is needed as we get multiple values for the same key
-	r.ParseForm() //nolint: errcheck
+	if err := r.ParseForm(); err != nil {
+		return nil, err
+	}
+
 	nodes := r.Form["nodes"]
 	c := GetServerChain()
 	keys := make([]util.Key, len(nodes))
--- a/code/go/0chain.net/chaincore/chain/n2n_handler.go
+++ b/code/go/0chain.net/chaincore/chain/n2n_handler.go
@@ -82,7 +82,8 @@ func SetupX2XResponders(c *Chain) {
 
 //StateNodesHandler - return a list of state nodes
 func StateNodesHandler(ctx context.Context, r *http.Request) (interface{}, error) {
-	r.ParseForm() // this is needed as we get multiple values for the same key
+	// this is needed as we get multiple values for the same key
+	r.ParseForm() //nolint: errcheck
 	nodes := r.Form["nodes"]
 	c := GetServerChain()
 	keys := make([]util.Key, len(nodes))
--- a/code/go/0chain.net/chaincore/node/n2n_send.go
+++ b/code/go/0chain.net/chaincore/node/n2n_send.go
@@ -224,14 +224,23 @@ func SendEntityHandler(uri string, options *SendOptions) EntitySendHandler {
 		timeout = options.Timeout
 	}
 	return func(entity datastore.Entity) SendHandler {
-		data := getResponseData(options, entity).Bytes()
+		buf, err := getResponseData(options, entity)
+		if err != nil {
+			logging.N2n.Error("getResponseData failed", zap.Error(err))
+		}
+
+		data := buf.Bytes()
 
 		toPull := options.Pull
 		if len(data) > LargeMessageThreshold || toPull {
 			toPull = true
 			key := p2pKey(uri, entity.GetKey())
 			pdce := &pushDataCacheEntry{Options: *options, Data: data, EntityName: entity.GetEntityMetadata().GetName()}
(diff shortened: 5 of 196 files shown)


service-0chain:
| config  | value |
| tests skipped  | FALSE  |
| system-tests  | fix/interest_pool_update_config  |
| gosdk  | NOT OVERRIDDEN  |
| zboxcli  | staging  |
| zwalletcli  | staging  |
| 0chain  | fix/lint-errors  |
| blobber  | staging  |
| authorizer  | staging  |
| 0box  | staging  |
| 0block  | staging  |
| 0dns  | staging  |
| explorer  | staging  |
| 0proxy  | staging  |
| 0search  | staging  |

service-0chain:
| config  | value |
| tests skipped  | FALSE  |
| system-tests  | fix/interest_pool_update_config  |
| gosdk  | NOT OVERRIDDEN  |
| zboxcli  | staging  |
| zwalletcli  | staging  |
| 0chain  | fix/lint-errors  |
| blobber  | staging  |
| authorizer  | staging  |
| 0box  | staging  |
| 0block  | staging  |
| 0dns  | staging  |
| explorer  | staging  |
| 0proxy  | staging  |
| 0search  | staging  |

code/go/0chain.net/chaincore/block/magicBlock/main.go
```diff
@@ -38,7 +38,7 @@ type cmdMagicBlock struct {
 	dkgs map[string]*bls.DKG
 	// summaries collection
 	summaries       map[int]*bls.DKGSummary
-	states          *state.InitStates
+	states          *state.InitStates //nolint: structcheck,unused
```
Sriep: looks safe to remove,

code/go/0chain.net/chaincore/block/magicBlock/main.go
```diff
@@ -125,7 +125,9 @@ func (cmd *cmdMagicBlock) createShareOrSigns() {
 			cmd.summaries[node.SetIndex].SecretShares[partyId.GetHexString()] = share.GetHexString()
 			if mid != id {
 				var privateKey bls.Key
-				privateKey.SetHexString(cmd.yml.MinersMap[id].PrivateKey)
+				if err := privateKey.SetHexString(cmd.yml.MinersMap[id].PrivateKey); err != nil {
+					panic(err)
```
Sriep: The calling method uses `log.panic`, maybe we should be consistent?

code/go/0chain.net/chaincore/block/magicBlock/main.go
```diff
@@ -161,7 +163,7 @@ func verifyKeys(hexSecKey, hexPubKey, hexId string) error {
 	return nil
 }
 
-func verifySummaries(cmd *cmdMagicBlock, key datastore.Key, index int) error {
+func verifySummaries(cmd *cmdMagicBlock, key datastore.Key, index int) error { //nolint unused
```
Sriep: unexported and unused, safe to remove.

code/go/0chain.net/chaincore/block/magicBlock/main.go
```diff
@@ -197,7 +199,7 @@ func (cmd *cmdMagicBlock) setupBlockHash() {
 }
 
 // jsonMB method return indented json of block
-func (cmd *cmdMagicBlock) jsonMB() ([]byte, error) {
+func (cmd *cmdMagicBlock) jsonMB() ([]byte, error) { //nolint
```
Sriep, line 181: unused and unexported, safe to remove.

Sriep (changes requested):

code/go/0chain.net/chaincore/chain/handler.go
```diff
@@ -927,9 +927,11 @@ func DiagnosticsDKGHandler(w http.ResponseWriter, r *http.Request) {
 
 	if !config.DevConfiguration.ViewChange {
 		w.Header().Set("Content-Type", "text/html;charset=UTF-8")
-		w.Write([]byte(`<doctype html><html><head>
+		ss := []byte(`<doctype html><html><head>
 <title>DKG process informations</title></head><body>
-<h1>DKG process disabled</h1></body></html>`))
+<h1>DKG process disabled</h1></body></html>`)
+
+		w.Write(ss) //nolint: errcheck
```
Sriep: should handle this error?

code/go/0chain.net/chaincore/chain/handler.go
```diff
@@ -1699,7 +1701,7 @@ func StateDumpHandler(w http.ResponseWriter, r *http.Request) {
 
 	if contract == "" {
 		contract = "global"
-	} else {
+	} else { //nolint: staticcheck
 		//TODO: get the smart contract as an optional parameter and pick the right state hash
```
Sriep: This todo was added [3 years ago](https://github.com/0chain/0chain/blame/staging/code/go/0chain.net/chaincore/chain/handler.go#L1700), I think it is no longer of use to whoever put it there. If necessary we should replace with a comment before the `if` statement and remove the empty if block.

code/go/0chain.net/chaincore/chain/json_handler.go
```diff
@@ -436,5 +436,5 @@ func JSONHandler(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 	w.Header().Set("Content-Type", "application/json")
-	w.Write(data)
+	w.Write(data) //nolint: errcheck
```
Sriep: Should we not handle this error?

code/go/0chain.net/chaincore/chain/n2n_handler.go
```diff
@@ -82,7 +82,8 @@ func SetupX2XResponders(c *Chain) {
 
 //StateNodesHandler - return a list of state nodes
 func StateNodesHandler(ctx context.Context, r *http.Request) (interface{}, error) {
-	r.ParseForm() // this is needed as we get multiple values for the same key
+	// this is needed as we get multiple values for the same key
+	r.ParseForm() //nolint: errcheck
```
Sriep: Should we not  handle this error?

Sriep (changes requested):
/nolint: errcheck
There are at my count 87  places where errors have been ignored. I think errors should be handled by
* passing up the stack
* leaving a log
* acknowledge that you are ignoring the error using something like replace
```go
metrics.Register("finalization_lag", FinalizationLagMetric) //nolint: errcheck
```
with
```go
_ = metrics.Register("finalization_lag", FinalizationLagMetric)
```
which should remove the lint complaint.

peterlimg:
> /nolint: errcheck There are at my count 87 places where errors have been ignored. I think errors should be handled by
> 
> 
>     * acknowledge that you are ignoring the error using something like replace
> 
> with
> 
> ```go
> _ = metrics.Register("finalization_lag", FinalizationLagMetric)
> ```
> 
> which should remove the lint complaint.

I prefer to use //nolint: errcheck to remove the lint complaint as we can find them easily later in the future.

code/go/0chain.net/chaincore/node/n2n_send.go
```diff
@@ -518,7 +532,7 @@ func ToN2NReceiveEntityHandler(handler datastore.JSONEntityReqResponderF, option
 				})
 			}
 			// TODO:
-			root, _ := context.WithTimeout(common.GetRootContext(), 5*time.Second)
+			root, _ := context.WithTimeout(common.GetRootContext(), 5*time.Second) //nolint:govet
```
Sriep, line 535: I suspect govet is complaining about using an expression in a parameter. Try replacing `5*time.Second` with a constant.

code/go/0chain.net/chaincore/node/node.go
```diff
@@ -41,7 +38,7 @@ func DeregisterNode(nodeID string) {
 
 	return // TODO (sfxdx): temporary disable nodes deregistering
 
-	nodesMutex.Lock()
+	nodesMutex.Lock() //nolint: govet
```
Sriep: The [TODO](https://github.com/0chain/0chain/blame/staging/code/go/0chain.net/chaincore/node/node.go#L41) was added 2 years ago. I think it is no longer temporary. I suggest if there is nothing better remove everything after the `return`.

code/go/0chain.net/chaincore/node/node.go
```diff
@@ -50,7 +47,7 @@ func DeregisterNode(nodeID string) {
 func DeregisterNodes(keep map[string]struct{}) {
 	return // never deregister nodes for now
 
-	nodesMutex.Lock()
+	nodesMutex.Lock() //nolint: govet
```
Sriep: As above, this code was returned out two years. Recommend removing unless something better suggests itself.

code/go/0chain.net/chaincore/node/node.go
```diff
@@ -276,72 +273,19 @@ func (n *Node) SetLastActiveTime(lat time.Time) {
...
-
 /*NewNode - read a node config line and create the node */
 func NewNode(nc map[interface{}]interface{}) (*Node, error) {
 	node := Provider()
 	node.Type = nc["type"].(NodeType)
 	node.Host = nc["public_ip"].(string)
 	node.N2NHost = nc["n2n_ip"].(string)
 	node.Port = nc["port"].(int)
-	node.SetID(nc["id"].(string))
+	node.SetID(nc["id"].(string)) //nolint: errcheck
```
Sriep: I can't see any reason not to pass this error up the stack. `SetId` only seems to error if nc["id"].(string) has non hex characters so `hex.DecondeString` fails. And the node's id does not get set.

code/go/0chain.net/chaincore/node/node.go
```diff
@@ -616,7 +560,7 @@ func serveMetricKey(uri string) string {
 	return "p?" + uri
 }
 
-func isPullRequestURI(uri string) bool {
+func isPullRequestURI(uri string) bool { //nolint
```
Sriep: I think we can remove this unused method. Is it related to view change?

code/go/0chain.net/chaincore/node/node_pool.go
```diff
@@ -330,7 +330,7 @@ func (np *Pool) DecodeMsgpack(dec *msgpack.Decoder) error {
 }
 
 func (np *Pool) MarshalMsg(o []byte) ([]byte, error) {
-	d := poolDecode(*np)
+	d := poolDecode(*np) //nolint: govet
```
Sriep, line 333: `Call of 'poolDecode' copies the lock value: type 'Pool' contains 'sync.RWMutex' which is 'sync.Locker'`  (Goland inspection)
I suspect this lint issue will go away if [mmx      sync.RWMutex ](https://github.com/0chain/0chain/blob/staging/code/go/0chain.net/chaincore/node/node_pool.go#L36) is replaced with `mmx *sync.RWMutex`. (use pointer)

I think the idea is that the lock `d` ends up with is not the same one that `np` has. Which is a potential GOTCA.

This should also remove the issue on line 358.

code/go/0chain.net/chaincore/round/round_storage.go
```diff
@@ -127,7 +127,7 @@ func (s *roundStartingStorage) putToSlice(round int64) {
 	}
 }
 
-func (s *roundStartingStorage) check(round int64) error {
+func (s *roundStartingStorage) check(round int64) error { //nolint
```
Sriep, line 130: does not seem to be used.

code/go/0chain.net/conductor/config/node.go
```diff
@@ -111,10 +111,10 @@ func (n *Node) Stop() (err error) {
 		return fmt.Errorf("command %v: kill: %v", n.Name, err)
 	}
 	if stdin, ok := startCmd.Stdin.(*os.File); ok {
-		stdin.Close() // ignore error
+		stdin.Close() //nolint // ignore error
```
Sriep: Isn't this an example of where we want to deliberately ignore an error?

code/go/0chain.net/miner/m_handler.go
```diff
@@ -278,7 +282,7 @@ func VerifyBlockHandler(ctx context.Context, entity datastore.Entity) (
 		return nil, nil
 	}
 
-	var lfb = mc.GetLatestFinalizedBlock()
+	var lfb = mc.GetLatestFinalizedBlock() //nolint:typecheck
```
Sriep: What about trying
```go
var lfb  *block.Block  = mc.GetLatestFinalizedBlock()
```
maybe that will remove the lint complaint.

code/go/0chain.net/miner/miner_test.go
```diff
@@ -361,7 +361,7 @@ func setupTempRocksDBDir() func() {
 	}
 }
 
-func setupSelfNodeKeys() {
+func setupSelfNodeKeys() { //nolint
```
Sriep, line 364: This is only used in setupSelf which is unused and can be removed.

code/go/0chain.net/miner/miner_test.go
```diff
@@ -495,7 +495,7 @@ func SetUpSingleSelf() func() {
 	}
 }
 
-func setupSelf() func() {
+func setupSelf() func() { //nolint
```
Sriep, line 498: This function is unused so can be removed. Is there a good reason for keeping it.

code/go/0chain.net/miner/protocol_block.go
```diff
@@ -147,7 +156,7 @@ func (mc *Chain) VerifyBlockMagicBlockReference(b *block.Block) (err error) {
 
 	var (
 		round = b.Round
-		lfmbr = mc.GetLatestFinalizedMagicBlockRound(round)
+		lfmbr = mc.GetLatestFinalizedMagicBlockRound(round) //nolint:typecheck
```
Sriep: maybe replace with 
```go
  lfmbr *block.Block = mc.GetLatestFinalizedMagicBlockRound(round)
```

code/go/0chain.net/miner/protocol_block.go
```diff
@@ -385,7 +394,7 @@ func (mc *Chain) ValidateTransactions(ctx context.Context, b *block.Block) error
 				if cancel {
 					return
 				}
-				if mc.GetCurrentRound() > b.Round {
+				if mc.GetCurrentRound() > b.Round { //nolint: typecheck
```
Sriep: I can't make sense of this as both are `int64`. If you run the linter again does it still think this is a problem?

code/go/0chain.net/miner/protocol_bls.go
```diff
@@ -25,9 +25,8 @@ import (
 // ////////////  BLS-DKG Related stuff  /////////////////////
 
 var (
-	roundMap = make(map[int64]map[int]string)
+	roundMap = make(map[int64]map[int]string) //nolint
```
Sriep, line 28: Can't we remove this line?

code/go/0chain.net/miner/protocol_round.go
```diff
@@ -1536,7 +1524,7 @@ func (mc *Chain) kickRoundByLFB(ctx context.Context, lfb *block.Block) {
 		nr *Round
 	)
 
-	if !mc.ensureState(ctx, lfb) {
+	if !mc.ensureState(ctx, lfb) { //nolint: staticcheck
```
Sriep, line 1527: Maybe its complaining about the empty `if` block. We could replace this with
```go
  // ignore state error
  _ = mc.ensureState(ctx, lfb)
```
?

code/go/0chain.net/smartcontract/benchmark/main/cmd/control/benchmark_tests.go
```diff
@@ -84,7 +82,7 @@ func controlIndividual(balances cstate.StateContextI) error {
 		if err != nil {
 			return err
 		}
-		itArray = append(itArray, it)
+		itArray = append(itArray, it) //nolint
```
Sriep, line 85: Its complaining because itArray is never used. However we don't really want itArray we  are just timing how long it takes to make it. Maybe remove `itArray` and add a 
```go
_= it
```
line.

code/go/0chain.net/smartcontract/storagesc/helper_test.go
```diff
@@ -65,7 +60,7 @@ func newClient(balance state.Balance, balances chainState.StateContextI) (
 	client *Client) {
 
 	var scheme = encryption.NewBLS0ChainScheme()
-	scheme.GenerateKeys()
+	scheme.GenerateKeys() //nolint
```
Sriep, line 63: `scheme.GenerateKeys() ` does not seem able to return a non nil error, however, it needs to return an error to implement the [SignatureScheme](https://github.com/0chain/0chain/blob/staging/code/go/0chain.net/core/encryption/signature_scheme.go#L20) interface. Probably use `_ =` here. Its a unit test so  prioritise keeping the code simple to read or put in a panic.

code/go/0chain.net/smartcontract/storagesc/helper_test.go
```diff
@@ -185,7 +180,7 @@ func addBlobber(t testing.TB, ssc *StorageSmartContract, cap, now int64,
 	blob *Client) {
 
 	var scheme = encryption.NewBLS0ChainScheme()
-	scheme.GenerateKeys()
+	scheme.GenerateKeys() //nolint
```
Sriep, line 183: As above. Probably use `_ =`

code/go/0chain.net/smartcontract/storagesc/helper_test.go
```diff
@@ -215,7 +210,7 @@ func addValidator(t testing.TB, ssc *StorageSmartContract, now int64,
 	balances chainState.StateContextI) (valid *Client) {
 
 	var scheme = encryption.NewBLS0ChainScheme()
-	scheme.GenerateKeys()
+	scheme.GenerateKeys() //nolint
```
Sriep, line 213: As above. Probably use `_ =`

code/go/0chain.net/smartcontract/vestingsc/benchmark_setup.go
```diff
@@ -45,7 +45,7 @@ func AddVestingPools(
 			panic(err)
 		}
 
-		vestingPools = append(vestingPools, vestingPool.ID)
+		vestingPools = append(vestingPools, vestingPool.ID) //nolint: staticcheck
```
Sriep: vestingPools does not seem to be being used. I suggest we can remove the variable.

code/go/0chain.net/smartcontract/zcnsc/authorizer_test.go
```diff
@@ -45,7 +45,7 @@ func Test_BasicAuthorizersShouldBeInitialized(t *testing.T) {
 func Test_Basic_GetGlobalNode_InitsNode(t *testing.T) {
 	ctx := MakeMockStateContext()
 
-	node, err := GetGlobalSavedNode(ctx)
+	node, err := GetGlobalSavedNode(ctx) //nolint: typecheck
```
Sriep: There seems to be a lot of `typecheck` issues in this zcnsc. My guess the linter has found some ambiguities. Anyway suggest where the linter is complaining, be explicit and use var.
Maybe its confused between `*mockStateContext` and `StateContextI`? 
```go
var ctx state.StateContextI = MakeMockStateContext()
```
Same for other typecheck issues below.

Alternatively  changing MakeMockStateContext() to return `state.StateContextI` might be easier, and fix a large number of issues in one go. 

I guess if these zcnsc files got though, our existing lint check can't be that good! We could leave these issues, but someone ought to clean up the code at some point.

code/go/0chain.net/smartcontract/zcnsc/authorizer_test.go
```diff
@@ -101,7 +101,7 @@ func TestAuthorizerNodeShouldBeAbleToAddTransfer(t *testing.T) {
 	// Response params
 	responseDigPoolEqualityCheck(t, resp, tr, &an.Staking.ID, &stringEmpty)
 
-	err = sc.AddTransfer(transfer)
+	err = sc.AddTransfer(transfer) //nolint: typecheck
```
Sriep: Not sure exactly why its complaining, my guess its confused by this bit of code.
```go
	var transfer *state.Transfer
	transfer, resp, err := an.Staking.DigPool(tr.Hash, tr)
```
Maybe try being consistent here either remove `var transfer *state.Transfer` or maybe add `var resp string; var err error` and use `=` instead of `:=`

code/go/0chain.net/smartcontract/zcnsc/burn_test.go
```diff
@@ -36,7 +36,7 @@ func Test_FuzzyBurnTest(t *testing.T) {
 	contract := CreateZCNSmartContract()
 	ctx := MakeMockStateContext()
 
-	burn, err := contract.Burn(tr, payload.Encode(), ctx)
+	burn, err := contract.Burn(tr, payload.Encode(), ctx) //nolint: typecheck
```
Sriep: Same as in `authorizer_test.go`. If declaring ctx explicitly worked before then use it here as well.
```go
	var ctx cstate.StateContextI = MakeMockStateContext()
```

code/go/0chain.net/smartcontract/zcnsc/context_helpers_test.go
```diff
@@ -130,14 +105,14 @@ func (m *mockStateContext) GetTrieNode(key datastore.Key, v util.MPTSerializable
 	return util.ErrValueNotPresent
 }
 
-func MakeMockStateContext() *mockStateContext {
+func MakeMockStateContext() *mockStateContext { //nolint
```
Sriep: Not sure what it is being compained about here. But changing return type to state.StateContextI might fix all the other issues.

code/go/0chain.net/smartcontract/zcnsc/context_helpers_test.go
```diff
@@ -172,13 +147,13 @@ func MakeMockStateContext() *mockStateContext {
 	/// GetClientBalance
 
 	ctx.
-		On("GetClientBalance", mock.AnythingOfType("string")).
+		On("GetClientBalance", mock.AnythingOfType("string")). //nolint: typecheck
```
Sriep: Maybe we should replace `ctx.On` with ctx.StateContextI.On?  Or explicitly declaring
```go
var ctx  *mocks.StateContextI  := &mockStateContext{
  StateContextI: &mocks.StateContextI{},
}
```
Why are we wrapping the StateContextI anyway, bit confused.

code/go/0chain.net/smartcontract/zcnsc/context_helpers_test.go
```diff
@@ -336,11 +308,11 @@ func MakeMockStateContext() *mockStateContext {
 }
 
 func createTestAuthorizer(ctx *mockStateContext, id string) *Authorizer {
-	scheme := ctx.GetSignatureScheme()
+	scheme := ctx.GetSignatureScheme() //nolint: typecheck
```
Sriep: I suggest explicitly setting the type with `var`.

code/go/0chain.net/smartcontract/zcnsc/context_helpers_test.go
```diff
@@ -336,11 +308,11 @@ func MakeMockStateContext() *mockStateContext {
...
 	_ = scheme.GenerateKeys()
 
 	node := NewAuthorizer(id, scheme.GetPublicKey(), fmt.Sprintf("https://%s", id))
-	tr := CreateAddAuthorizerTransaction(defaultClient, ctx, 100)
+	tr := CreateAddAuthorizerTransaction(defaultClient, ctx, 100) //nolint: typecheck
```
Sriep: Again I suggest explicitly setting the type with `var`.

code/go/0chain.net/smartcontract/zcnsc/factory_helpers_test.go
```diff
@@ -44,7 +44,7 @@ func (n *Authorizer) Verify(sig, hash string) (bool, error) {
 }
 
 func CreateDefaultTransactionToZcnsc() *transaction.Transaction {
-	return CreateAddAuthorizerTransaction(defaultClient, MakeMockStateContext(), tokens)
+	return CreateAddAuthorizerTransaction(defaultClient, MakeMockStateContext(), tokens) //nolint: typecheck
```
Sriep: I assume the problem is still the ambiguity between ` state.StateContextI,` and ` *mockStateContext`. I guess we need to either explicitly declare the second parameter or change MakeMockStateContext's return type.

Sriep (changes requested):
```
I prefer to use //nolint: errcheck to remove the lint complaint as we can find them easily later in the future.
```
The best idea might be to use `_ =` when we `know` we want to ignore the error message, and keep `//nolint: errcheck` when we are'nt sure yet of the correct approch.

Also maybe we should link https://github.com/0chain/0chain/issues/202 to this PR.

code/go/0chain.net/chaincore/node/n2n_send.go
peterlimg: The govet is complaining the second returned value (the cancel callback function) is not called, usually we should call it with `defer cancel()` to avoid memory leak, but in this situation, we can not do it as the ctx's lifetime is longer than the function, and it won't cause memory leak issue actually.

code/go/0chain.net/chaincore/node/node.go
peterlimg: It's not being used, I will remove it.

code/go/0chain.net/chaincore/node/node.go
peterlimg: Update this would need to update all the unit tests that set ID as non hex encoded string, let me check.

code/go/0chain.net/smartcontract/zcnsc/context_helpers_test.go
peterlimg: We're not checking the lint errors on the `_test.go` files actually, there are too many warnings there so we won't fix them in this round.

code/go/0chain.net/smartcontract/zcnsc/factory_helpers_test.go
peterlimg: We will ignore all lint errors on test files, so should be fine.

code/go/0chain.net/chaincore/node/node.go
Sriep: So unit tests have been written that rely on this error not being reported? Oh, dear. While this lint issue is not in a `_test.go` file `NewNode` is only called from `_test.go` files but in various different packages. Ok, I see why you might want to leave it.

peterlimg: Fixed

Sriep (dismissed):

Sriep (dismissed):

Sriep (approved):

service-0chain:
| config  | value |
| tests skipped  | FALSE  |
| system-tests  | master  |
| gosdk  | NOT OVERRIDDEN  |
| zboxcli  | staging  |
| zwalletcli  | staging  |
| 0chain  | fix/lint-errors  |
| blobber  | staging  |
| authorizer  | staging  |
| 0box  | staging  |
| 0block  | staging  |
| 0dns  | staging  |
| explorer  | staging  |
| 0proxy  | staging  |
| 0search  | staging  |

PR #1069, opened 2022-03-20
ZCNSC stake pool & delegates

Dmdv:
## Tests 

- blobber:
- gosdk:
- system_test:
- zboxcli:
- zwalletcli:
- Other: ...


code/go/0chain.net/smartcontract/zcnsc/stakepool.go
```diff
@@ -0,0 +1,292 @@
+package zcnsc
+
+import (
+	"encoding/json"
+	"errors"
+	"fmt"
+
+	chainstate "0chain.net/chaincore/chain/state"
+	cstate "0chain.net/chaincore/chain/state"
```
peterlimg, line 8: `chainstate` and `cstate` are the same package, so use cstate perhaps.

peterlimg (changes requested):

code/go/0chain.net/smartcontract/zcnsc/stakepool.go
Dmdv: Fixed

code/go/0chain.net/smartcontract/zcnsc/authorizer.go
```diff
@@ -92,62 +78,84 @@ func (zcn *ZCNSmartContract) AddAuthorizer(tran *transaction.Transaction, inputD
...
+		//	return "", common.NewError(code, "saving stake pool error: "+err.Error())
+		//}
+	} else {
+
+		// compare the global min of authorizerNode Authorizer to that of the transaction amount
+		if globalNode.MinStakeAmount > state.Balance(tran.Value*1e10) {
+			msg := fmt.Sprintf("min stake amount '(%d)' > transaction value '(%d)'",
+				globalNode.MinStakeAmount, tran.Value)
+			err = common.NewError(code, msg)
+			Logger.Error("min stake amount > transaction value", zap.Error(err))
```
dabasov: most likely we will log error higher in the stack, so this can be removed `zap.Error(err)`

code/go/0chain.net/smartcontract/zcnsc/stakepool.go
```diff
@@ -0,0 +1,292 @@
...
+		return fmt.Errorf("min_stake is less than allowed by SC: %v < %v", min, gn.MinStakeAmount)
+	}
+	if max < min {
+		return fmt.Errorf("max_stake less than min_stake: %v < %v", min, max)
+	}
+
+	return
+}
+
+func (zcn *ZCNSmartContract) DistributeRewards(
```
dabasov: why is this method empty?

Dmdv: This is a draft PR

code/go/0chain.net/smartcontract/zcnsc/authorizer.go
Dmdv: No, this is the last call to log everything until the call is passed to transaction caller

ma2b0043:
@Dmdv when can we expect to close this PR?

code/go/0chain.net/smartcontract/zcnsc/stakepool.go
```diff
@@ -0,0 +1,284 @@
...
+	}
+
+	// the tokens can't be unlocked due to opened offers, but we mark it
+	// as 'unstake' and returns maximal time to wait to unlock the pool
+	if !unstake {
+		// save the pool and return special result
+		if err = sp.save(zcn.ID, spr.AuthorizerID, ctx); err != nil {
+			return "", common.NewErrorf(code, "saving stake pool: %v", err)
+		}
+		return toJson(&unlockResponse{Unstake: false}), nil
```
peterlimg: If tokens can't be unlocked due to opened offers, the `unstake` could still return true, it would only return false when error happens, please refer the `empty()` method for details. So we might need to update the comments above, and maybe the process logic if needed

code/go/0chain.net/smartcontract/zcnsc/authorizer.go
```diff
@@ -92,62 +78,84 @@ func (zcn *ZCNSmartContract) AddAuthorizer(tran *transaction.Transaction, inputD
...
-		return "", err
+		// create stake pool for the validator to count its rewards
+		//var sp *stakePool
+		//sp, err = zcn.getOrUpdateStakePool(globalNode, authorizerID, spenum.Authorizer, params.StakePoolSettings, ctx)
+		//if err != nil {
+		//	return "", common.NewError(code, "get or create stake pool error: "+err.Error())
+		//}
+		//if err = sp.save(zcn.ID, authorizerID, ctx); err != nil {
+		//	return "", common.NewError(code, "saving stake pool error: "+err.Error())
+		//}
```
peterlimg: return error early here to avoid the `else` below

code/go/0chain.net/smartcontract/zcnsc/authorizer.go
```diff
@@ -92,62 +85,110 @@ func (zcn *ZCNSmartContract) AddAuthorizer(tran *transaction.Transaction, inputD
...
-		Logger.Error("emitting event", zap.Error(err))
-		return "", err
+	authorizer, err = GetAuthorizerNode(authorizerID, ctx)
+	if err == nil && authorizer != nil {
+		errs := fmt.Errorf("authorizer(authorizerID: %v) already exists", authorizerID)
+		err = createOrUpdateStakePool()
+		if err != nil {
+			errs = multierror.Append(errs, errors.Wrap(err, "failed to get or create stake pool"))
+		} else {
+			errs = errors.Wrap(errs, "create or update stake pool completed successfully")
```
peterlimg: We don't need to add success msg to errors

code/go/0chain.net/smartcontract/zcnsc/authorizer.go
```diff
@@ -92,62 +85,110 @@ func (zcn *ZCNSmartContract) AddAuthorizer(tran *transaction.Transaction, inputD
...
+		}
+
+		if authorizer.UpdateStakePoolSettings(&authorizerStakingPoolSettings) {
+			err = authorizer.Save(ctx)
+			if err != nil {
+				errs = multierror.Append(errs, errors.Wrap(err, "failed to update stake pool settings"))
+			} else {
+				errs = errors.Wrap(errs, "update pool settings completed successfully")
```
peterlimg: no need to add success msg to errors

code/go/0chain.net/smartcontract/zcnsc/authorizer.go
```diff
@@ -92,62 +85,110 @@ func (zcn *ZCNSmartContract) AddAuthorizer(tran *transaction.Transaction, inputD
...
-		msg := fmt.Sprintf("error marshalling authorizer(authorizerID: %v) to event, err: %v", authorizerID, err)
-		err = common.NewError(code, msg)
-		Logger.Error("emitting event", zap.Error(err))
-		return "", err
+	authorizer, err = GetAuthorizerNode(authorizerID, ctx)
+	if err == nil && authorizer != nil {
+		errs := fmt.Errorf("authorizer(authorizerID: %v) already exists", authorizerID)
+		err = createOrUpdateStakePool()
+		if err != nil {
+			errs = multierror.Append(errs, errors.Wrap(err, "failed to get or create stake pool"))
```
peterlimg: Do not introduce new third-party package easily, use error slice instead. Basically we should keep our code base clean and reduce the dependency from third-parties.

code/go/0chain.net/smartcontract/zcnsc/stakepool.go
```diff
@@ -0,0 +1,313 @@
...
+	}
+
+	if sp.Settings.MaxNumDelegates != settings.MaxNumDelegates {
+		sp.Settings.MaxNumDelegates = settings.MaxNumDelegates
+		changed = true
+	}
+
+	if changed {
+		return sp, nil
+	} else {
```
peterlimg: avoid `if .. else..`, use the following style instead:

```go
if changed {
    return sp, nil
}

return nil, fmt.Errorf("...")
```

peterlimg (dismissed):

code/go/0chain.net/smartcontract/zcnsc/stakepool.go
Dmdv: updated

code/go/0chain.net/smartcontract/zcnsc/authorizer.go
Dmdv: Removed it and replaced with logging

code/go/0chain.net/smartcontract/zcnsc/authorizer.go
Dmdv: Ok

code/go/0chain.net/smartcontract/zcnsc/stakepool.go
Dmdv: @peterlimg Peter, I'm not following. Let's discuss in chat

code/go/0chain.net/smartcontract/zcnsc/stakepool.go
```diff
@@ -0,0 +1,313 @@
...
+	if sp.Settings.MaxNumDelegates != settings.MaxNumDelegates {
+		sp.Settings.MaxNumDelegates = settings.MaxNumDelegates
+		changed = true
+	}
+
+	if changed {
+		return sp, nil
+	}
+
+	return nil, fmt.Errorf("no changes have been made to stakepool for authorizerID (%s)", authorizerId)
```
Sriep: The function is called `getOrUpdateStakePool` but you return an error rather than the stake pool if it is unchanged.

I suggest changing the function's name to `createStakePool`.

code/go/0chain.net/smartcontract/zcnsc/authorizer.go
```diff
@@ -92,62 +85,110 @@ func (zcn *ZCNSmartContract) AddAuthorizer(tran *transaction.Transaction, inputD
...
-	ev, err := authorizer.ToEvent()
-	if err != nil {
-		msg := fmt.Sprintf("error marshalling authorizer(authorizerID: %v) to event, err: %v", authorizerID, err)
-		err = common.NewError(code, msg)
-		Logger.Error("emitting event", zap.Error(err))
-		return "", err
+	authorizer, err = GetAuthorizerNode(authorizerID, ctx)
+	if err == nil && authorizer != nil {
+		errs := fmt.Errorf("authorizer(authorizerID: %v) already exists", authorizerID)
+		err = createOrUpdateStakePool()
```
Sriep: Why do we need to touch the stake pool here? Aren't we about to return an 'authoriser already exits' error?

code/go/0chain.net/smartcontract/zcnsc/authorizer.go
```diff
@@ -92,62 +85,110 @@ func (zcn *ZCNSmartContract) AddAuthorizer(tran *transaction.Transaction, inputD
...
-				authorizerID,
-				recipientID,
-				transfer.ClientID,
-				transfer.ToClientID,
-			),
-		)
-		return "", err
+	// create stake pool for the validator to count its rewards'
+
+	var createOrUpdateStakePool = func() (err error) {
```
Sriep: Why is this a local function? I think this is adding confusion. 

We want to create a new stake pool for a new authoriser. So create a function that does exactly that. This can return an error if the stake pool already exists. Call it `createStakePool`.

code/go/0chain.net/smartcontract/zcnsc/authorizer.go
```diff
@@ -16,24 +22,25 @@ import (
...
 // AddAuthorizer sc API function
 // Transaction must include ClientID, ToClientID, PublicKey, Hash, Value
-// inputData is a publicKey in case public key in Tx is missing
-// Either PK or inputData must be present
-// balances have `GetTriedNode` implemented to get nodes
 // ContractMap contains all the SC addresses
 // ClientID is an authorizerID - used to search for authorizer
 // ToClient is an SC address
-func (zcn *ZCNSmartContract) AddAuthorizer(tran *transaction.Transaction, inputData []byte, ctx cstate.StateContextI) (string, error) {
+func (zcn *ZCNSmartContract) AddAuthorizer(
```
Sriep, line 23: I assume below that `AddAuthorizer` does exactly that adds an authoriser and returns an error if an authoriser already exits. 

If the intention is to multipurpose `AddAuthorizer` so that it also updates authorises then we should change the name. But I think one endpoint for one action works best, so I would suggest a separate `update_authoriser` method to change the stake pool settings. This would work the same as [update_blobbers](https://github.com/0chain/0chain/blob/staging/code/go/0chain.net/smartcontract/storagesc/sc.go#L280) in storagesc.

code/go/0chain.net/smartcontract/zcnsc/stakepool.go
```diff
@@ -0,0 +1,313 @@
...
+func (spr *stakePoolRequest) decode(p []byte) (err error) {
+	if err = json.Unmarshal(p, spr); err != nil {
+		return
+	}
+	return // ok
+}
+
+// ----------- LockingPool pool --------------------------
+
+type StakePool struct {
```
Sriep, line 42: Why do we need to wrap the stakePool.StakePool object?

For example, if we need new methods can't we
```go
type stakePool stakepool.StakePool
```
work as there are no new fields. Unlike [storagesc.stakePool ](https://github.com/0chain/0chain/blob/staging/code/go/0chain.net/smartcontract/storagesc/stakepool.go#L49) where there are two extra fields.

code/go/0chain.net/smartcontract/zcnsc/stakepool.go
```diff
@@ -0,0 +1,313 @@
...
+	return json.Unmarshal(input, sp)
+}
+
+// save the stake pool
+func (sp *StakePool) save(sscKey, providerID string, balances cstate.StateContextI) (err error) {
+	_, err = balances.InsertTrieNode(stakePoolKey(sscKey, providerID), sp)
+	return
+}
+
+// The stake() returns total stake size including delegate pools want to unstake.
```
Sriep: The base class has a [stake()](https://github.com/0chain/0chain/blob/staging/code/go/0chain.net/smartcontract/stakepool/stakepool.go#L215) method, probably a better idea to make it public.

code/go/0chain.net/smartcontract/zcnsc/stakepool.go
```diff
@@ -0,0 +1,313 @@
...
+	}
+	if max < min {
+		return fmt.Errorf("max_stake less than min_stake: %v < %v", min, max)
+	}
+
+	return
+}
+
+func (zcn *ZCNSmartContract) DistributeRewards(
+	t *transaction.Transaction,
```
Sriep: Can't you use the [DistributeRewards](https://github.com/0chain/0chain/blob/staging/code/go/0chain.net/smartcontract/stakepool/stakepool.go#L170) from the stakepool.StakePool object. So this can be removed.

code/go/0chain.net/smartcontract/zcnsc/sc.go
```diff
@@ -25,35 +25,72 @@ const (
...
 // InitSC ...
-func (zcn *ZCNSmartContract) InitSC() {}
+func (zcn *ZCNSmartContract) InitSC() {
+	zcn.smartContractFunctions[MintFunc] = zcn.Mint
+	zcn.smartContractFunctions[BurnFunc] = zcn.Burn
+	zcn.smartContractFunctions[AddAuthorizerFunc] = zcn.AddAuthorizer
+	zcn.smartContractFunctions[DeleteAuthorizerFunc] = zcn.DeleteAuthorizer
+	zcn.smartContractFunctions[UpdateGlobalConfigFunc] = zcn.UpdateGlobalConfig
+	zcn.smartContractFunctions[UpdateAuthorizerConfigFunc] = zcn.UpdateAuthorizerConfig
+	zcn.smartContractFunctions[DistributeRewardsFunc] = zcn.DistributeRewards
```
Sriep: This is not a smart contract endpoint. Rather we should call the stake pool's DistributeRewards method whenever we pay rewards to stakeholders.

Instead, we would want a `collect_rewards` endpoint that will transfer the rewards previously distributed to delegate wallets. Similar to storages [collect_reward](https://github.com/0chain/0chain/blob/staging/code/go/0chain.net/smartcontract/storagesc/collect_reward.go#L13).

Sriep (changes requested):

code/go/0chain.net/smartcontract/zcnsc/stakepool.go
Dmdv: According to out talk, this is resolved

code/go/0chain.net/smartcontract/zcnsc/authorizer.go
Dmdv: Updated

code/go/0chain.net/smartcontract/zcnsc/stakepool.go
Dmdv: Updated. it's changes now

code/go/0chain.net/smartcontract/zcnsc/authorizer.go
Dmdv: True. Violation of single responsibility conception. I will create a new function to update stake pool settings

code/go/0chain.net/smartcontract/zcnsc/authorizer.go
Dmdv: Local function is a closure to reduce the overabundance of arguments to be passed.
But as I mentioned above, this is single responsibility violation. I will create another smart contract function to add staking pool if authorizer exists

code/go/0chain.net/smartcontract/zcnsc/authorizer.go
Dmdv: Ok

code/go/0chain.net/smartcontract/zcnsc/stakepool.go
Dmdv: Ok

code/go/0chain.net/smartcontract/zcnsc/stakepool.go
Dmdv: Removed

code/go/0chain.net/smartcontract/zcnsc/sc.go
Dmdv: Done

code/go/0chain.net/smartcontract/zcnsc/stakepool.go
Dmdv: OK

code/go/0chain.net/smartcontract/zcnsc/sc.go
Dmdv: Done

code/go/0chain.net/smartcontract/zcnsc/stakepool.go
Dmdv: I'd prefer leaving this because we have to implement utils.MPTSerializable with many other involved just repeating the code otherwise. Composite inheritance helps here to get already implemented functions for free

code/go/0chain.net/smartcontract/zcnsc/sc.go
```diff
@@ -25,35 +25,76 @@ const (
...
+	// Bridge related
+	zcn.smartContractFunctions[MintFunc] = zcn.Mint
+	zcn.smartContractFunctions[BurnFunc] = zcn.Burn
+	// Authorizer
+	zcn.smartContractFunctions[AddAuthorizerFunc] = zcn.AddAuthorizer
+	zcn.smartContractFunctions[DeleteAuthorizerFunc] = zcn.DeleteAuthorizer
+	// StakePool
+	zcn.smartContractFunctions[AddOrUpdateStakePoolFunc] = zcn.AddOrUpdateAuthorizerStakePool
+	// Rewards
+	zcn.smartContractFunctions[CollectRewardsFunc] = zcn.CollectRewards
```
Sriep, line 66: We should add an entry to [BenchmarkTests](https://github.com/0chain/0chain/blob/staging/code/go/0chain.net/smartcontract/zcnsc/benchmark_tests.go#L57) for new endponts.

code/go/0chain.net/smartcontract/zcnsc/sc.go
```diff
@@ -25,35 +25,76 @@ const (
...
+	zcn.smartContractFunctions[UpdateGlobalConfigFunc] = zcn.UpdateGlobalConfig
+	zcn.smartContractFunctions[UpdateAuthorizerConfigFunc] = zcn.UpdateAuthorizerConfig
+	// Bridge related
+	zcn.smartContractFunctions[MintFunc] = zcn.Mint
+	zcn.smartContractFunctions[BurnFunc] = zcn.Burn
+	// Authorizer
+	zcn.smartContractFunctions[AddAuthorizerFunc] = zcn.AddAuthorizer
+	zcn.smartContractFunctions[DeleteAuthorizerFunc] = zcn.DeleteAuthorizer
+	// StakePool
+	zcn.smartContractFunctions[AddOrUpdateStakePoolFunc] = zcn.AddOrUpdateAuthorizerStakePool
```
Sriep: This should only be `updateAuthorizerStakePool`

The authoriser's stakePool is automatically created during `AddAuthorizer`. This means that a user can immidaitly lock tokens in a new authoriser.

code/go/0chain.net/smartcontract/zcnsc/nodes.go
```diff
@@ -188,22 +190,22 @@ func (c *AuthorizerConfig) Decode(input []byte) (err error) {
...
-	ID        string                    `json:"id"`
-	PublicKey string                    `json:"public_key"`
-	Staking   *tokenpool.ZcnLockingPool `json:"staking"`
-	URL       string                    `json:"url"`
-	Config    *AuthorizerConfig         `json:"config"`
+	ID          string                    `json:"id"`
+	PublicKey   string                    `json:"public_key"`
+	URL         string                    `json:"url"`
+	Config      *AuthorizerConfig         `json:"config"`
+	LockingPool *tokenpool.ZcnLockingPool `json:"lock_pool"`
```
Sriep: What is the purpose of this LockingPool? Shouldn't this be a StakePool object?

code/go/0chain.net/smartcontract/zcnsc/authorizer.go
```diff
@@ -104,7 +100,8 @@ func (zcn *ZCNSmartContract) AddAuthorizer(tran *transaction.Transaction, inputD
 
 	// Dig pool for authorizer
 
-	transfer, response, err := authorizer.Staking.DigPool(tran.Hash, tran)
+	var transfer *state.Transfer
```
Sriep: If I understand what is happening here correctly all authorizers should start with a default delegate. 

In which case this locking pool needs to be replaced with a stakepoolStakePool object. Then instead of calling `LockingPool.DigPoo`l, we can call [StakePool.Lock](https://github.com/0chain/0chain/blob/staging/code/go/0chain.net/smartcontract/stakepool/lock.go#L44). 

The values for [StakePool.Settings](https://github.com/0chain/0chain/blob/staging/code/go/0chain.net/smartcontract/stakepool/stakepool.go#L29) can either be added to the input or inherited from the Global node. The global node settings are the bonds on individual authoriser stakePool settings.

This way an authoriser always has a stakePool hence `updateAuthorizerStakePool` not `AddOrUpdateAuthorizerStakePool`.

code/go/0chain.net/smartcontract/zcnsc/authorizer.go
```diff
@@ -51,7 +52,7 @@ func (zcn *ZCNSmartContract) AddAuthorizer(tran *transaction.Transaction, inputD
 	// Decode input
 
 	params := AuthorizerParameter{}
-	err = params.Decode(inputData)
```
Sriep, line 54: We might want to add [stakePool settings](https://github.com/0chain/0chain/blob/staging/code/go/0chain.net/smartcontract/stakepool/stakepool.go#L33) to the input. These could then be validated against the bounds in the global node before creating the authoriser's stakePool.

Sriep (changes requested):

code/go/0chain.net/smartcontract/zcnsc/sc.go
Dmdv: Renamed

code/go/0chain.net/smartcontract/zcnsc/nodes.go
Dmdv: Same as with miners: when a miners is added, some amount from miners wallet is being locked in TokenlockPool

code/go/0chain.net/smartcontract/zcnsc/authorizer.go
Dmdv: Done

code/go/0chain.net/smartcontract/zcnsc/authorizer.go
Dmdv: Done

code/go/0chain.net/smartcontract/zcnsc/sc.go
Dmdv: I'd like to add benchmarks in a separate PR

Sriep (approved):

Issue #1063, opened 2022-03-16
update cost related config


dabasov:
Cost related settings in the sc.yaml should be updatable the same way the other settings are updatable. Implement for smart contracts:
- faucetsc
- interestpoolsc
- minersc
- storagesc
- vestingsc
- zcnsc

Cost section example
```
    cost:
      add_miner: 100
      add_sharder: 100
      delete_miner: 100
      delete_sharder: 100
      miner_health_check: 100
      sharder_health_check: 100
      contributeMpk: 100
      shareSignsOrShares: 100
      wait: 100
      update_globals: 100
      update_settings: 100
      update_miner_settings: 100
      update_sharder_settings: 100
      payFees: 100
      feesPaid: 100
      mintedTokens: 100
      addToDelegatePool: 100
      deleteFromDelegatePool: 100
      sharder_keep: 100
```

ma2b0043:
@dabasov when can we expect to close this issue?

iamrz1:
> @dabasov when can we expect to close this issue?

I'm working on a PR to address this. We will have it merged by the end of this coming week.

peterlimg:
Closed via #1076

PR #1076, opened 2022-03-24
Update cost related config of smart contract

iamrz1:
## Changes
- Makes cost related settings in the sc.yaml updatable the same way the other settings are updatable in:
  - faucetsc
  - interestpoolsc
  - minersc 
  - storagesc 
  - vestingsc 
  - zcnsc

## Fixes
https://github.com/0chain/0chain/issues/1063

### DIFF
--- a/code/go/0chain.net/smartcontract/faucetsc/models.go
+++ b/code/go/0chain.net/smartcontract/faucetsc/models.go
@@ -5,6 +5,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"strconv"
+	"strings"
 	"time"
 
 	"0chain.net/core/common"
@@ -103,10 +104,36 @@ func (gn *GlobalNode) updateConfig(fields map[string]string) error {
 			gn.OwnerId = value
 
 		default:
-			return fmt.Errorf("key %s not recognised as setting", key)
+			return gn.setCostValue(key, value)
 		}
 	}
 	return nil
+}
+
+func (gn *GlobalNode) setCostValue(key, value string) error {
+	if !strings.HasPrefix(key, Settings[Cost]) {
+		return fmt.Errorf("key %s not recognised as setting", key)
+	}
+	costKey := strings.ToLower(strings.TrimPrefix(key, Settings[Cost]+"."))
+	for _, costFunction := range costFunctions {
+		if costKey != strings.ToLower(costFunction) {
+			continue
+		}
+		costValue, err := strconv.Atoi(value)
+		if err != nil {
+			return fmt.Errorf("key %s, unable to convert %v to integer", key, value)
+		}
+
+		if costValue < 0 {
+			return fmt.Errorf("cost.%s contains invalid value %s", key, value)
+		}
+
+		gn.Cost[costKey] = costValue
+
+		return nil
+	}
+
+	return fmt.Errorf("cost config setting %s not found", costKey)
 }
 
 func (gn *GlobalNode) validate() error {
@@ -124,6 +151,7 @@ func (gn *GlobalNode) validate() error {
 	case gn.GlobalReset < gn.IndividualReset:
 		return common.NewError("failed to validate global node", fmt.Sprintf("global reset(%v) is less than individual reset(%v)", gn.GlobalReset, gn.IndividualReset))
 	}
+
 	return nil
 }
 
--- a/code/go/0chain.net/smartcontract/interestpoolsc/node.go
+++ b/code/go/0chain.net/smartcontract/interestpoolsc/node.go
@@ -5,6 +5,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"strconv"
+	"strings"
 	"time"
 
 	"0chain.net/chaincore/state"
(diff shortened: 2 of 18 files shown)


code/go/0chain.net/smartcontract/faucetsc/models.go
```diff
@@ -113,6 +114,16 @@ func (gn *GlobalNode) updateConfig(fields map[string]string) error {
 			gn.OwnerId = value
 
 		default:
+			if strings.HasPrefix(key, Settings[Cost]) {
+				costKey := strings.TrimPrefix(key, Settings[Cost]+".")
```
dabasov: We should validate that costKey is one of `costFunctions `

iamrz1: We are doing that in the validation function against a defined array of keys, so only values of matching keys get updated. Link:
https://github.com/0chain/0chain/blob/3b9abebecb077730fe177bbddd3bce816f802bff/code/go/0chain.net/smartcontract/faucetsc/models.go#L156

code/go/0chain.net/smartcontract/faucetsc/models.go
```diff
@@ -113,6 +114,16 @@ func (gn *GlobalNode) updateConfig(fields map[string]string) error {
...
+				costValue, err := strconv.Atoi(value)
+				if err != nil {
+					return fmt.Errorf("key %s, unable to convert %v to integer", key, value)
+				}
+				gn.Cost[costKey] = costValue
```
peterlimg: Even the validateCostConfig will filter it out, we should still check if `costKey` is in the `costFunctions` to  return error early here.

peterlimg (changes requested):

code/go/0chain.net/smartcontract/faucetsc/models.go
iamrz1: sure

code/go/0chain.net/smartcontract/storagesc/config_settigns.go
```diff
@@ -448,8 +557,17 @@ func (conf *Config) set(key string, change string) error {
 		} else {
 			return fmt.Errorf("cannot convert key %s value %v to boolean: %v", key, change, err)
 		}
+	case smartcontract.Cost:
+		if key == SettingName[Cost] {
```
peterlimg, line 558: I think the key value is `cost.*`, checking if its "cost" will always fail here.

code/go/0chain.net/smartcontract/storagesc/config_settigns.go
```diff
@@ -544,6 +662,8 @@ func (conf *Config) get(key Setting) interface{} {
 		return conf.BlockReward.BlobberUsageWeight
 	case ExposeMpt:
 		return conf.ExposeMpt
+	case Cost:
```
peterlimg, line 662: So we support setting cost with key like `cost.update_settings`, but we only support getting the whole cost map?

code/go/0chain.net/smartcontract/faucetsc/models.go
```diff
@@ -113,12 +114,39 @@ func (gn *GlobalNode) updateConfig(fields map[string]string) error {
 			gn.OwnerId = value
 
 		default:
+			if strings.HasPrefix(key, Settings[Cost]) {
```
peterlimg: Do the strings.HasPrefix in the `gn.setCostValue` to make this easier to read.

code/go/0chain.net/smartcontract/interestpoolsc/node.go
```diff
@@ -100,11 +100,38 @@ func (gn *GlobalNode) set(key string, value string) error {
 		}
 		gn.MaxMint = state.Balance(fValue * 1e10)
 	default:
+		if strings.HasPrefix(key, Settings[Cost]) {
```
peterlimg: do prefix check in setCostValue method

code/go/0chain.net/smartcontract/minersc/settings.go
```diff
@@ -211,6 +279,15 @@ func (gn *GlobalNode) set(key string, change string) error {
 		}
 		gn.setKey(key, change)
 	case smartcontract.Cost:
+		if key == SettingName[Cost] {
```
peterlimg, line 287: key is not "cost"

peterlimg (changes requested):

ma2b0043:
@iamrz1 @dabasov @peterlimg when can we expect to close this PR?

code/go/0chain.net/smartcontract/storagesc/config_settigns.go
iamrz1: It will not, the type has been defined as Cost using enum.

code/go/0chain.net/smartcontract/storagesc/config_settigns.go
iamrz1: Yes, for now it is this way. Basically that's how I found the output for cost in previous bit of codes. but I'm going to have a discussion with @dabasov about it, whether map or cost.* should be the way we send output back for get requests.

code/go/0chain.net/smartcontract/faucetsc/models.go
iamrz1: That will not work as I need the return statement to work only if it has the prefix, otherwise I need it to return the error from the next line.

code/go/0chain.net/smartcontract/interestpoolsc/node.go
iamrz1: That will not work as I need the return statement to work only if it has the prefix, otherwise I need it to return the error from the next line.

code/go/0chain.net/smartcontract/minersc/settings.go
iamrz1: yes, key is not cost, it is cost.*

iamrz1:
> @iamrz1 @dabasov @peterlimg when can we expect to close this PR?

By the end of this week.

code/go/0chain.net/smartcontract/storagesc/config_settigns.go
iamrz1: output updated to cost.function_name, for all the values in cost map.

dabasov (commented):
lgtm

code/go/0chain.net/smartcontract/faucetsc/models.go
peterlimg: The way I suggested was do it like this:
```go
func (gn *GlobalNode) setCostValue(key, value string) error {
    if !strings.HasPrefix(key, Settings[Cost]) {
      return fmt.Errorf("key %s not recognised as setting", key)
  }
  // do the same as before
}

func updateConfig(..) error {
...
  switch {
    ...
    default:
      return gn.setCostValue()
  }
...
}
```

The reasons I suggest to move the prefix checking to `gn.setCostValue` are:
a) if the `gn.setCostValue()` is called in more than one place, you will need to do the prefix checking in all the places that it is called, although it is only used in `updateConfig` now.
b) the code will be cleaner
c) The `setCostValue()` actually should do a prefix checking itself to avoid unnecessary work instead of relying on the checking by others.

iamrz1: got it, will have it changed.

iamrz1: done

dabasov (dismissed):
lgtm

peterlimg (approved):
lgtm

PR #1078, opened 2022-03-25
Fix: #1077 update blobber stake

### LINKED ISSUE #1077: Total stake should be saved in blobber
We need to keep the[ total stake ](https://github.com/0chain/0chain/blob/staging/code/go/0chain.net/smartcontract/storagesc/stakepool.go#L101)on the blobber object. Update the blobber each time it changes. Then we can use the equation from [cleanCapacity](https://github.com/0chain/0chain/blob/staging/code/go/0chain.net/smartcontract/storagesc/stakepool.go#L226) directly in an SQL where clause.
So somewhere, [PR960](https://github.com/0chain/0chain/pull/960) or  [PR684](https://github.com/0chain/0chain/pull/684) or somewhere eilse, we need to add a TotalStake field to the blobber and update the object each time a stake changes - usually stakePoolLock and stakePoolUnlock.

https://0chain.slack.com/archives/C02QDMYFQJG/p1647819416329659

samtholiya:
## Tests 

- blobber:
- gosdk: https://github.com/0chain/gosdk/pull/421
- system_test:
- zboxcli: https://github.com/0chain/zboxcli/pull/203
- zwalletcli:
- Other: ...


code/go/0chain.net/smartcontract/storagesc/stakepool.go
```diff
@@ -460,7 +460,11 @@ func (ssc *StorageSmartContract) stakePoolLock(t *transaction.Transaction,
 	}
 
 	// TO-DO: Update stake in eventDB
-
+	data, _ := json.Marshal(event.UpdateTotalStake{
```
Sriep, line 463: Can you reuse ` event.TagUpdateBlobber`. 
```go
	data, err := json.Marshal(&dbs.DbUpdates{
		Id: sn.ID,
		Updates: map[string]interface{}{
			"total_stake":                  sp.stake(),
		},
	})
	if err != nil {
		return fmt.Errorf("marshalling update: %v", err)
	}
	balances.EmitEvent(event.TypeStats, event.TagUpdateBlobber, sn.ID, string(data))
```
This should mean you don't need any `dbs/event` changes.

Also once you have finished `// TO-DO: Update stake in eventDB` can be removed.

Sriep (approved):

Sriep (changes requested):
Can you reuse ` event.TagUpdateBlobber`. 
```go
	data, err := json.Marshal(&dbs.DbUpdates{
		Id: sn.ID,
		Updates: map[string]interface{}{
			"total_stake":                  sp.stake(),
		},
	})
	if err != nil {
		return fmt.Errorf("marshalling update: %v", err)
	}
	balances.EmitEvent(event.TypeStats, event.TagUpdateBlobber, sn.ID, string(data))
```
This should mean you don't need any `dbs/event` changes.

Also once you have finished `// TO-DO: Update stake in eventDB` can be removed.

code/go/0chain.net/smartcontract/storagesc/models.go
```diff
@@ -342,6 +342,7 @@ type StorageNode struct {
 	SavedData       int64                  `json:"saved_data"`
 	// StakePoolSettings used initially to create and setup stake pool.
 	StakePoolSettings stakepool.StakePoolSettings `json:"stake_pool_settings"`
+	TotalStake        int64                       `json:"total_stake"`
```
peterlimg: Run `make msgp` to update the `_gen.go` file, if you don't have msgp installed, please run `make install-msgp` in advance. 

I don't see the `TotalStake` is being used other than the event db, please update related code that would update it, otherwise, we'd better keep the MPT data struct untouched.

peterlimg (changes requested):

code/go/0chain.net/smartcontract/storagesc/models.go
samtholiya: @peterlimg we will be using total stake in newAllocation pr right... also we added the zboxcli in bl-info


But thanks for the msgp tip will do that

samtholiya: also the total stake is updated https://github.com/0chain/0chain/pull/1078/files#diff-ce2ed4fb820ddf8bc8517075d5242e73b2e7547262aababb0be2e39ea481a56dR469
Here if that is what you missed... not sure if thatis what you were searching for

peterlimg: Hi @samtholiya, can you share the newAllocation PR. BTW, the code you referenced is still the place for update the total stake in event db. 

I suppose there are some place (perhaps block rewards) the `blobbler.TotalStake` is updated and saved to MPT, then emit the total stake changes db events, then save it to events db, but in this PR I  only see the `TotalStake` field is added to MPT struct, and only event db is using it.  @Sriep can you check

ma2b0043:
@samtholiya @peterlimg @Sriep when can we expect to close this PR?

code/go/0chain.net/smartcontract/storagesc/models.go
Sriep: We don't need to keep ToptalStake in `StorageNode`, as we can always use [stakePool.stakes()](https://github.com/0chain/0chain/blob/staging/code/go/0chain.net/smartcontract/storagesc/stakepool.go#L101) instead, however, it will need to be part of the object returned by GetBlobberHandler; hence probably easier to add to `StorageNode`. 

Whichever, the important thing is that we emit a `TagUpdateBlobber` event every time a delegate pool's balance changes.

We will at least need to know the total stake in the new allocation request sc in order to validate the input blobbers. We do not need it for paying rewards, as in block rewards as this is added to a separate reward account, not the pool's balance.

Sriep:
This needs to be merged together with the [zbox and goSDK changes](https://0chain.slack.com/archives/C02QDMYFQJG/p1648218267843289?thread_ts=1647819416.329659&cid=C02QDMYFQJG).

In particular, this will allow local integration testing using the updated `zbox bl-info` command. Once we have eyeballed the blobber total stake field returning a correct value in a local 0chain and the trio of PRs have passed system tests then we should be good to go.

[We should add the total-stake](https://github.com/0chain/zboxcli/blob/staging/cmd/storage.go#L169) to fields prited out in zbox bl-info. This should make it easy to test the changes.
That means we need to [change the goSDK object](https://github.com/0chain/gosdk/blob/staging/zboxcore/sdk/sdk.go#L651). 
Two one line changes.  Fortunately, we are adding a field so it is not a breaking change for system tests.

code/go/0chain.net/chaincore/chain/config.go
```diff
@@ -9,7 +9,6 @@ import (
 	"go.uber.org/zap"
 
 	"0chain.net/smartcontract/dbs"
-
```
Sriep, line 12: We don't need this change, can we remove the file from the PR.

code/go/0chain.net/smartcontract/dbs/event/types.go
```diff
@@ -1,4 +1,4 @@
-package dbs
+package event
```
Sriep, line 1: Why are we moving the DbUpdate type into the evnet package?


 The idea of keeping it in the dbs package is to avoid cyclic dependencies, as the type is also used in [storagesc](https://github.com/0chain/0chain/blob/staging/code/go/0chain.net/smartcontract/storagesc/eventdb.go#L48)

The dbs package should not import any other 0chain pacakges so safe to include anywhere.

Sriep (changes requested):

code/go/0chain.net/smartcontract/dbs/event/types.go
samtholiya: https://0chain.slack.com/archives/C02QDMYFQJG/p1648229723268449?thread_ts=1647819416.329659&cid=C02QDMYFQJG

we have an issue we are importing higher package in the lower package this should not be done
we are importing [0chain.net/smartcontract/dbs](http://0chain.net/smartcontract/dbs) in [0chain.net/smartcontract/dbs/event](http://0chain.net/smartcontract/dbs/event)
This is not a good practice

samtholiya:
@Sriep here are the pr you are looking for https://0chain.slack.com/archives/C02QDMYFQJG/p1648247973911919?thread_ts=1647819416.329659&cid=C02QDMYFQJG

https://github.com/0chain/gosdk/pull/421
https://github.com/0chain/zboxcli/pull/203

code/go/0chain.net/smartcontract/storagesc/models.go
peterlimg: If the `TotalStake` is only used by `GetBlobberHandler` and for API only, we'd better it out of MPT, the principle is if not necessary, please do not change MPT data struct.

peterlimg:
> @Sriep here are the pr you are looking for https://0chain.slack.com/archives/C02QDMYFQJG/p1648247973911919?thread_ts=1647819416.329659&cid=C02QDMYFQJG
> 
> [0chain/gosdk#421](https://github.com/0chain/gosdk/pull/421) 
> [0chain/zboxcli#203](https://github.com/0chain/zboxcli/pull/203)

Please fill the PR template about the associated PRs, that's why we create the template, thanks.

samtholiya:
@debasov 
I was of the mindset that we call a pr associated only if "current" pr if merged would require other pr's to merge so as to not break it so I did not put them as this pr is independent and is actually required in new Allocation pr. But thanks for clarification.
Will make sure this does not happen next time
Should we also add the pr which are blocked because of this change?

peterlimg:
> @debasov I was of the mindset that we call a pr associated only if "current" pr if merged would require other pr's to merge so as to not break it so I did not put them as this pr is independent and is actually required in new Allocation pr. But thanks for clarification. Will make sure this does not happen next time Should we also add the pr which are blocked because of this change?

No worries, yes, if any PRs that are blocked by this PR, please add them too.

service-0chain:
| config  | value |
| tests skipped  | FALSE  |
| system-tests  | master  |
| gosdk  | totalstakeBlobber  |
| zboxcli  | displaytotalstake  |
| zwalletcli  | staging  |
| 0chain  | addStakepool  |
| blobber  | staging  |
| authorizer  | staging  |
| 0box  | staging  |
| 0block  | staging  |
| 0dns  | staging  |
| explorer  | staging  |
| 0proxy  | staging  |
| 0search  | staging  |

code/go/0chain.net/smartcontract/storagesc/handler.go
```diff
@@ -25,16 +25,37 @@ import (
...
+	Capacity        int64                  `json:"capacity"` // total blobber capacity
+	Used            int64                  `json:"used"`     // allocated capacity
+	LastHealthCheck common.Timestamp       `json:"last_health_check"`
+	PublicKey       string                 `json:"-" msg:"-"`
+	SavedData       int64                  `json:"saved_data"`
+	// StakePoolSettings used initially to create and setup stake pool.
+	StakePoolSettings stakepool.StakePoolSettings `json:"stake_pool_settings"`
+	TotalStake        int64                       `json:"total_stake"`
+	Information       Info                        `json:"info"`
+}
```
peterlimg, line 47: Embed the `StorageNode` would make code cleaner, see code:
```go
type storageNodeResponses struct {
  StorageNode
  TotalStake int64 `jon:"total_stake"`
}

File: code/go/0chain.net/smartcontract/storagesc/stakepool.go [hunk 6]
```diff
@@ -425,6 +425,13 @@ func (ssc *StorageSmartContract) stakePoolUnlock(
 			return "", common.NewErrorf("stake_pool_unlock_failed",
 				"saving stake pool: %v", err)
 		}
+		data, _ := json.Marshal(event.DbUpdates{
```
peterlimg, line 428: Please do not ignore the error as we don't know what would happen, and if error happens, we should avoid the empty `data` going into event db.

peterlimg (changes requested):

samtholiya:
was merged in another pr

Issue #1142, opened 2022-04-11
Partitions package refactoring required


peterlimg:
There are a few problems about the partitions package, which I suggested to fix it before in PR #963 [here](https://github.com/0chain/0chain/pull/963#pullrequestreview-880957636), anyway the main issues are summarized as below:
- It is an intrusive design for adding new partition types, which means each time when we need to add a new partition type, we will have to change the partitions code [here](https://github.com/0chain/0chain/blob/65f647a6450a04f4024a4c289535e56b8a49e474/code/go/0chain.net/smartcontract/partitions/random.go#L24-L31) and [here](https://github.com/0chain/0chain/blob/65f647a6450a04f4024a4c289535e56b8a49e474/code/go/0chain.net/smartcontract/partitions/random.go#L310-L322). We should avoid this kind of design when possible, as it makes it hard to maintain and extend. 

- Lots of copy&paste code when adding a new partition type, check [here](https://github.com/0chain/0chain/blob/65f647a6450a04f4024a4c289535e56b8a49e474/code/go/0chain.net/smartcontract/partitions/blobber_challenge.go) for how many methods we need to create `blobber_challenge` partitions. And compare it to [blobber_list](https://github.com/0chain/0chain/blob/65f647a6450a04f4024a4c289535e56b8a49e474/code/go/0chain.net/smartcontract/partitions/blobber_list.go) partition code, you will find most of the code are the same. I believe they are an easy 'copy and paste' work, which is sometimes the source of potential errors and hidden bugs. So basically it's too much trivial works to do if we want to add new partition types.

- Easy to use it wrong, just check how we get the blobber reward item from partition [here](https://github.com/0chain/0chain/blob/65f647a6450a04f4024a4c289535e56b8a49e474/code/go/0chain.net/smartcontract/storagesc/challenge.go#L612-L614), the `blobberRewardItem` is already decoded from bytes to `BlobberRewardNode` instance in msgpack way, why should we waste time to encode it to json bytes and decode it again to get the instance. This is the place where performance get worse.

- Also, there's a bug in [GetRandomSlice()](https://github.com/0chain/0chain/blob/65f647a6450a04f4024a4c289535e56b8a49e474/code/go/0chain.net/smartcontract/partitions/random.go#L210), so if the last partition is not full, we will try to fulfill the return list with the items in previous (index-1) partition, however, the code here would not work, the `itemRange` will always give `nil`, just check the `start` and `end` index, and what partition it is on. The correct way should be: 
[...]

PR #1143, opened 2022-04-11
Partition refactoring

peterlimg:
## Fixes
- #1142 

## Changes

TLDR:
- Make it easy to support new partitions
- Clean APIs
- Move partition data types to their own packages
- Fixed the `GetRandomSlice()` bug
- Performance improvements
- Check [README](https://github.com/0chain/0chain/tree/staging/code/go/0chain.net/smartcontract/partitions#readme) for the details about how to use the partition package.

### Purpose
The purpose of this PR was to make the partition easy to use and easy to support new partitions, check issue #1142 for details. 

After refactoring, the `partitions` package is a pure utility package, hence no business data structs defined there. 

Main APIs are:

```go
// Create partitions if not exists
func CreateIfNotExists(state state.StateContextI, name string, size int) (*Partitions, error)

// Get partitions of given name
func GetPartitions(state state.StateContextI, name string) (*Partitions, error)

// Get random items from given partitions, the items will be returned through the last parameter
func GetRandomItems(state state.StateContextI, name string, v interface{}) error

// Get item from given partition index and item id
func (p *Partitions) GetItem(state state.StateContextI, partIndex int, id string, v interface{}) error

// Update item in given partition
func (p *Partitions) UpdateItem(state state.StateContextI, partIndex int, item PartitionItem) error

// Save saves the partitions to MPT
func (p *Partitions) Save(state state.StateContextI) error
...
```

### Unexpected performance gain
Apart from the main purpose for making the package easy to use, the performance was also improved especially the `GetItem()` method. Previously if we want to get an item from partition, we will need to decode all the items in the partition and go through the item list to find it. But now, we don't need to decode every pieces of struct fields of each item, instead we only decode the `ID` field as an index, and the remaining data are kept as `[]byte`, so only when we find the `ID` is matched, we will decode the `[]byte` into concrete type. The new `GetItem()` methods is about 20~30x faster than before if the previous way code is written correctly(i.e, use interface assert instead calling `Decode(data.Encode())` to get item), or it would be even faster.
  -  The `GetRandomItems()` (previously as `GetRandomSlice()`) is slightly faster, but since we are using reflection partially to make this API clean, I didn't expect it could be faster than the previous version, but yeah the truths is, it is faster indeed.
  - The `UpdateItem()` is also faster than before apparently, as finding the item is faster.
  - I believe remove item would be faster as well, because we don't need to decode the full item list to find the item to remove. 

## Tests 

No


dabasov (approved):
lgtm

code/go/0chain.net/smartcontract/partitions/partitions.go
```diff
@@ -0,0 +1,111 @@
...
+
+	"0chain.net/core/util"
+
+	"0chain.net/chaincore/chain/state"
+)
+
+// APIs
+
+type Partitions struct {
+	rp *randomSelector
```
ssardana96: would rs be a better name here?

code/go/0chain.net/smartcontract/partitions/partitions.go
```diff
@@ -0,0 +1,111 @@
...
+
+type PartitionItem interface {
+	util.MPTSerializableSize
+	GetID() string
+}
+
+// CreateIfNotExists creates a partition if not exist
+// It's a common patten to call this to create partitions on
+// top-level partitions when project start
+func CreateIfNotExists(state state.StateContextI, name string, partitionSize int) (*Partitions, error) {
```
ssardana96, line 25: it can call GetPartitions

code/go/0chain.net/smartcontract/partitions/partitions.go
```diff
@@ -0,0 +1,111 @@
...
+// GetRandomItems returns items of partition size number from random partition,
+// if the last partition is not full, it will try to get and fill it with its partition
+// of index - 1.
+func (p *Partitions) GetRandomItems(state state.StateContextI, r *rand.Rand, v interface{}) error {
+	return p.rp.GetRandomItems(state, r, v)
+}
+
+type ChangePartitionCallback = func(string, []byte, int, int, state.StateContextI) error
+
+// Partitioner interface wraps the methods for processing a partition
```
ssardana96: we can remove this

code/go/0chain.net/smartcontract/partitions/random.go
```diff
@@ -67,34 +53,48 @@ func (rs *randomSelector) SetCallback(callback ChangePartitionCallback) {
...
-	item PartitionItem,
-	balances state.StateContextI,
-) (int, error) {
-	var part PartitionItemList
+func (rs *randomSelector) Add(state state.StateContextI, item PartitionItem) (int, error) {
+	var part *partition
 	var err error
 	if len(rs.Partitions) > 0 {
-		part, err = rs.getPartition(len(rs.Partitions)-1, balances)
+		part, err = rs.getPartition(state, len(rs.Partitions)-1)
```
ssardana96, line 60: state looks like a better name than balances.
I guess we should change everywhere to state

code/go/0chain.net/smartcontract/storagesc/blobber_reward_partition.go
```diff
@@ -0,0 +1,49 @@
+package storagesc
+
+import (
+	"strconv"
+	"strings"
+
+	c_state "0chain.net/chaincore/chain/state"
```
ssardana96, line 7: we should think of better name to distinguish them?
can chaincore/chain/state be moved to chaincore/state only?

ssardana96 (changes requested):

code/go/0chain.net/smartcontract/partitions/partitions.go
peterlimg: We have `GetPartitions()` function, and `Get*` function should not have `update` action

code/go/0chain.net/smartcontract/partitions/random.go
peterlimg: Yes, but the name is conflict with the `state` package, so perhaps we need another better name. Anyway, I'll keep it here and think about it later.

code/go/0chain.net/smartcontract/storagesc/blobber_reward_partition.go
peterlimg: They are two difference thing, but yes, we need to think of a better package name. We can do it later perhaps

ssardana08 (approved):

Issue #912, opened 2022-01-17
New reward table for event database


Sriep:
New reward table for event database. This will contain the tokens minted in `pay_me` transactions. 

We will want the 
- block number
- user wallet id
- pool id
- amount

Emit an event during `pay_me` transactions with the details. 

We want a REST API endpoint that will give the sum of pay_me transaction payments for a range of block numbers.

This issue is blocked by [Stake holder rewards pools and pay_rewards endpoint](https://github.com/0chain/0chain/issues/774)

ma2b0043:
@Sriep when can we expect to close this issue?

iamrz1:
Closed by https://github.com/0chain/0chain/pull/1175

Issue #1169, opened 2022-04-13
collect-reward command fails on sharder


iamrz1:
collect-reward command is not working anymore.
Seems to fail due to error in transaction-confirmation [here](https://github.com/0chain/0chain/blob/2573fbdfb3f2e188227e5b9485d06fcd3c47fd8f/code/go/0chain.net/sharder/transaction.go#L39). 

Need to find out why that happens, and how to fix it.

iamrz1:
`collect-reward` command seems to be working in miner-reward [PR](https://github.com/0chain/0chain/pull/1025). So hopefully fixed when the PR is merged.

iamrz1:
We need a system-test to cover this pr.

iamrz1:
Closed by https://github.com/0chain/0chain/pull/1186

PR #1175, opened 2022-04-15
EventsDB: New Rewards Table 

iamrz1:
## Fixes
https://github.com/0chain/0chain/issues/912

## Tests 

### DIFF
--- a/code/go/0chain.net/smartcontract/storagesc/handler.go
+++ b/code/go/0chain.net/smartcontract/storagesc/handler.go
@@ -970,3 +970,30 @@ func (ssc *StorageSmartContract) GetTotalData(_ context.Context, balances cstate
 
 	return 0, fmt.Errorf("storageSmartContract is nil")
 }
+
+func (ssc *StorageSmartContract) GetCollectedRewardHandler(ctx context.Context, params url.Values, balances cstate.StateContextI) (resp interface{}, err error) {
+	if balances.GetEventDB() == nil {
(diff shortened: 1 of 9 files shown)


code/go/0chain.net/smartcontract/storagesc/handler.go
```diff
@@ -880,3 +880,26 @@ func (ssc *StorageSmartContract) GetTotalData(_ context.Context, balances cstate
...
+func (ssc *StorageSmartContract) GetCollectedRewards(ctx context.Context, params url.Values, balances cstate.StateContextI) (resp interface{}, err error) {
+	var (
+		startBlock, _ = strconv.Atoi(params.Get("start_block"))
+		endBlock, _   = strconv.Atoi(params.Get("end_block"))
+	)
+
+	query := event.RewardQuery{
+		StartBlock: startBlock,
+		EndBlock:   endBlock,
+	}
```
peterlimg, line 989: This query will return all the rewards of given blocks range, we don't need to return rewards by client IDs?

iamrz1: yes,  we can return by client_id, by range of blocks, or both. I have kept options for other queries as well. Just let me know which one we need.
FYI, I cant test codes from this pr as it is blocked by https://github.com/0chain/0chain/issues/1169

peterlimg (approved):
lgtm

PR #1185, opened 2022-04-19
-  challenge partition changes (allocation remove blobber)

### LINKED ISSUE #1184: challenge partition update on remove blobber (storageAllocation)
- Blobbers removed from Allocation should be updated from blobberChallengeNode as well [here](https://github.com/0chain/0chain/blob/d559d33e634ea4b5da5f53d50e6c7cbb4ca7a9aa/code/go/0chain.net/smartcontract/storagesc/models.go#L832)

ssardana08:
…blobber

## Tests 

- blobber:
- gosdk:
- system_test:
- zboxcli:
- zwalletcli:
- Other: ...

### DIFF
--- a/code/go/0chain.net/smartcontract/partitions/partitions.go
+++ b/code/go/0chain.net/smartcontract/partitions/partitions.go
@@ -80,6 +80,11 @@ func (p *Partitions) Size(state state.StateContextI) (int, error) {
 	return p.rs.Size(state)
 }
 
+// Remove removes the partition item from given partIndex in Partitions
+func (p *Partitions) Remove(state state.StateContextI, partIndex int, item PartitionItem) error {
+	return p.rs.Remove(state, item, partIndex)
+}
+
 // GetRandomItems returns items of partition size number from random partition,
 // if the last partition is not full, it will try to get and fill it with its partition
 // of index - 1.
--- a/code/go/0chain.net/smartcontract/partitions/partitions.go
+++ b/code/go/0chain.net/smartcontract/partitions/partitions.go
@@ -80,9 +80,9 @@ func (p *Partitions) Size(state state.StateContextI) (int, error) {
 	return p.rs.Size(state)
 }
 
-// Remove removes the partition item from given partIndex in Partitions
-func (p *Partitions) Remove(state state.StateContextI, partIndex int, item PartitionItem) error {
-	return p.rs.Remove(state, item, partIndex)
+// RemoveItem removes the partition item from given partIndex in Partitions
+func (p *Partitions) RemoveItem(state state.StateContextI, partIndex int, id string) error {
+	return p.rs.RemoveItem(state, id, partIndex)
 }
 
 // GetRandomItems returns items of partition size number from random partition,
--- a/code/go/0chain.net/smartcontract/storagesc/models.go
+++ b/code/go/0chain.net/smartcontract/storagesc/models.go
@@ -864,50 +864,9 @@ func (sa *StorageAllocation) removeBlobber(
 			removedBlobber.Used -= d.Size
 
 			if d.ChallengePartitionLoc != nil {
-				blobberAllocChallPartition, err := getBlobbersChallengeAllocationList(removeId, balances)
-				if err != nil {
-					return nil, fmt.Errorf("cannot getch blobber allocation partition: %v", err)
-				}
-				err = blobberAllocChallPartition.Remove(balances, d.ChallengePartitionLoc.Location,
-					&BlobberChallengeAllocationNode{ID: sa.ID})
-				if err != nil {
-					return nil, fmt.Errorf("error removing allocation from challenge partition: %v", err)
-				}
-				sa.BlobberDetails[i].ChallengePartitionLoc = nil
-				if err = blobberAllocChallPartition.Save(balances); err != nil {
-					return nil, fmt.Errorf("error saving allocation challenge partition: %v", err)
-				}
-
-				blobberAllocChallSize, err := blobberAllocChallPartition.Size(balances)
-				if err != nil {
-					return nil, fmt.Errorf("error getting size of challenge partition: %v", err)
-				}
-				if blobberAllocChallSize == 0 {
-					bcPartitionLoc, err := ssc.getBlobberChallengePartitionLocation(removeId, balances)
-					if err != nil {
-						return nil, fmt.Errorf("error retrieving blobber challenge partition location: %v", err)
-					}
-
-					bcPartition, err := getBlobbersChallengeList(balances)
-					if err != nil {
-						return nil, fmt.Errorf("error retrieving blobber challenge partition: %v", err)
-					}
-
-					err = bcPartition.Remove(balances, bcPartitionLoc.PartitionLocation.Location,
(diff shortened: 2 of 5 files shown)


Sriep (approved):

code/go/0chain.net/smartcontract/partitions/partitions.go
```diff
@@ -80,6 +80,11 @@ func (p *Partitions) Size(state state.StateContextI) (int, error) {
 	return p.rs.Size(state)
 }
 
+// Remove removes the partition item from given partIndex in Partitions
+func (p *Partitions) Remove(state state.StateContextI, partIndex int, item PartitionItem) error {
```
peterlimg: Rename to `RemoveItem()` as this is for removing item from specific partition, and we'd better have the consistent APIs language, e.g. `GetItem()`, `UpdateItem`. The `Remove()` API may be used to remove specific partition in the future if needed.

We don't need`item` param, instead an `id string` is enough to locate the item.

code/go/0chain.net/smartcontract/storagesc/models.go
```diff
@@ -862,6 +862,54 @@ func (sa *StorageAllocation) removeBlobber(
...
+				blobberAllocChallPartition, err := getBlobbersChallengeAllocationList(removeId, balances)
+				if err != nil {
+					return nil, fmt.Errorf("cannot getch blobber allocation partition: %v", err)
+				}
+				err = blobberAllocChallPartition.Remove(balances, d.ChallengePartitionLoc.Location,
+					&BlobberChallengeAllocationNode{ID: sa.ID})
+				if err != nil {
+					return nil, fmt.Errorf("error removing allocation from challenge partition: %v", err)
+				}
+				sa.BlobberDetails[i].ChallengePartitionLoc = nil
```
peterlimg: sa.BlobberDetail[i] has already been removed and replaced with the last item, so change it will cause unexpected error.

code/go/0chain.net/smartcontract/storagesc/models.go
```diff
@@ -862,6 +862,54 @@ func (sa *StorageAllocation) removeBlobber(
...
+				if err = blobberAllocChallPartition.Save(balances); err != nil {
```
peterlimg: We'd better call the `Save()` at the end of all the action in case any error happen, in this way we can avoid an unnecessary MPT writing.

code/go/0chain.net/smartcontract/storagesc/models.go
```diff
@@ -862,6 +862,54 @@ func (sa *StorageAllocation) removeBlobber(
...
+						return nil, fmt.Errorf("error retrieving blobber challenge partition: %v", err)
+					}
+
+					err = bcPartition.Remove(balances, bcPartitionLoc.PartitionLocation.Location,
+						&BlobberChallengeNode{BlobberID: removeId})
+					if err != nil {
+						return nil, fmt.Errorf("error removing blobber from challenge partition: %v", err)
+					}
+
+					err = bcPartition.Save(balances)
```
peterlimg, line 902: Call the bcPartition.Save() out of the loop.

code/go/0chain.net/smartcontract/storagesc/models.go
```diff
@@ -862,6 +862,54 @@ func (sa *StorageAllocation) removeBlobber(
...
+				if err != nil {
+					return nil, fmt.Errorf("error getting size of challenge partition: %v", err)
+				}
+				if blobberAllocChallSize == 0 {
+					bcPartitionLoc, err := ssc.getBlobberChallengePartitionLocation(removeId, balances)
+					if err != nil {
+						return nil, fmt.Errorf("error retrieving blobber challenge partition location: %v", err)
+					}
+
+					bcPartition, err := getBlobbersChallengeList(balances)
```
peterlimg, line 891: Get the bcPartition out of the loop

code/go/0chain.net/smartcontract/storagesc/models.go
```diff
@@ -862,6 +862,54 @@ func (sa *StorageAllocation) removeBlobber(
 			sa.BlobberDetails[i] = sa.BlobberDetails[len(sa.BlobberDetails)-1]
 			sa.BlobberDetails = sa.BlobberDetails[:len(sa.BlobberDetails)-1]
 			removedBlobber.Used -= d.Size
+
+			if d.ChallengePartitionLoc != nil {
```
peterlimg, line 866: Move the code block in the for loop for removing a blobber into a separate function or method to keep single method short, this is for keeping the method easy to read and maintain.

Also, we'd better wraps those partitions related actions in blobber's own methods, and call them to handle the business logic.

peterlimg (changes requested):

code/go/0chain.net/smartcontract/storagesc/models.go
ssardana08: ahh yes! great catch @peterlimg

code/go/0chain.net/smartcontract/storagesc/models.go
ssardana08: if we don't save here, then Size() in next line will not return right size

ssardana08: I've made changes, please check if that looks fine

code/go/0chain.net/smartcontract/storagesc/models.go
```diff
@@ -935,6 +942,58 @@ func (sa *StorageAllocation) changeBlobbers(
...
+	err = blobberAllocChallPartition.RemoveItem(balances, allocPartitionIndex, allocID)
+	if err != nil {
+		return fmt.Errorf("error removing allocation from challenge partition: %v", err)
+	}
+
+	blobberAllocChallSize, err := blobberAllocChallPartition.Size(balances)
+	if err != nil {
+		return fmt.Errorf("error getting size of challenge partition: %v", err)
+	}
+	if blobberAllocChallSize-1 <= 0 {
```
peterlimg: The blobberAllocChallSize is the total item number in the partitions, so if all have been removed, it should be value of 0, not 1.

code/go/0chain.net/smartcontract/storagesc/models.go
```diff
@@ -935,6 +942,58 @@ func (sa *StorageAllocation) changeBlobbers(
 	return blobbers, nil
 }
 
+func removeBlobberAllocationFromChallPartition(
```
peterlimg: Rename to removeBlobberAllocation() would be easier to read.

peterlimg (changes requested):

code/go/0chain.net/smartcontract/storagesc/models.go
ssardana08: If we're doing Save at the end. It will be 1.. 
as it gets data from MPT and does size.
Since our partition is not saved yet. The size will be same as old one..

ssardana08: But in case if its 0 it should be 0. Yea you're right

peterlimg (approved):
lgtm

PR #1219, opened 2022-04-30
Updated configs to helmchart version

dabasov:
## Changes
Merged helm charts version
Adjusted costs to loadtest results
## Tests 

- blobber:
- gosdk:
- system_test:
- zboxcli:
- zwalletcli:
- Other: ...

### DIFF
--- a/docker.local/config/0chain.yaml
+++ b/docker.local/config/0chain.yaml
@@ -1,11 +1,9 @@
 version: 1.0
-
 logging:
   level: "debug"
-  console: true # printing log to console is only supported in development mode
+  console: false # printing log to console is only supported in development mode
   goroutines: false
   memlog: false
-
 development:
   state: true
   dkg: true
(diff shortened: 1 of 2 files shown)


service-0chain (approved):

docker.local/config/sc.yaml
```diff
@@ -199,13 +197,16 @@ smart_contracts:
...
+      blobber_usage_ratio: 40
+    expose_mpt: true
     cost:
       update_settings: 100
       read_redeem: 100
       commit_connection: 100
       new_allocation_request: 3000
       update_allocation_request: 2500
-      finalize_allocation: 100
+      finalize_allocation: 9500
```
peterlimg, line 209: Any reason to change the value?

docker.local/config/0chain.yaml
```diff
@@ -16,26 +14,29 @@ development:
...
     miner: true
     multisig: true
     vesting: true
+    magma: false #set true to enable magma
     zcn: true
   txn_generation:
     wallets: 50
+    transactions: 70
+    pour_point: 100
+    pour_amount: 10000
```
peterlimg: Are these new added configs being used? Perhaps we should do these changes along with the code that use them.

docker.local/config/0chain.yaml
```diff
@@ -48,32 +49,33 @@ server_chain:
     max_block_size: 10
     max_block_cost: 10000 #equal to 100ms
     max_byte_size: 1638400
+    generators: 2
```
peterlimg: What does the `generators` do, we are getting generators number base on the `generators_percent` and if it is less than the `min_generators`, we will use the `min_generators`.

peterlimg (changes requested):

docker.local/config/0chain.yaml
dabasov: it is not used, removed

docker.local/config/0chain.yaml
dabasov: the same

docker.local/config/sc.yaml
```diff
@@ -218,7 +219,7 @@ smart_contracts:
       curator_transfer_allocation: 100
       challenge_request: 100
       challenge_response: 1600
-      generate_challenges: 9999 #exceeds limit, might be switched off
+      generate_challenges: 0
```
peterlimg, line 222: Generate challenge should have been optimized, so perhaps we ca use 100 for it now.

docker.local/config/sc.yaml
peterlimg, line 209: Is finalize_allocation sc optimized? if so we should not use 9500 perhaps

peterlimg (changes requested):

PR #1250, opened 2022-05-07
Feature/remove all blobbers

### LINKED ISSUE #1115: Remove ALL_BLOBBER_KEY
- ALL_BLOBBER_KEY should be removed as any MPT key acces
- Blobbers data should be accessed by individual blobber only
- if needed, by partition logic

dabasov:
## Changes
Removed ALL_BLOBBERS_KEY from MPT

## Tests 


code/go/0chain.net/smartcontract/storagesc/handler.go
```diff
@@ -197,11 +183,12 @@ func (ssc *StorageSmartContract) GetBlobbersHandler(
 	params url.Values, balances cstate.StateContextI,
 ) (interface{}, error) {
 	if balances.GetEventDB() == nil {
-		return ssc.GetBlobbersHandlerDeprecated(ctx, params, balances)
+		return nil, common.NewErrInternal("events db is not initialised")
+		//return nil, errors.New("events db is not initialised")
```
peterlimg: Remove commented code

code/go/0chain.net/smartcontract/storagesc/handler.go
```diff
@@ -562,32 +549,26 @@ func (ssc *StorageSmartContract) GetAllocationMinLockHandler(ctx context.Context
...
-	var sa = request.storageAllocation()
+	sa := req.storageAllocation()
+	var gbSize = sizeInGB(sa.bSize())
+	var minLockDemand state.Balance
 
-	blobberNodes, bSize, err := ssc.selectBlobbers(
-		creationDate, *allBlobbersList, sa, creationDate.Unix(), balances)
-	if err != nil {
-		return "", common.NewErrInternal("selecting blobbers", err.Error())
+	ids := append(req.Blobbers, blobbers...)
```
peterlimg, line 564: Append the blobber ids to req.Blobbers? Duplicate ids could happen if the blobbers contains IDs that are in req.Blobbers too.

peterlimg (changes requested):

code/go/0chain.net/smartcontract/storagesc/handler.go
```diff
@@ -563,11 +562,19 @@ func (ssc *StorageSmartContract) GetAllocationMinLockHandler(ctx context.Context
 	var minLockDemand state.Balance
 
 	ids := append(req.Blobbers, blobbers...)
-	if len(ids) > req.ParityShards+req.DataShards {
-		ids = ids[:req.ParityShards+req.DataShards]
+	uniqueMap := make(map[string]bool)
```
peterlimg: Instead of using `bool` as map value, I suggest to use `struct{}`, also init it with capacity

peterlimg (changes requested):

code/go/0chain.net/smartcontract/storagesc/blobber.go
```diff
@@ -75,13 +74,27 @@ func (sc *StorageSmartContract) updateBlobber(t *transaction.Transaction,
 		return fmt.Errorf("can't get or decode saved blobber: %v", err)
 	}
 
+	if savedBlobber.BaseURL != blobber.BaseURL {
```
Sriep, line 72: In line 72
```go
savedBlobber, err := sc.getBlobber(blobber.ID, balances)
```
Is this MPT access needed? I think blobber `blobber.Id` might already have been retrieved [further up the stack](https://github.com/0chain/0chain/blob/feature/remove-all-blobbers/code/go/0chain.net/smartcontract/storagesc/blobber.go#L835), so could we pass this in as a parameter?

peterlimg: Yes, the blobber should have been acquired, we don't need to get from MPT again.

code/go/0chain.net/smartcontract/storagesc/handler.go
dabasov: @peterlimg what's the difference between bool and struct{}? I suppose they both take 1 WORD size?

Sriep: https://itnext.io/set-in-go-map-bool-and-map-struct-performance-comparison-5315b4b107b
```
Although map[]struct{} is faster and consumes less memory,
it is more obvious when it comes to a big Set. 
So if you are in a hurry and the resource is not your top concern,
using map[]bool should be totally fine. 
```

peterlimg: @dabasov `struct{}{}` doesn't take up any space, while `bool` does, https://dave.cheney.net/2014/03/25/the-empty-struct. We actually don't use the map value, so we don't need to waste resources on unused field.

dabasov: thanks guys

peterlimg (approved):
LGTM 👍 

Please fix the lint errors

code/go/0chain.net/smartcontract/storagesc/handler.go
```diff
@@ -1053,20 +1041,6 @@ func (ssc *StorageSmartContract) GetBlocksHandler(_ context.Context, params url.
...
-
-		var totalSavedData int64
-		for _, sn := range storageNodes.Nodes {
-			totalSavedData += sn.SavedData
-		}
-
-		return totalSavedData, nil
-	}
-
 	return 0, fmt.Errorf("storageSmartContract is nil")
```
Sriep: This orphaned error message does not make sense. Maybe something like "total_saved_data is not implemented yet".

PR #1251, opened 2022-05-08
Updated gitactions for running conductor tests.

service-0chain:
## Tests

- blobber:
- gosdk:
- system_test:
- zboxcli:
- zwalletcli:
- Other: ...


.github/workflows/conductor-vc-tests.yml
```diff
@@ -1,4 +1,4 @@
-name: CONDUCTOR_TEST
+name: Conductor VC Tests
```
rrrooommmaaa: Please don't use the name "VC tests", that's misleading.
We have to run 4 groups of tests:

no-view-change.byzantine (with view_change: false)
no-view-change.fault-tolerance (with view_change: false)
view-change.byzantine (with view_change: true)
view-change.fault-tolerance (with view_change: true)

There is no such file `view-change-3` -- so this job can't be run.
Let's rename the group of these 4 jobs to `Complex Scenarios` or something?
I'll update README.md file accordingly
and also I'll need to update conductor code to be able to run a group of scripts by mask,
e.g. `view-change.fault-tolerance.*` because we have tests there split into a dozen of files (running all of them consequently would take 2-3 hours) and running `start.conductor.ch view-change.fault-tolerance` would only run a single script `conductor.view-change.fault-tolerance.yaml` (only a portion of fault-tolerance scripts).
So we can later spread the remaining `view-change.fault-tolerance.*` tests across 2 more machines, for example.

rrrooommmaaa (dismissed):

.github/workflows/conductor-vc-tests.yml
service-0chain: Re: view-change-3
It was mentioned in your doc. So, I added it into gitactions also.
https://github.com/0chain/0chain/blob/d9a3398e8a6c78344621c9487a10e64571233d6c/code/go/0chain.net/conductor/README.md#running-view-change-tests

vinayakgautamops (approved):

PR #1253, opened 2022-05-08
Sample vscode config for debugging

rrrooommmaaa:
If someone else uses VSCode for debugging, this may be helpful


peterlimg:
We don't save the IDE settings in the project, so I'll close this one. Anyway, good to check this if someone use vscode for debugging

peterlimg:
We don't save the IDE settings in the project, so I'll close this one. Anyway, good to check this if someone use vscode for debugging

PR #1256, opened 2022-05-09
Move to global config

### LINKED ISSUE #1029: Move from global config to MPT based Config
The following settings are to be kept in the MPT. They need a `update_global_settings` transaction to change.

The indicated settings [following global settings](https://github.com/0chain/0chain/blob/staging/code/go/0chain.net/smartcontract/minersc/globals.go#L26). These are either stored in the [global development object](https://github.com/0chain/0chain/blob/staging/code/go/0chain.net/chaincore/config/dev_config.go#L5), in the chain or other ad-hock global variables.

We need to move the vales from their current location to the chain's [Config object](https://github.com/0chain/0chain/blob/staging/code/go/0chain.net/chaincore/chain/config.go#L53).

```go
	State         // todo from development
	Dkg                               // todo from development
	ViewChange                        // todo from development
	BlockRewards                      // todo from development
	Storage                           // todo from development
	Faucet                            // todo from development
	Interest                          // todo from development
	Miner                             // todo from development
	Multisig                          // todo from development
	Vesting // todo from development
	Owner   // probably don't want this changeable
	BlockGenerationTimout        // todo from chain
	BlockGenerationRetryWaitTime // todo from chain
	RoundTimeoutsTimeoutCap // todo from chain
	TransactionTimeout // todo from global
	TransactionMinFee  // todo from global
	ClientDiscover // todo from chain
	StateSyncTimeout   // todo from chain
	StuckCheckInterval // todo from chain
	StuckTimeThreshold // todo from chain
	LfbTicketAhead                           // todo from chain

```

The following entries should not be held in the MPT but read from the `0chain.yaml` file
```  
dbs:
    events:
      enabled: true
      name: events_db
      user: zchain_user
      password: zchian
      host: postgres #localhost
      port: 5432
      max_idle_conns: 100
      max_open_conns: 200
      conn_max_lifetime: 20s
```

https://github.com/0chain/0chain/issues/736 also updates settings with new security levels.

mohsenno1:
Resolves https://github.com/0chain/0chain/issues/1029

## Tests 
- [x]  Merge your system tests PR to master AFTER merging this PR

-actions: https://github.com/0chain/actions/pull/44
- blobber:
- gosdk:
- system_test:
- zboxcli:
- zwalletcli:
- Other: ...


service-0chain:
| config  | value |
| tests skipped  | FALSE  |
| system-tests  | master  |
| gosdk  | NOT OVERRIDDEN  |
| zboxcli  | staging  |
| zwalletcli  | staging  |
| 0chain  | move-to-global-config  |
| blobber  | staging  |
| authorizer  | staging  |
| 0box  | staging  |
| 0block  | staging  |
| 0dns  | staging  |
| explorer  | master  |
| 0proxy  | staging  |
| 0search  | staging  |

code/go/0chain.net/chaincore/chain/config.go
```diff
@@ -95,6 +110,93 @@ func NewConfigImpl(conf *ConfigData) *ConfigImpl {
 	return &ConfigImpl{conf: conf}
 }
 
+// This is to provide access to chain configuration from packages which don't have
+// direct access to 'chain' package. Instead they can read the configuration using
+// 'conf' package. example: conf.Configuration.ReadValue("BlockSize")
+func (c *ConfigImpl) ReadValue(name string) (interface{}, error) {
```
peterlimg: I'm confused about the purpose of this `ReadValue()` method. Why do we need this instead of returning the config directly to users. The calling of this `ReadValue` also confused me, such as `isFeeEnabled, err := config.Configuration().ChainConfig.ReadValue("Miner")`, what's the relationship between the value "Miner" to the `isFeeEnabled`. We need to design APIs that is easy to use and read.

code/go/0chain.net/chaincore/chain/config.go
```diff
@@ -38,6 +40,18 @@ const (
...
+	ViewChange() bool
+	BlockRewards() bool
+	Storage() bool
+	Faucet() bool
+	Interest() bool
+	// Indicates is fees enabled
+	Miner() bool
+	Multisig() bool
+	Vesting() bool
+	Zcn() bool
```
peterlimg: Rename the methods that return bool value to `Is*Enabled()` or `Enable*` such as `IsVierChangeEnabled()` `EnableViewChange()` so that users could get what it is for.

code/go/0chain.net/chaincore/chain/entity.go
```diff
@@ -418,6 +415,10 @@ func NewChainFromConfig() *Chain {
 func Provider() datastore.Entity {
 	c := &Chain{}
 	c.Config = NewConfigImpl(&ConfigData{})
+	config.Configuration().ChainConfig = c.Config
+
+	c.Config.FromViper()
```
peterlimg: Run this before `config.Configuration().ChainConfig = c.Config`

code/go/0chain.net/chaincore/chain/state.go
```diff
@@ -313,7 +312,7 @@ func (c *Chain) updateState(ctx context.Context, b *block.Block, bState util.Mer
 		return nil, fmt.Errorf("invalid transaction type: %v", txn.TransactionType)
 	}
 
-	if config.DevConfiguration.IsFeeEnabled {
+	if c.Config.Miner() {
```
peterlimg: Rename to `IsFeeEnabled()` , it does not matter if this is a miner or sharder, if fee is enabled, we need to run the following transaction on both miner and sharder.

code/go/0chain.net/chaincore/config/config.go
```diff
@@ -135,30 +137,69 @@ const (
 	DeploymentMainNet     = 2
 )
 
+type ConfigReader interface {
+	ReadValue(name string) (interface{}, error)
+}
+
+// This is defined to avoid 'nil pointer reference' to ConfigChain
+// And also to be used for setting the configuration for Test purposes
+type TestConfigReader struct {
```
peterlimg: We should put test code or related helper file in _test.go file only.

code/go/0chain.net/chaincore/config/config.go
```diff
@@ -135,30 +137,69 @@ const (
...
+	ChainConfig    ConfigReader
 }
 
 /*Configuration of the system */
-var Configuration Config
+var configuration Config
+
+func Configuration() *Config {
+	if configuration.ChainConfig == nil {
+		configuration.ChainConfig = &TestConfigReader{
```
peterlimg: We should not use Test* code here, if it could be nil, initialize it explicitly in top main before it is used.

code/go/0chain.net/chaincore/smartcontract/handler.go
```diff
@@ -87,5 +88,8 @@ func ExecuteSmartContract(t *transaction.Transaction, scData *sci.SmartContractT
 
 func EstimateTransactionCost(t *transaction.Transaction, scData sci.SmartContractTransactionData, balances c_state.StateContextI) (int, error) {
 	contractObj := getSmartContract(t.ToClientID)
+	if contractObj == nil {
+		return 0, errors.New("EstimateTransactionCost - invalid Client ID")
```
peterlimg: We should not return capitalized error message. So I suggested to return `errors.New("invalid to client id") , the EstimateTransactionCost function caller shall take the responsibility to add more error info.

code/go/0chain.net/chaincore/wallet/wallet2localblockchain.go
```diff
@@ -43,7 +45,12 @@ func (w *Wallet) CreateSendTransaction(toClient string, value int64, msg string,
 	txn.ToClientID = toClient
 	txn.Value = value
 	txn.TransactionData = msg
-	if config.DevConfiguration.IsFeeEnabled {
+
+	isFeeEnabled, err := config.Configuration().ChainConfig.ReadValue("Miner")
+	if err != nil {
+		logging.Logger.Error("Wallet - CreateSendTransaction - Cannot read chain configuration", zap.Any("error", err))
```
peterlimg: Return if error happens or panic.

The error can be logged in the following way:
`logging.Logger.Error("...", zap.Error(err))`

code/go/0chain.net/chaincore/wallet/wallet2localblockchain.go
```diff
@@ -43,7 +45,12 @@ func (w *Wallet) CreateSendTransaction(toClient string, value int64, msg string,
...
+	}
+	if isFeeEnabled == true {
```
peterlimg: Do `if isFeeEnabled {`

code/go/0chain.net/chaincore/wallet/wallet2localblockchain.go
```diff
@@ -59,7 +66,12 @@ func (w *Wallet) CreateSCTransaction(toClient string, value int64, msg string, f
 	txn.ToClientID = toClient
 	txn.Value = value
 	txn.TransactionData = msg
-	if config.DevConfiguration.IsFeeEnabled {
+
+	isFeeEnabled, err := config.Configuration().ChainConfig.ReadValue("Miner")
+	if err != nil {
+		logging.Logger.Error("Wallet - CreateSCTransaction - Cannot read chain configuration", zap.Any("error", err))
```
peterlimg: Same as above

code/go/0chain.net/chaincore/wallet/wallet2localblockchain.go
```diff
@@ -59,7 +66,12 @@ func (w *Wallet) CreateSCTransaction(toClient string, value int64, msg string, f
...
+	}
+	if isFeeEnabled == true {
```
peterlimg: Same as above

code/go/0chain.net/chaincore/wallet/wallet2localblockchain.go
```diff
@@ -81,7 +93,12 @@ func (w *Wallet) CreateDataTransaction(msg string, fee int64) *transaction.Trans
 	txn.ClientID = w.ClientID
 	txn.TransactionData = msg
 	txn.TransactionType = transaction.TxnTypeData
-	if config.DevConfiguration.IsFeeEnabled {
+
+	isFeeEnabled, err := config.Configuration().ChainConfig.ReadValue("Miner")
+	if err != nil {
+		logging.Logger.Error("Wallet - CreateDataTransaction - Cannot read chain configuration", zap.Any("error", err))
+	}
+	if isFeeEnabled == true {
```
peterlimg: Same issue as above

code/go/0chain.net/smartcontract/minersc/dkg.go
```diff
@@ -224,8 +224,13 @@ func GetPhaseNode(statectx cstate.StateContextI) (
 
 func (msc *MinerSmartContract) setPhaseNode(balances cstate.StateContextI,
 	pn *PhaseNode, gn *GlobalNode, t *transaction.Transaction) error {
+
+	isViewChange, err := config.Configuration().ChainConfig.ReadValue("ViewChange")
+	if err != nil {
+		logging.Logger.Error("MinerSmartContract - setPhaseNode - Cannot read chain configuration", zap.Any("error", err))
```
peterlimg: Return if error happens

code/go/0chain.net/smartcontract/minersc/dkg.go
```diff
@@ -224,8 +224,13 @@ func GetPhaseNode(statectx cstate.StateContextI) (
 
 func (msc *MinerSmartContract) setPhaseNode(balances cstate.StateContextI,
 	pn *PhaseNode, gn *GlobalNode, t *transaction.Transaction) error {
+
+	isViewChange, err := config.Configuration().ChainConfig.ReadValue("ViewChange")
```
peterlimg: Add `Config()` method to the `cstate.StateContextI` to get config from chain struct, or add the config access function to the `MinerSmartContract` struct. For example, 

```go
type GetConfigFunc func() *config.Config

type MinerSmartContract struct {
    getConfig smartcontract.GetConfigFunc
}

Init the MinerSmartContract with the function to get config from Chain when the smart contract is called. Then in the `setPhaseNode` method, we can run `msc.getConfig()` to get the config directly. 

```


Do this to all other smart contracts that need to access the config.

code/go/0chain.net/smartcontract/minersc/dkg.go
```diff
@@ -224,8 +224,13 @@ func GetPhaseNode(statectx cstate.StateContextI) (
...
+	if err != nil {
+		logging.Logger.Error("MinerSmartContract - setPhaseNode - Cannot read chain configuration", zap.Any("error", err))
+	}
 	// move phase condition
-	var movePhase = config.DevConfiguration.ViewChange &&
+	var movePhase = isViewChange == true &&
```
peterlimg: `isViewChange` is enough, we don't need to do `isViewChange == true`.

code/go/0chain.net/smartcontract/minersc/fees.go
```diff
@@ -320,7 +321,12 @@ func (msc *MinerSmartContract) payFees(t *transaction.Transaction,
 	_ []byte, gn *GlobalNode, balances cstate.StateContextI) (
 	resp string, err error) {
 
-	if config.DevConfiguration.ViewChange {
+	isViewChange, err := config.Configuration().ChainConfig.ReadValue("ViewChange")
+	if err != nil {
```
peterlimg: Get config from `msc` as suggested before

peterlimg (changes requested):

code/go/0chain.net/chaincore/chain/config.go
mohsenno1: The value of `Miner` had been already checked to examine if fee is enabled before I touch it. It was strange to me too but this was there before I change it.
https://github.com/0chain/0chain/blob/15b1511a03e10fa38c8a55fd4a12ab71e627f2b4/code/go/0chain.net/chaincore/config/dev_config.go#L27
About the ConfigReader interface:
The problem is we need to access the chain configuration from many packages. To read this config we need to add a reference to 'chain' package. But we cannot do that. Since we are reading the chain config from packages which already added as a reference to chain package (either directly or indirectly), so adding chain as a reference to these packages creates a `circular dependency issue`. 

There was different workaround to solve this problem. Best one was to move the chain configuration to a package accessible from all (or most) of the packages. However this demands a lot of changes and I guess we don't want to do this at the time (We are close to Mainnet).

So I solved the problem by adding a ConfigReader Interface in `chaincore/config` package. 

https://github.com/0chain/0chain/blob/df022bb2d701af666a428fe0c87a300f598c41f1/code/go/0chain.net/chaincore/config/config.go#L140-L142

then I changed the ConfigImpl to implement this interface. And I defined a ChainConfig field to be of type ConfigReader interface in Config struct in `chaincore/config` package to hold the ConfImpl instance. 

https://github.com/0chain/0chain/blob/df022bb2d701af666a428fe0c87a300f598c41f1/code/go/0chain.net/chaincore/config/config.go#L170-L176

And I assign it at the time of creating a chain instance. 

https://github.com/0chain/0chain/blob/62b291f11033fed01736b60015a81ced7a792f7d/code/go/0chain.net/chaincore/chain/entity.go#L416-L420

Using this approach we can access Chain Configuration by accessing to  `chaincore/config` package. 
example: 
https://github.com/0chain/0chain/blob/62b291f11033fed01736b60015a81ced7a792f7d/code/go/0chain.net/chaincore/wallet/wallet2localblockchain.go#L49

code/go/0chain.net/chaincore/chain/entity.go
mohsenno1: I don't think this makes any difference, but I can change it.

code/go/0chain.net/chaincore/chain/state.go
mohsenno1: This is the current version. 
https://github.com/0chain/0chain/blob/15b1511a03e10fa38c8a55fd4a12ab71e627f2b4/code/go/0chain.net/chaincore/config/dev_config.go#L27
As you can see it's setting the IsFeeEnabled  from `"development.smart_contract.miner" `setting, which now has been changed to:
https://github.com/0chain/0chain/blob/89884cb0fdf825d04df1c1d6967b633512c8baf4/code/go/0chain.net/chaincore/chain/config.go#L521-L522


I was not sure should I rename it or not. So I left it like this to get other people opinion in review. Now if you think better to rename this then we have to also rename it in 0chain.yaml for consistency.

code/go/0chain.net/chaincore/config/config.go
mohsenno1: removed and the chain is working fine.

code/go/0chain.net/chaincore/wallet/wallet2localblockchain.go
mohsenno1: And I guess here as well we don't want capitalized error message.

code/go/0chain.net/chaincore/wallet/wallet2localblockchain.go
mohsenno1: I think the way I did it is correct, since the isFeeEnabled is an interface and not a boolean. We have the Type at runtime.

code/go/0chain.net/chaincore/wallet/wallet2localblockchain.go
mohsenno1: same as above

code/go/0chain.net/chaincore/chain/config.go
mohsenno1: What about their corresponding values in 0chain.yaml and viper. Do they need to be changed or I leave it as before:

https://github.com/0chain/0chain/blob/1b24b8d460cd4d281ad20fc357fee7b6b5d11ba2/docker.local/config/0chain.yaml#L101-L107

https://github.com/0chain/0chain/blob/89884cb0fdf825d04df1c1d6967b633512c8baf4/code/go/0chain.net/chaincore/chain/config.go#L514-L524

I can do them like this:
https://github.com/0chain/0chain/blob/1b24b8d460cd4d281ad20fc357fee7b6b5d11ba2/docker.local/config/0chain.yaml#L89-L93

code/go/0chain.net/chaincore/config/config.go
mohsenno1: removed

code/go/0chain.net/smartcontract/minersc/dkg.go
mohsenno1: isViewChange is an interface. So we compare like this.

code/go/0chain.net/smartcontract/minersc/dkg.go
mohsenno1: I didn't realize what are you trying to achieve by this? If you want to access the chain config implementation here and access its functions directly, I'm afraid it's not possible. The `chaincore/chain` package is dependent to `minersc` package:
https://github.com/0chain/0chain/blob/89884cb0fdf825d04df1c1d6967b633512c8baf4/code/go/0chain.net/chaincore/chain/config.go#L15-L16
This looks like a bad design to me and core packages must not be depended to the higher level packages, but this is the case now. So by applying your suggested changes accessing config will be like:
`msc.GetConfig().ChainConfig.ReadValue("ViewChange")` which is not much different than my code. But if you are suggesting something else, then consider that we cannot have access to ConfigImpl struct and its methods from minersc package and let me know what is your solution for it. https://github.com/0chain/0chain/blob/89884cb0fdf825d04df1c1d6967b633512c8baf4/code/go/0chain.net/chaincore/chain/config.go#L92-L95

code/go/0chain.net/smartcontract/minersc/fees.go
mohsenno1: as above

code/go/0chain.net/chaincore/chain/config.go
peterlimg: Rename the function names, but leave the yaml file unchanged, otherwise we will need to update deployment scripts to adjust the yaml file changes.

code/go/0chain.net/chaincore/chain/state.go
peterlimg: Yes, please rename to IsFeeEnabled(), but keep the yaml file unchanged.

code/go/0chain.net/smartcontract/minersc/dkg.go
peterlimg: Yes, the dependence circle issue could happen as the `Config` interface is defined in `chain` package, let's move it and the `ConfigImpl` all to `chaincore/config` package. Then have `chain` struct to embed the `config.Config` interace. One thing to be noted is the `Update()` method: https://github.com/0chain/0chain/blob/89884cb0fdf825d04df1c1d6967b633512c8baf4/code/go/0chain.net/chaincore/chain/config.go#L86
It uses the `minersc.GlobalSettings`, which don't have to be in `minersc` package, so we can move it to `chain/config` package too. In this way, both `chain` and smart contracts can access the config directly, and there won't be too much of changes.

code/go/0chain.net/smartcontract/storagesc/blobber.go
```diff
@@ -1,11 +1,12 @@
 package storagesc
 
 import (
-	"0chain.net/core/datastore"
 	"encoding/json"
 	"errors"
 	"fmt"
 
+	"0chain.net/core/datastore"
```
Sriep: no change, revert .

code/go/0chain.net/smartcontract/storagesc/benchmark_tests.go
```diff
@@ -1,17 +1,18 @@
...
-	"0chain.net/smartcontract/stakepool"
-	"0chain.net/smartcontract/stakepool/spenum"
 	"encoding/hex"
 	"encoding/json"
 	"fmt"
 	"log"
 	"testing"
 	"time"
 
+	"0chain.net/chaincore/state"
```
Sriep: no change, revert

code/go/0chain.net/smartcontract/minersc/sharder.go
```diff
@@ -153,7 +153,7 @@ func (msc *MinerSmartContract) AddSharder(
 	}
 
 	// save all sharders list
-	if err = updateAllShardersList(balances, allSharders); err != nil {
+	if err = UpdateAllShardersList(balances, allSharders); err != nil {
```
Sriep: UpdateAllShardersList is only used in minersc, there is no reason to export it.

code/go/0chain.net/smartcontract/minersc/models.go
```diff
@@ -910,7 +910,7 @@ func getMinersList(state cstate.StateContextI) (*MinerNodes, error) {
 	return minerNodes, nil
 }
 
-func updateMinersList(state cstate.StateContextI, miners *MinerNodes) error {
+func UpdateMinersList(state cstate.StateContextI, miners *MinerNodes) error {
```
Sriep: UpdateMinersList is only used in minersc, there is no reason to export it. Also, many other functions have been changed to be explored for no apparent reason. For example UpdateDKGMinersList, UpdateAllShardersList ...

code/go/0chain.net/smartcontract/minersc/fees.go
```diff
@@ -274,19 +274,20 @@ func (msc *MinerSmartContract) adjustViewChange(gn *GlobalNode,
...
+	if err := UpdateDKGMinersList(balances, dmn); err != nil {
 		return common.NewErrorf("adjust_view_change",
 			"can't cleanup DKG miners: %v", err)
 	}
 
 	return
 }
 
-func (msc *MinerSmartContract) payFees(t *transaction.Transaction,
+func (msc *MinerSmartContract) PayFees(t *transaction.Transaction,
```
Sriep: PayFees is only used in minersc, and does not need to be made public.

code/go/0chain.net/smartcontract/minersc/fees2_test.go
```diff
@@ -343,35 +345,36 @@ func testPayFees(t *testing.T, minerStakes []float64, sharderStakes [][]float64,
...
+	var allSharders = &minersc.MinerNodes{
 		Nodes: sharders,
 	}
-	err = updateAllShardersList(ctx, allSharders)
+	err = minersc.UpdateAllShardersList(ctx, allSharders)
 	require.NoError(t, err)
 
 	// Add information only relevant to view change rounds
-	config.DevConfiguration.ViewChange = zChainYaml.viewChange
+	config.Configuration().ChainConfig = chain.NewConfigImpl(&chain.ConfigData{IsViewChangeEnabled: true})
```
Sriep: So this is the reason for freely exporting methods in minersc. _Unit tests should not affect production code._ We should find another way to handle this. 

I suggest using a mock object;  `0chain/chiancore/mocks.ChainConfig`. Then you can use code similar to
```go
mockChainConfig := mocks.ChainConfig{}
...
mockChainConfig.On("IsViewChangeEnabled").Return(true)
...
config.Configuration().ChainConfig = mockChainConfig
```
To generate mocks use
```bash
./generate_mocks.sh
```
from 0chain root.

Sriep (commented):
Don't export functions just for the sake of unit tests. Find another way. Try mocking out the chaincore/config.ChainConfig object.

code/go/0chain.net/smartcontract/minersc/fees2_test.go
Sriep: To generate mocks use
```bash
./generate_mocks.sh
```
from 0chain root.

dabasov (approved):
lgtm, don't forget to update settings files in helm chart after the merge

code/go/0chain.net/smartcontract/minersc/models.go
```diff
@@ -956,8 +956,8 @@ func getDKGMinersList(state cstate.CommonStateContextI) (*DKGMinerNodes, error)
 	return dkgMiners, nil
 }
 
-// updateDKGMinersList update the dkg miners list
-func updateDKGMinersList(state cstate.StateContextI, dkgMiners *DKGMinerNodes) error {
+// UpdateDKGMinersList update the dkg miners list
+func UpdateDKGMinersList(state cstate.StateContextI, dkgMiners *DKGMinerNodes) error {
```
peterlimg: Keep it private if it is not used outside

peterlimg (changes requested):

peterlimg (approved):
lgtm

service-0chain:
| config  | value |
| tests skipped  | FALSE  |
| system-tests  | master  |
| gosdk  | NOT OVERRIDDEN  |
| zboxcli  | staging  |
| zwalletcli  | staging  |
| 0chain  | staging  |
| blobber  | staging  |
| authorizer  | staging  |
| 0box  | staging  |
| 0block  | staging  |
| 0dns  | staging  |
| explorer  | master  |
| 0proxy  | staging  |
| 0search  | staging  |

service-0chain:
| config  | value |
| tests skipped  | FALSE  |
| system-tests  | master  |
| gosdk  | NOT OVERRIDDEN  |
| zboxcli  | staging  |
| zwalletcli  | staging  |
| 0chain  | staging  |
| blobber  | staging  |
| authorizer  | staging  |
| 0box  | staging  |
| 0block  | staging  |
| 0dns  | staging  |
| explorer  | master  |
| 0proxy  | staging  |
| 0search  | staging  |

service-0chain:
| config  | value |
| tests skipped  | FALSE  |
| system-tests  | master  |
| gosdk  | NOT OVERRIDDEN  |
| zboxcli  | staging  |
| zwalletcli  | staging  |
| 0chain  | staging  |
| blobber  | staging  |
| authorizer  | staging  |
| 0box  | staging  |
| 0block  | staging  |
| 0dns  | staging  |
| explorer  | master  |
| 0proxy  | staging  |
| 0search  | staging  |

PR #1283, opened 2022-05-18
Order events example

Sriep:
## Fixes
Change so events are processed in order. 
A block end event is added, and events for the current round are not processed until the round end event has been received.
The events for a round are not processed until they are all gathered together after receiving a round end event, and the hash value checked.
## Todos
This is a bit roungh at the moment. For example we should order events inside a block properly. For this we should probably move the index from [generateBlock](https://github.com/0chain/0chain/blob/staging/code/go/0chain.net/miner/protocol_block.go#L927) into the transaction object so it can be used to index events.
```go
	for i := 0; i < len(iterInfo.currentTxns) && iterInfo.cost < mc.Config.MaxBlockCost() &&
		blockSize < mc.BlockSize() && iterInfo.byteSize < mc.MaxByteSize() && err != context.DeadlineExceeded; i++ {

```

When something goes wrong I just stopped the event database. However the events should be in the event table so we s should be able to recover with a bit more coding.

## Tests

- blobber:
- gosdk:
- system_test:
- zboxcli:
- zwalletcli:
- Other: ...

### DIFF
--- a/code/go/0chain.net/chaincore/block/entity.go
+++ b/code/go/0chain.net/chaincore/block/entity.go
@@ -150,7 +150,7 @@ type Block struct {
 	ChainID   datastore.Key `json:"chain_id"`
 	RoundRank int           `json:"-" msgpack:"-"` // rank of the block in the round it belongs to
 	PrevBlock *Block        `json:"-" msgpack:"-"`
-	Events    []event.Event
+	Events    event.EventList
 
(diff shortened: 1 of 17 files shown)


peterlimg:
I'll send a PR to run sharders in full node mode to ensure all blocks are executed sequentially, the events will be emitted and applied in order too, we will not have the blocks missing issue then. This can  be closed now.

PR #1245, opened 2022-05-06
Blobber and validator, health check kill and shut down

Sriep:
## Changes
https://github.com/0chain/0chain/issues/1040
https://github.com/0chain/0chain/issues/1044
https://github.com/0chain/0chain/issues/1045
https://github.com/0chain/0chain/issues/1046
## Tests 
Tasks to complete before merging PR:
- [ ]  Ensure system tests are passing. If not [Run them manually](https://github.com/0chain/0chain/actions/workflows/system_tests.yml) to check for any regressions :clipboard:
- [ ]  Do any new system tests need added to test this change? do any existing system tests need updated? If so create a PR at [0chain/system_test](https://github.com/0chain/system_test)
- [ ]  Merge your system tests PR to master AFTER merging this PR

## Associated PRs (Link as appropriate):
- blobber:
- gosdk: https://github.com/0chain/gosdk/pull/456
- system_test:
- zboxcli: https://github.com/0chain/zboxcli/pull/229
- zwalletcli:
- Other: ...

PR #1318, opened 2022-05-27
Feature/benchmark docker

dabasov:
## Fixes
Added blocker for benchmarks
Fixed run with pre-generated data
Fixed all errors
Bugfix


code/go/0chain.net/smartcontract/benchmark/main/cmd/mpt.go
```diff
@@ -44,6 +45,7 @@ import (
 )
 
 var BenchDataKey = encryption.Hash("benchData")
+var executor = common.NewWithContextFunc(4)
```
Sriep, line 48: If I understand correctly this is putting a cap on the number of concurrent processes. If so `4` should be a config option, different machines have different capacities.

code/go/0chain.net/smartcontract/storagesc/benchmark_setup.go
```diff
@@ -559,6 +570,37 @@ func AddMockBlobbers(
...
+	var sscId = StorageSmartContract{
+		SmartContract: sci.NewSC(ADDRESS),
+	}.ID
+
+	for i := 0; i < viper.GetInt(sc.NumBlobbers); i++ {
+		err := executor.Run(context.Background(), func(i int) func() error {
+			return func() error {
+				id := getMockBlobberId(i)
+				blobber := &StorageNode{ID: id}
+				err := balances.GetTrieNode(blobber.GetKey(sscId), blobber)
```
Sriep: Why do we need to do this? Whatever the reason we should try to find an alternative way instead of updating the MPT for each blobber. We will be adding health checks to other providers https://github.com/0chain/0chain/pull/1245.

Smartcontracts should get the information from the transaction.CreationDate filed, and REST queries from the event database.

Sriep (changes requested):
We need a solution for the event database

docker.local/bin/build.benchmark.sh
```diff
@@ -0,0 +1,18 @@
+#!/bin/sh
+
+cmd="build"
+
+docker $cmd -f docker.local/build.benchmarks/Dockerfile . -t zchain_benchmarks
+
+if [ "$#" -ne 3 ]; 
+    then echo "illegal number of parameters usage: docker.local/bin/gen_keys.sh signatureScheme [bls0chain or ed25519] absolute_path_to_keyfiles_folder key_file_name"
```
Sriep: Rather than return an error can we use sensible defaults? Otherwise, everyone will need to study this file before they can use it. Or alternately find some other way to make it easy for users.

code/go/0chain.net/smartcontract/benchmark/main/cmd/mpt.go
dabasov: I missed that data generation is run in parallel atm, I think that we should rework it with executor and execute one by one, but in parallel each step.
Also, every step should be stoppable, so we can start failed or stopped generation just were it happened.

Sriep: Yes, but why the magic number. Should this not be configurable.

code/go/0chain.net/smartcontract/benchmark/main/readme.md
```diff
@@ -6,7 +6,37 @@ The blockchain database used in these tests is constructed from the parameters i
 file. Smartcontracts do not (or should not) access tha chain so a populated 
 MPT database is enough to give a realistic benchmark.
 
-To run
+## To run
```
Sriep, line 9: Maybe we should add instructions to the main [0chian readme.md](https://github.com/0chain/0chain/blob/staging/README.md), or a link to here.

code/go/0chain.net/smartcontract/benchmark/main/cmd/mpt.go
```diff
@@ -199,19 +205,36 @@ func setUpMpt(
 
 	var eventDb *event.EventDb
 	if viper.GetBool(benchmark.EventDbEnabled) {
-		timer = time.Now()
+		timer := time.Now()
+
+		tick := func() (*event.EventDb, error) {
+			return event.NewEventDb(dbs.DbAccess{
```
Sriep: I am getting lots of warnings like
```
postgres_1   | 2022-06-02 16:26:46.104 UTC [73] ERROR:  duplicate key value violates unique constraint "idx_miners_miner_id"
```
Perhaps we could turn them off somehow?

dabasov: imo we better fix them, I couldn't find place where duplicates are added

Sriep: I suspect it might be something like this [exist](https://github.com/0chain/0chain/blob/staging/code/go/0chain.net/smartcontract/dbs/event/miner.go#L229) methods. 

Someone decided it was quicker to do queries they expect to fail and respond sensibly to the error than use a more controlled approach. This might or might not be true, but as a side effect, it gives lots of annoying warnings. Also could go wrong on an upgrade that modifies the error messages.

Changing this is probably out of the scope of this PR, but muting the warnings might not be.

dabasov: yep, we can address devops to tune postresss

Sriep (approved):

PR #1326, opened 2022-05-31
Full node sharders

peterlimg:
## Fixes
- Event db inconsistent issue caused by missing blocks
- Issue on getting genesis block from `/v1/block/get`. Previously genesis block is not added to local store, so basically we can not get it by calling the API.

## Changes
- Run sharders in full node mode so that all sharders will have full node data, no block should be missed
- Update block fetcher to support getting finalized block by round number from sharders. Previously, the block hash must be provided, otherwise it can not return full block data. But since sharder need to sync finalized blocks from others, so updated the fetcher to support it.
- Stricter finalize process. For sharders, the finalizing block must connect to the latest finalized block, also it must have at least 3 successors, i.e., there are 3 more blocks extended from it. 

Yes.

## Tests

- blobber:
- gosdk:
- system_test:
- zboxcli:
- zwalletcli:
- Other: ...

### DIFF
--- a/code/go/0chain.net/chaincore/chain/block_fetcher.go
+++ b/code/go/0chain.net/chaincore/chain/block_fetcher.go
@@ -685,6 +685,76 @@ func (c *Chain) GetNotarizedBlock(ctx context.Context, hash string, rn int64) (*
 	return b, nil
 }
 
+func (c *Chain) GetNotarizedBlockFromSharders(ctx context.Context, hash string, rn int64) (*block.Block, error) {
+
+	var bfr = new(blockFetchRequest)
+	bfr.sharders = true
+	bfr.hash = hash
+	bfr.round = rn
+
+	var reply = make(chan BlockFetchReply, 1)
+	bfr.replies = append(bfr.replies, reply)
+
+	cctx, cancel := context.WithTimeout(ctx, 200*time.Millisecond)
+	defer cancel()
+	if err := c.blockFetcher.fetch(cctx, bfr); err != nil {
+		return nil, common.NewErrorf("get_notarized_block",
+			"push to block fetch channel failed, round: %d, err: %v", bfr.round, err)
+	}
+
+	var (
+		cround = c.GetCurrentRound()
+
+		rpl BlockFetchReply
+	)
+
+	select {
+	case <-ctx.Done():
+		return nil, ctx.Err()
+	case rpl = <-reply:
+	}
+
+	switch rpl.Err {
+	case nil:
+	case context.Canceled:
+		return nil, context.Canceled
+	default:
+		return nil, rpl.Err
+	}
+
+	// the block validated and its notarization verified
+	nb := rpl.Block
+
+	var r = c.GetRound(nb.Round)
+	if r == nil {
+		logging.Logger.Info("get notarized block - no round, creating...",
+			zap.Int64("round", nb.Round), zap.String("block", nb.Hash),
+			zap.Int64("cround", cround))
+
+		r = c.RoundF.CreateRoundF(nb.Round)
+	}
+
+	logging.Logger.Info("got notarized block", zap.String("block", nb.Hash),
+		zap.Int64("round", nb.Round),
+		zap.Int("verification_tickers", nb.VerificationTicketsSize()))
+
+	var b *block.Block
+	// This is a notarized block. So, use this method to sync round info
+	// with the notarized block.
+	b, r = c.AddNotarizedBlockToRound(r, nb)
+
+	// Add the round if chain does not have it
+	if c.GetRound(nb.Round) == nil {
+		c.AddRound(r)
+	}
+
+	if b == nb {
+		go c.fetchedNotarizedBlockHandler.NotarizedBlockFetched(ctx, nb)
+	}
+
+	return b, nil
+}
+
 type AfterBlockFetchFunc func(b *block.Block)
 
 func (c *Chain) AsyncFetchFinalizedBlockFromSharders(ctx context.Context,
--- a/code/go/0chain.net/chaincore/chain/block_fetcher.go
+++ b/code/go/0chain.net/chaincore/chain/block_fetcher.go
@@ -189,12 +189,6 @@ func (bf *BlockFetcher) StartBlockFetchWorker(ctx context.Context,
 				continue
 			}
 
-			//if bfr.hash != "" {
-			//	fetching[bfr.hash] = bfr // add, increasing map length
-			//} else {
-			//	fetching[strconv.FormatInt(bfr.round, 10)] = bfr
-			//}
-
 			// if force from sharders
 			if bfr.sharders {
 				if bf.acquire(ctx, shardersl) {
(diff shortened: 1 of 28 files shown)


code/go/0chain.net/chaincore/chain/block_fetcher.go
```diff
@@ -194,7 +205,11 @@ func (bf *BlockFetcher) StartBlockFetchWorker(ctx context.Context,
 
 			// fetch from miners first
 			if bf.acquire(ctx, minersl) {
-				fetching[bfr.hash] = bfr
+				if bfr.hash != "" {
+					fetching[bfr.hash] = bfr // add, increasing map length
+				} else {
+					fetching[strconv.FormatInt(bfr.round, 10)] = bfr
```
dabasov, line 211: not clear for me,what if there will be different blocks for a round?

code/go/0chain.net/chaincore/chain/protocol_round.go
```diff
@@ -308,6 +327,11 @@ func (c *Chain) finalizeRound(ctx context.Context, r round.RoundI) {
 			zap.String("lfb block", lfb.Hash))
 		for idx := range frchain {
 			fb := frchain[len(frchain)-1-idx]
+			if roundNumber-fb.Round < 3 {
```
dabasov, line 330: - roundNumber is current block to finalize, most likely last finalized
- frchain starts at single notarized block on around (newest to finalize) and ends with oldest not finalized (ancestor of single on round)

so we can safely finalize all frchain regardless to ``roundNumber-fb.Round`` length, am I missing something?

code/go/0chain.net/chaincore/chain/worker.go
```diff
@@ -290,45 +286,61 @@ func (c *Chain) finalizeBlockProcess(ctx context.Context, fb *block.Block, bsh B
...
-		return
+		return nil
 	}
 
 	Logger.Debug("start to finalize block",
 		zap.Int64("round", fb.Round),
 		zap.String("block", fb.Hash),
 		zap.String("prev block", fb.PrevHash))
 
+	isSharder := node.Self.IsSharder()
```
dabasov, line 297: this code should be moved to miner/sharder packages perhaps, and chaincore/chain should not know anything about difference between miners and sharders, but it's easy said than done. lgtm for now, maybe can refactor it after mainnet

code/go/0chain.net/chaincore/chain/worker.go
```diff
@@ -340,12 +352,39 @@ func (c *Chain) finalizeBlockProcess(ctx context.Context, fb *block.Block, bsh B
...
+		// get previous finalized block
+		pr := c.GetRound(fb.Round - 1)
+		if pr == nil {
+			Logger.Error("finalize block - previous round not found",
+				zap.Int64("round", fb.Round))
+			return errors.New("previous round is missing")
+		}
+
+		prevBlockHash := pr.GetBlockHash()
+		if prevBlockHash == "" {
```
dabasov, line 369: probably, I'm missing something. sometimes previous round can't be finalize, since it has more than one notarizations, this round will be finalized by chain, when it's siblings will have singlefinalized block on a round. is it different for sharders?

code/go/0chain.net/chaincore/chain/block_fetcher.go
peterlimg: This is for when only round number is provided to get the block. If the request goes to sharders, the finalized block will be fetched. If it goes to miners, the heaviest notarized block will be returned. 

Even though, it is possible that the we send two requests for the same round, suppose a) request with block hash b) request with round number. These two requests will be cached in different `fetching` slots, `fetching[hash]` and `fetching[round number]`. The first one that successfully returns the block will try to remove the `fetching[hash]` and `fetching[round]` base on the `hash` and `round` of returned block. Son if the returned block's hash is the same as the a) block hash, then the two slots will both be removed from `fetching`. Otherwise, only round will be removed, the hash will be removed only when the response of that block is returned.

code/go/0chain.net/chaincore/chain/protocol_round.go
peterlimg: > frchain starts at single notarized block on a round

Yes, but the situation is that the 'single notarized block' is not true actually, it only means the node only received one notarized block of that round when doing the finalization. One case is there are 2 blocks notarized for a round, sharders received block_1, and pushed to finalize process, before it received the second notarized block, this block_1 is treated as the `single notarized block` and finalized. This is a problem if the miners finalized the other block, which will then stuck the chain. Please also check https://0chain.slack.com/archives/C03788G8N2D/p1653894136793029 for the issue.

code/go/0chain.net/chaincore/chain/worker.go
peterlimg: We are finalizing blocks in `frchain` from LFB+1, it's normal for miners that the previous round is not finalized as they could be restarted and come up with LFB acquired from remote. But for sharders, they all have full blocks till LFB, the frchain is start from LFB+1, and blocks in it are all chained, also we finalize blocks one by one, so it's previous round is actually finalized. If its siblings have single finalized block, but since they are not in the frchain, that means the sibling singlefinalized block does not have enough successors. 

BTW, if I don't do this checking, the previous logic will finalize the block anyway, which means we will finalize the block even it's previous round is not finalized. Apparently this is not what we want, if this happens, the event db will be messed up.

code/go/0chain.net/chaincore/chain/protocol_round.go
dabasov: completely missed this discussion

code/go/0chain.net/chaincore/chain/worker.go
dabasov: > which means we will finalize the block even it's previous round is not finalized. 

it's normal for our protocol, previous round can be not finalized, 
so we can have b0<-b1<-b2<-b3
imagine we want to finalize b3 now, b0 can be finalized and b1 and b2 can be not finalized, simply because they have other notarized blocks on this round due to leaders equivocations. 
if b3 is the only on his round (is in the lock set, b4 is build upon it and finalized) then b3 can be finalized and at this moment b2 and b3 will be finalized, not earlier.

this branches of notarized can be build by a sequence of equivocating leaders, they should cooperate. we have this byzantine test (not implemented yet), in general such chains won't be too long, since it is not likely to have several dishonest leaders in a row, but two leaders can cooperate easily

peterlimg: > it's normal for our protocol, previous round can be not finalized

If it's previous round can not be finalized, then the current round can not be finalized too. Of course we will have round that it's previous round is not finalized, but that's not the case the code is processing here. There are mainly 3 steps about the finalize round process:
1) Block is notarized and push it to finalize process channel, so for the example you provided, the b3 is pushed to the channel.
2) In finalize worker, compute new LFB and getting the blocks that can be finalized. 
   For the `b0<-b1<-b2<-b3`, we will only finalized `b0` due to the new rule that we will not finalized a block until it has at least 3 successors built on it. So only `b0` will go to the next step. 
3) Finalize the blocks got from step 1) and finalized them one by one from LFB. 
  In this step, we will check if the current finalizing block's previous round is finalized on sharders, because the first block is connect to LFB, and sharders have full blocks, so the LFB must be finalized, and since we finalize one by one, it's previous round must have been finalized too. 
  
So I think you are talking about the case that happen in step 1), i.e we try to finalized b3 and push it to finalized process channel. And yes, we will push it to the process channel even its previous round is not finalized, but it's not the case we are processing here, the code is doing the work on step 3).

dabasov (approved):
lgtm

PR #1339, opened 2022-06-01
fixed empty response on get_block and get_block_by_hash

dabasov:
the problem was that events were not written in the events db, now we store all blocks that were added to the chain

### DIFF
--- a/code/go/0chain.net/chaincore/block/block_eventdb.go
+++ b/code/go/0chain.net/chaincore/block/block_eventdb.go
@@ -36,7 +36,7 @@ func CreateBlockEvent(block *Block) (error, event.Event) {
 		return fmt.Errorf("error marshalling block: %v", err), event.Event{}
 	}
 
-	event := event.Event{
+	return nil, event.Event{
 		BlockNumber: block.Round,
 		TxHash:      "",
 		Type:        int(event.TypeStats),
@@ -44,6 +44,4 @@ func CreateBlockEvent(block *Block) (error, event.Event) {
 		Index:       block.Hash,
 		Data:        string(data),
 	}
-
-	return nil, event
 }
--- a/code/go/0chain.net/chaincore/block/block_eventdb.go
+++ b/code/go/0chain.net/chaincore/block/block_eventdb.go
@@ -1,10 +1,10 @@
 package block
 
 import (
-	"0chain.net/core/util"
 	"encoding/json"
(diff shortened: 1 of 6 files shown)


Sriep (approved):

code/go/0chain.net/chaincore/chain/entity.go
```diff
@@ -693,6 +693,9 @@ func (c *Chain) addBlock(b *block.Block) *block.Block {
 		return eb
 	}
 	c.blocks[b.Hash] = b
+
+	c.pushBlockEvent(b)
```
peterlimg: We only save finalized blocks to event db, so perhaps do it in https://github.com/0chain/0chain/blob/9b72325418072680429bc00b654367b1234b6af9/code/go/0chain.net/chaincore/chain/protocol_block.go#L351.

peterlimg (changes requested):

code/go/0chain.net/chaincore/block/block_eventdb.go
```diff
@@ -1,6 +1,7 @@
 package block
 
 import (
+	"0chain.net/core/util"
```
peterlimg: Run goimports to format

peterlimg (changes requested):

code/go/0chain.net/chaincore/chain/entity.go
dabasov: I've done it intentionally, we might need all proposed blocks here, not only finalized blocks, at least all notarized so I suggest to save them here

peterlimg: okay, we can leave this and consider to do finalized blocks only if we find those none finalized proposed blocks are not being used at all in the future.

peterlimg (approved):
lgtm

code/go/0chain.net/chaincore/chain/entity.go
peterlimg: Oh, event db only exists in sharders, so we can not do this for miners.

peterlimg (changes requested):

code/go/0chain.net/chaincore/chain/entity.go
dabasov: oooops

peterlimg (approved):
lgtm

PR #1345, opened 2022-06-05
Enhacement/blocks management

lpoli:
## Fixes
This PR will fix the inode issues that prevails when writing in single disk and taking 9 first digit of hex of 256 hash value resulting in creation of 9^16 number of directories.
The implementation is simply even distribution of blocks in directories. So sharder would create a base directory for blocks, say `/path/to/store/blocks`, then there will be k0...k1999 directories each containing 0...1999 directories each containing 2000 blocks.
An example path  that is storing thousands blocks is; `/path/to/store/blocks/k0/0`. All blocks are stored with `hash`.dat as a file name. They are compressed and stored.

There can be any number of disks to store the blocks. Sharder will choose disk based on the strategy set up in yaml file. It can be one of `round_robin`, `random`, `min_size_first` and `min_count_first`. All failed disks will be stored in a map and wiil be removed from active list. A goroutine checks if failed disk is able to store block and if able then it will moved to active list.

The whereabout of the block i.e. path of block will be stored in rocksdb in `bwr` column family. A block can be in either disk or cold storage or both. This information is stored in rocksdb with `hash` as key and it `whereabout` as value.

**Note**: Directory will be created on demand. It won't be created while initializing. But once directory is created it will not be deleted because it will be used to store the blocks. Will be clear in blocks movement section.

### Block Movement:
When sharder has put configuration for cold Tiering or cold Storage, then older blocks will move to this cold tier based on the interval set up in yaml file. The cold storage must be minio compatible.
If sharder has set up for cold storage then recently written block's hash as value with its timestamp as key will be written in rocksdb in `ubr` column family. Timestamp is its creation date which is always same for the block. A goroutine will pick blocks from the database and will move to the cold Tier. The new path of the block is stored in above mentioned `bwr` column family.

If the `delete_local` is set to true then it will delete block from the disk. So with block movement, oldest blocks will be deleted leaving directories empty. Now if the sharder has written all the k1999/1999 directories then it will come back to k0/directory, and if it is able to store block it will continue otherwise it will remove the disk from active list and put it in inactive list. 

### Cache:
A simple cache is implemented based on LRU replacement algorithm. This cache is simply a directory in SSD drive that will store recently read block. It will store block uncompressed. When the cache is full, it will delete oldest unused block.

## Tests

- blobber:
- gosdk:
- system_test:
- zboxcli:
- zwalletcli:
- Other: ...


service-0chain:
| config  | value |
| tests skipped  | FALSE  |
| system-tests  | master  |
| gosdk  | NOT OVERRIDDEN  |
| zboxcli  | staging  |
| zwalletcli  | staging  |
| 0chain  | enhacement/blocks-management  |
| blobber  | staging  |
| authorizer  | staging  |
| 0box  | staging  |
| 0block  | staging  |
| 0dns  | staging  |
| explorer  | master  |
| 0proxy  | staging  |
| 0search  | staging  |

code/go/0chain.net/core/ememorystore/connection.go
```diff
@@ -46,7 +46,37 @@ func CreateDB(dataDir string) (*gorocksdb.TransactionDB, error) {
 	return gorocksdb.OpenTransactionDb(opts, tdbopts, dataDir)
 }
 
+func OpenDB(
+	dir string,
+	cfs []string,
```
peterlimg: Comments the cfs or rename to columnFamilies to make it readable. Otherwise, users need to guess what does it mean.

code/go/0chain.net/core/ememorystore/connection.go
```diff
@@ -46,7 +46,37 @@ func CreateDB(dataDir string) (*gorocksdb.TransactionDB, error) {
 	return gorocksdb.OpenTransactionDb(opts, tdbopts, dataDir)
 }
 
+func OpenDB(
```
peterlimg: I would suggest to name it as OpenDBWithColumnFamilies and remove the db creation code to make this function simpler. We can call the CreateDB() to create db.

code/go/0chain.net/sharder/blockstore/bwr.go
```diff
@@ -0,0 +1,188 @@
...
+const (
+	DiskTier        WhichTier = 2 // Block is in disk only
+	ColdTier        WhichTier = 4 // Block is in cold storage only
+	DiskAndColdTier WhichTier = 8 // Block is in both disk and cold storages
+)
+
+const (
+	BWRCF = "bwr" // bwr column family
+	UBRCF = "ubr" // ubr column family
+	BMR   = "bmr" // block meta record
```
peterlimg: Please comment what do 'bwr' 'ubr'  'bmr' stand for

code/go/0chain.net/sharder/blockstore/bwr.go
```diff
@@ -0,0 +1,188 @@
...
+	ColdPath  string `json:"cp,omitempty"`
+}
+
+func (bwr *blockWhereRecord) addOrUpdate() error {
+	data, err := json.Marshal(bwr)
+	if err != nil {
+		return err
+	}
+	wo := gorocksdb.NewDefaultWriteOptions()
+	err = bmrDB.PutCF(gorocksdb.NewDefaultWriteOptions(), bwrHandle, []byte(bwr.Hash), data)
```
peterlimg: use the created `wo` instead of creating a new WriteOptions object

code/go/0chain.net/sharder/blockstore/bwr.go
```diff
@@ -0,0 +1,188 @@
...
+
+type blockWhereRecord struct {
+	Hash    string    `json:"-"`
+	Tiering WhichTier `json:"tr"`
+	//For disk volume it is simple unix path. For cold storage it is "storageUrl:bucketName"
+	BlockPath string `json:"vp,omitempty"`
+	ColdPath  string `json:"cp,omitempty"`
+}
+
+func (bwr *blockWhereRecord) addOrUpdate() error {
```
peterlimg: method name could be  `save` and simpler.

code/go/0chain.net/sharder/blockstore/bwr.go
```diff
@@ -0,0 +1,188 @@
...
+	if err != nil {
+		return err
+	}
+	wo := gorocksdb.NewDefaultWriteOptions()
+	err = bmrDB.PutCF(gorocksdb.NewDefaultWriteOptions(), bwrHandle, []byte(bwr.Hash), data)
+	wo.Destroy()
+	return err
+}
+
+func getBWR(hash string) (*blockWhereRecord, error) {
```
peterlimg, line 58: getBlockWhereRecord() would be easier to read

code/go/0chain.net/sharder/blockstore/bwr.go
```diff
@@ -0,0 +1,188 @@
...
+
+const (
+	BWRCF = "bwr" // bwr column family
+	UBRCF = "ubr" // ubr column family
+	BMR   = "bmr" // block meta record
+)
+
+var bmrDB *gorocksdb.DB
+var bwrHandle *gorocksdb.ColumnFamilyHandle
+var ubrHandle *gorocksdb.ColumnFamilyHandle
```
peterlimg: Please use meaningful variables to make it easier for users. So 
`blockMetaRecordHandle`, `blockWhereRecordHandle`, and actually I don't know what `ubr` stands for yet. I'll need to read more  to understand what it is.

code/go/0chain.net/sharder/blockstore/bwr.go
```diff
@@ -0,0 +1,188 @@
...
+
+type blockWhereRecord struct {
+	Hash    string    `json:"-"`
+	Tiering WhichTier `json:"tr"`
+	//For disk volume it is simple unix path. For cold storage it is "storageUrl:bucketName"
+	BlockPath string `json:"vp,omitempty"`
```
peterlimg: Change vp to bp

code/go/0chain.net/sharder/blockstore/bwr.go
```diff
@@ -0,0 +1,188 @@
...
+
+/**********************************Unmoved block record***************************************/
+type unmovedBlockRecord struct {
+	Hash string `json:"h"`
+	// CreateAt duration passed from epoch date
+	CreatedAt common.Timestamp `json:"c"`
+}
+
+func (ubr *unmovedBlockRecord) Add() error {
+	k := strconv.FormatInt(int64(ubr.CreatedAt), 10)
```
peterlimg: could be more efficient to use binary package to write the integer value to bytes slice. For example:

```go
buf := bytes.NewBuffer(nil)
if err := binary.Write(buf, binary.LittleEndian, ubr.CreateAt); err != nil {
...
}

// get the bytes
data := buf.Bytes()
}
```

code/go/0chain.net/sharder/blockstore/bwr.go
```diff
@@ -0,0 +1,188 @@
...
+	wo := gorocksdb.NewDefaultWriteOptions()
+	err := bmrDB.PutCF(gorocksdb.NewDefaultWriteOptions(), ubrHandle, []byte(k), []byte(ubr.Hash))
+	wo.Destroy()
+	return err
+}
+
+func (ubr *unmovedBlockRecord) Delete() error {
+	k := strconv.FormatInt(int64(ubr.CreatedAt), 10)
```
peterlimg: Same as above, and we can create a common function to convert integer value to bytes

code/go/0chain.net/sharder/blockstore/bwr.go
```diff
@@ -0,0 +1,188 @@
...
+
+	go func() {
+		defer close(ch)
+
+		ro := gorocksdb.NewDefaultReadOptions()
+		defer ro.Destroy()
+
+		it := bmrDB.NewIteratorCF(ro, ubrHandle)
+		defer it.Close()
+		for it.SeekToFirst(); it.Valid() && bytes.Compare(it.Key().Data(), maxPrefix) != 1; it.Next() {
```
peterlimg: What does the `bytes.Compare() !=  1` expect, can we write it in more general way, or comment the logic behind it.

code/go/0chain.net/sharder/blockstore/bwr.go
```diff
@@ -0,0 +1,188 @@
...
+				CreatedAt: common.Timestamp(createdAt),
+			}
+			keyS.Free()
+			valueS.Free()
+		}
+	}()
+	return ch
+}
+
+func initBWR(viper *viper.Viper, mode, workDir string) {
```
peterlimg: Rename to `initBlockWhereRecord`

code/go/0chain.net/sharder/blockstore/bwr.go
```diff
@@ -0,0 +1,188 @@
...
+	if viper == nil {
```
peterlimg: Pass the cache size as param and remove the viper

peterlimg (changes requested):
reviewed partially

code/go/0chain.net/sharder/blockstore/bwr.go
```diff
@@ -0,0 +1,188 @@
...
+		panic("bwr config not provided")
+	}
+
+	dbPath := filepath.Join(workDir, "data/rocksdb", BMR)
```
peterlimg: The "data/rocksdb" is the root dir path of all rocksdb files, it also includes "data/rocksdb/state", "data/rocksdb/mb", "data/rocksdb/dkg", etc. The below `defer os.RemoveAll(dbPath)` could remove all the files. So perhaps we need to put this in a sub-dir like "data/rocksdb/bwr".

code/go/0chain.net/sharder/blockstore/bwr.go
```diff
@@ -0,0 +1,188 @@
...
+	cfs := []string{"default", BWRCF, UBRCF}
+	bwrOpt := gorocksdb.NewDefaultOptions()
+	bwrOpt.OptimizeForPointLookup(cacheSize)
+	bwrOpt.SetAllowConcurrentMemtableWrites(false)
+
+	ubrOpt := gorocksdb.NewDefaultOptions()
+
+	cfsOpts := []*gorocksdb.Options{gorocksdb.NewDefaultOptions(), bwrOpt, ubrOpt}
+	switch mode {
+	case "restart":
```
peterlimg, line 154: Please comment what `mode` could be, and what does the "restart" mean here.

code/go/0chain.net/sharder/blockstore/cache.go
```diff
@@ -0,0 +1,253 @@
+package blockstore
+
+// Cache is a simple implementation using simple lru for storing block's minimal data.
```
peterlimg, line 3: rocksdb has its own implementation of cache, so let's not implement new cache on top of rocksdb again. Unless we have encountered performance issue with the rocksdb itself, we'd better not optimize ahead. And since the blocks is the fundamental part of the blockchain, we need to make sure it works as simple as possible, and as robust as possible.

peterlimg (changes requested):

code/go/0chain.net/sharder/blockstore/common.go
```diff
@@ -0,0 +1,351 @@
...
+
+	//BlockLimit Limit the number of blocks to store in any tier
+	BlockLimitNumber = 10000000000 // 10^10
+	GB               = 1024 * 1024 * 1024
+)
+
+//Common errors
+var (
+	ErrInodesLimit = func(vPath string, inodesToMaintain uint64) error {
+		return fmt.Errorf("Volume %v has inodes lesser than inodes to maintain, %v", vPath, inodesToMaintain)
```
peterlimg: Please avoid capital error message, so as the ones below

code/go/0chain.net/sharder/blockstore/common.go
```diff
@@ -0,0 +1,351 @@
...
+
+	availableInodes = volStat.Ffree
+	totalInodes = volStat.Files
+	availableSize = volStat.Bfree * uint64(volStat.Bsize)
+	return
+}
+
+func getCurIndexes(fPath string) (curKInd, curDirInd int, err error) {
+	var f *os.File
+	if f, err = os.Open(fPath); err != nil {
```
peterlimg, line 118: The file is not closed

code/go/0chain.net/sharder/blockstore/common.go
```diff
@@ -0,0 +1,351 @@
...
+func getCurrentDirBlockNums(dir string) (int, error) {
+	dirEntries, err := os.ReadDir(dir)
+	if err != nil {
+		return 0, err
+	}
+	return len(dirEntries), nil
+}
+
+func updateCurIndexes(fPath string, curKInd, curDirInd int) error {
+	f, err := os.Create(fPath)
```
peterlimg, line 158: The file is not closed

code/go/0chain.net/sharder/blockstore/common.go
```diff
@@ -0,0 +1,351 @@
...
+		}
+
+		for j := 0; j < dcl; j++ {
+			blocksPath := filepath.Join(subDirPath, fmt.Sprint(j))
+			_, err = os.Stat(subDirPath)
+			if err != nil {
+				continue
+			}
+
+			f, err := os.Open(blocksPath)
```
peterlimg, line 193: file is not closed

peterlimg (changes requested):

code/go/0chain.net/sharder/blockstore/bwr.go
lpoli: Since return type is `blockWhereRecord` it should not be harder to guess for reader.

code/go/0chain.net/sharder/blockstore/bwr.go
lpoli: It is insider sub-dir already.
`dbPath := filepath.Join(workDir, "data/rocksdb", BMR)` puts data inside `data/rocksdb/bmr`

code/go/0chain.net/sharder/blockstore/cache.go
lpoli: @peterlimg The rocksdb cache is for its keys. Not sure if it could work in this case.
The uncompressed blocks are stored in a directory as specified in `yaml` file. Sharder can provide cache-size in config file and ready to go.

Blocks could be in Hard drive or in minio. Recently used block are stored uncompressed in cache directory and when the read request for block is made, it first checks cache directory.

code/go/0chain.net/sharder/blockstore/disk.go
```diff
@@ -0,0 +1,740 @@
...
+
+var volumesMap map[string]*volume
+var unableVolumes map[string]*volume
+
+const (
+	DirPrefix = "blocks/K"
+	// Hot directory content limit
+	// Contains 2000 directories that contains 2000 blocks each, so one twoKilo directory contains 4*10^6blocks.
+	// So 2000 such twokilo directories will contain 8*10^9 blocks
+	DCL = 2000
```
peterlimg: Use readable name when possible, so use DirectoryContentLimit please

code/go/0chain.net/sharder/blockstore/disk.go
```diff
@@ -0,0 +1,740 @@
...
+	prevInd int
+	err     error
+}
+
+type diskTier struct {
+	Volumes          []*volume // List of hot volumes
+	SelectNextVolume func(volumes []*volume, prevInd int)
+	SelectedVolumeCh <-chan selectedDiskVolume // volume that will be used to store blocks next
+	PrevVolInd       int
+	Mu               Mutex
```
peterlimg, line 55: Please make the member variables private if the diskTier itself is private.

code/go/0chain.net/sharder/blockstore/store.go
```diff
@@ -1,29 +1,403 @@
...
+func (sm *blockStore) Read(hash string, round int64) (b *block.Block, err error) {
+	logging.Logger.Info("Reading block: " + hash)
+	return sm.read(hash, round)
+}
+
+func (sm *blockStore) Delete(hash string) error {
+	return nil // Not implemented
+}
+
+func InitializeStore(ctx context.Context, sViper *viper.Viper, workDir string) {
```
peterlimg: Rename to avoid name repeating, so could be Init(), and be used like `blockstore.Init()`.

Also, please define struct to accept params instead of passing the viper.

code/go/0chain.net/sharder/blockstore/store.go
```diff
@@ -1,29 +1,403 @@
...
+
+			return nil
+		}
+
+		store.read = func(hash string, round int64) (b *block.Block, err error) {
+			bwr, err := getBWR(hash)
+			if err != nil {
+				return nil, err
+			}
+			return store.diskTier.read(bwr.BlockPath)
```
peterlimg, line 214: Sharders block syncing process relies on getting block by round number, but seems this could only get by hash. So please make sure getting by round number is supported, otherwise the chain would not work as expected.

peterlimg (changes requested):

sagchain:
@lpoli what are the next steps and ETA on this? Can you please share an update?

lpoli:
This PR is replaced by  https://github.com/0chain/0chain/pull/2044 as the requirement of blocks management changed as well.

PR #1362, opened 2022-06-08
added nonce validations

### LINKED ISSUE #1320: Add transaction_nonce as a required field on  /v1/transaction/put
When submitting a transaction to /v1/transaction/put transaction_nonce is not a required field despite it being fundamental to how transactions are processed.
Internally this results in the field defaulting to zero which is a value that will never succeed as the first nonce is required to be 1 
(0 indexed + 1)
The result is a transaction that submits fine but is never processed by the miners.

Acceptance criteria:
Add field validation to the API to ensure transaction_nonce is present and a positive integer

dabasov:
added transaction nonce validations


peterlimg (approved):
lgtm

PR #1399, opened 2022-06-17
fix sharder standard tests

fabioDMFerreira:
Allow rounds to go beyond while testing sharders conductor tests. 

The rounds are generated too fast and it is not possible to stop the network execution at a specific round. I think  `allow_beyond:true` should be the default behavior.


dabasov (approved):
lgtm

PR #1406, opened 2022-06-18
Fix conductor standard tests building step

fabioDMFerreira:
Use global configuration to check DKG is enabled

This PR fixes the building of the miner integration image in the conductor standard tests (https://github.com/0chain/0chain/runs/6944935658?check_suite_focus=true).

@peterlimg I realized this PR change was done by you previously (https://github.com/0chain/0chain/commit/8e107fc0cf4d27d7cb2b72e4e770d476e1a68fa8#diff-2ade906b5071a51cf9366cd5cd98126ad8e9c979fa5649c9e54cef547c526d98R343). For some reason your change was overridden. I recommend finding the root cause as other changes may be overridden as well.


peterlimg (approved):
lgtm

PR #1410, opened 2022-06-20
Fix/accept round beyond on wait round

fabioDMFerreira:
This fixes conductor standard tests (https://github.com/0chain/0chain/actions/runs/2528036570).

The tests are failing because while waiting for a round it always goes beyond the specified round. The creation of rounds is too fast. One solution would be to block round creation after reaching the round specified in the test configuration. The solution of this PR allows the creation of rounds higher than the specified in the test configuration.

FYI @rrrooommmaaa I have inverted the behavior of the `allow_beyond`.


rrrooommmaaa (approved):
Why not?

peterlimg (approved):
lgtm

Issue #1187, opened 2022-04-19
Remove errors for context cancel


dabasov:
Context cancel can happen during generating and verifying the block, if notarization is received before these actions are finished. Now interrupted actions (transaction calculation, block generation, verification etc) logs errors, it can be confusing for admins of miners/sharders

peterlimg:
We can change it to `DEBUG` logs as we'll need to know what happened when issue happens

dabasov:
Yeah, under "remove" I meant sort them out and log with DEBUG level

dabasov:
fixed https://github.com/0chain/0chain/pull/1416

PR #1416, opened 2022-06-21
Log debug messages when the context of block computing is cancelled.

fabioDMFerreira:
#1187 Change log message types of errors related to context cancellation on computing block.


peterlimg (approved):
lgtm

PR #1418, opened 2022-06-21
Remove (allocation,blobber) from challenge partition if blobber is empty now

### LINKED ISSUE #646: Remove challenges for deleted files.
WrtieMarkers need to be sent to chain when we `add` a new file, `update` a file, `delete` a file, `copy` or `move` a file, and `rename` a file.

0chain needs to make appropriate changes to challenges. For instance, removing challenges for deleted files. Some of this already works, for example when a file is added. We need to fill in the blanks.

This needs to be coordinated with a matching change to the blobber.

Any crud action that changes the tree is reported to the chain because the Merkle root of the allocation changes as a result

ssardana08:
## Changes
- remove allocation from blobber_allocation_challenge partition, if blobber.UsedSize is now empty

## Tests

- blobber:
- gosdk:
- system_test:
- zboxcli:
- zwalletcli:
- Other: ...

### DIFF
--- a/code/go/0chain.net/smartcontract/storagesc/blobber.go
+++ b/code/go/0chain.net/smartcontract/storagesc/blobber.go
@@ -671,7 +671,13 @@ func (sc *StorageSmartContract) commitBlobberConnection(
 	}
 
 	// the first time the allocation is added  to the blobber, created related resources
-	if blobAlloc.BlobberAllocationsPartitionLoc == nil {
+	if blobAlloc.Stats.UsedSize == 0 {
+		err = removeAllocationFromBlobber(sc, t.ClientID, blobAlloc.BlobberAllocationsPartitionLoc, alloc.ID, balances)
+		if err != nil {
+			return "", common.NewErrorf("commit_connection_failed",
+				"removing allocation from blobAlloc partition: %v")
+		}
+	} else if blobAlloc.BlobberAllocationsPartitionLoc == nil {
 		if err := sc.blobberAddAllocation(t, blobAlloc, uint64(blobber.BytesWritten), balances); err != nil {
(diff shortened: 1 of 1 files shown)


code/go/0chain.net/smartcontract/storagesc/blobber.go
```diff
@@ -685,7 +685,13 @@ func (sc *StorageSmartContract) commitBlobberConnection(
 	}
 
 	// the first time the allocation is added  to the blobber, created related resources
-	if blobAlloc.BlobberAllocationsPartitionLoc == nil {
+	if blobAlloc.Stats.UsedSize == 0 {
```
peterlimg, line 673: Isn't the UsedSize incremental only? so this is actually checking if the blobber is redeeming a read marker that its allocation had no data been written to it yet, is that possible?

peterlimg (changes requested):

code/go/0chain.net/smartcontract/storagesc/blobber.go
ssardana08: it's not always incrementing...
`blobAlloc.Stats.UsedSize += commitConnection.WriteMarker.Size`
even though it's a +=, but on deletion of file Writemarker.Size is in negative which will reduce the usedSize

peterlimg: @cnlangzi please confirm that if the WriteMarker.Size could be negative.

cnlangzi: @sum-it08 @peterlimg  WriteMaker.Size can be any value( >0,0,<0). eg
- upload/update: >0
- delete: <=0
- rename: 0

peterlimg (approved):
lgtm

PR #1438, opened 2022-06-24
Fix/nullify blobber loc

ssardana08:
## Fixes
- on removal of allocation from blobAlloc partition, make location nil
## Tests

- blobber:
- gosdk:
- system_test:
- zboxcli:
- zwalletcli:
- Other: ...

### DIFF
--- a/code/go/0chain.net/smartcontract/storagesc/models.go
+++ b/code/go/0chain.net/smartcontract/storagesc/models.go
@@ -920,6 +920,8 @@ func removeAllocationFromBlobber(
 		return fmt.Errorf("could not update blobber partitions locations node: %v", err)
 	}
 
+	allocPartLoc = nil
+
 	return nil
 }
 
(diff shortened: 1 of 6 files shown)


code/go/0chain.net/smartcontract/storagesc/models.go
```diff
@@ -890,6 +890,8 @@ func removeAllocationFromBlobber(
 	if err := blobAllocsParts.Save(balances); err != nil {
 		return fmt.Errorf("could not update blobber allocation partitions: %v", err)
 	}
+	// nullifying the blobber alloc challenge pasrtition location
+	allocPartLoc = nil
```
peterlimg: This nullifying would not work as the allocPartLoc is a pointer param passed by the method caller, we can modify it's value through dereference (`*allocPartLoc`), but can not modify the pointer itself. So  perhaps we need to pass the blobberAllocation, and set its `BlobberAllocationsPartitionLoc` as nil.

peterlimg (changes requested):

peterlimg (approved):
lgtm

PR #1439, opened 2022-06-25
Blobber and validator, health check kill and shut down 

### LINKED ISSUE #1020: Delete provider plus blacklist
We need to be able to delete any provider ([miner](https://github.com/0chain/0chain/blob/staging/code/go/0chain.net/smartcontract/minersc/miner.go#L138), [sharder](https://github.com/0chain/0chain/blob/staging/code/go/0chain.net/smartcontract/minersc/sharder.go#L158), blobber, validator or [athoriser](https://github.com/0chain/0chain/blob/staging/code/go/0chain.net/smartcontract/zcnsc/authorizer.go#L136)). Further we need to keep a record of deleted providers to prevent a deleted provider immediately rejoining the 0chain.

1. Add delete blobber smart contract endpoint.
  * Add a new field, `bool disabled` to a StorageNode.
  * A disabled blobber cannot be used for any new allocation.
 2. Add a `delete validator` smart contract endpoint.
  * Add a new field `bool disdabled` to ValidatorNode.
  * A disabled validator cannot be used to validate any new challenges.
3. Modify the `add provider` sc endpoints to not allow previously deleted providers to be added. 
 * As part of the validation process, ensure that the provider does not already exist with `disabled` set to `true`.

### LINKED ISSUE #1040: Providers
# Provider

A provider is one of currently five objects that perform services for 0chian. There will be a linked wallet that can control the provider's settings. Each provider will also have a number of delegates; These are users that have invested 
tokens in the provider. 

Providers are given rewards when they perform services. These rewards are distributed amongst the delegates based on
creative stake size, the provider's control wallet can also get a fixed percentage of the reward as a service change.

## Commitment

While active each provider gains commitments to provide services.

- `Miners` produce blocks.
- `Sharders` stores the blockchain and other related data, such as the event database. They also support a query REST API.
- `blobbers` store user data files.
- `validators` validate challenges
- `authorizers` validate client transactions.

## Active and inactive

An active provider can take on new commitments while an inactive provider, for whatever reason, will not be able to take on any new commitments. Below are possible reasons for a provider to be inactive.

#### Envoirmental

If a provider is guaranteed to always fail its validation test to take on a new commitment we can consider it inactive.

An example of this might be if we reduced the maximum [max_challenge_completion_time](https://github.com/0chain/0chain/blob/staging/docker.local/config/sc.yaml#L71) this could result in blobbers with a higher [challenge_completion_time](https://github.com/0chain/blobber/blob/staging/config/0chain_blobber.yaml#L34) becoming inactive.

To help in this situation we will have an API endpoint that will return the active and inactive status of a provider. And indicate the reason. So in the example of `max_challenge_completion_time` above a blobber could run an [update_blobber_settings](https://github.com/0chain/0chain/blob/staging/code/go/0chain.net/smartcontract/storagesc/sc.go#L259) transaction to reduce its challenge_completion_time and regain its active state.

#### Health check

Every so often a provider sends a heath check transaction. If the provider has not provided a health check in the required time it cannot take on any new commitments. This is similar to environmental inactivity, but for the [moment the health time is hard coded](https://github.com/0chain/0chain/blob/staging/code/go/0chain.net/smartcontract/storagesc/blobber.go#L22).

#### Shuts down

 A provider can send a `shut_down` transaction requesting that it be shut down. It no longer gets any new commitments. It is still expected to fulfil existing commitments and revives rewards as normal. Cannot be reversed.

#### Kill

A special user, the chain owner, can `kill` a provider by sending a kill transaction. This is similar to shut down but the provider no longer receives any rewards. This is irreversible.

#### Pending issues

New `am i active` API query https://github.com/0chain/0chain/issues/1041
Update provider settings https://github.com/0chain/0chain/issues/1042
Provider health check https://github.com/0chain/0chain/issues/1044
Provider shut down https://github.com/0chain/0chain/issues/1045
Killing provider https://github.com/0chain/0chain/issues/1046

### LINKED ISSUE #1044: Provider health check
Every so often a provider sends a heath check transaction. If the provider has not provided a health check in the required time it cannot take on any new commitments https://github.com/0chain/0chain/issues/1040.

This functionally works for blobbers but is only partially implemented for other providers (miners, sharders, blobbers, validators and authorised. We wish to move the health check code into common code where possible.

1. A `?_health_check` smart contract endpoint for each provider. This will reset the LastHeathCheck member of a provider.
2. New health check period. A new validation check of this period against the time since the last health is to be provided, to be used before a provider commits to various tasks. 
3. The last heath check needs to be kept with the object in the MPT. This and summary stats of period passing or railing health checks can be updated in the event database after each health check.

#### Actions to be dependant on health checks
- `sharder` acquiring new blocks.
- `sharder` sending REST API queries.
- `miner` submitting new blocks of transactions.
- `authoriser` sing transactions
- `blobber` acquire new allocations
- `validators` selected to validate challenges.

### LINKED ISSUE #1045: Provider shut down
A provider should be able to deactivate itself. This can use common code associated with a new embedded object (Provider or ShutDownable)

- For each provider type, miner, sharder, blobber, validator and authoriser, we need a new `?_shut_down` transaction. This will set a new `ShutDown` field in the provider object stored in the MPT and mirrored in the event database. 
- If a provider is flagged as `shut_down` it will fail all validation to take on new provider commitments. However, it will still be expected to carry out existing commitments and will receive the full value for rewards and payments for doing so.
- Shutting a provider down is irreversible.
- There is a partial implementation for [miners](https://github.com/0chain/0chain/blob/staging/code/go/0chain.net/smartcontract/minersc/miner.go#L138) and [sharders](https://github.com/0chain/0chain/blob/staging/code/go/0chain.net/smartcontract/minersc/sharder.go#L158) which can be copied and updated to work the new `shut_down` flag https://github.com/0chain/0chain/issues/499.  
- For an overview see https://github.com/0chain/0chain/issues/1040.

### LINKED ISSUE #1046: Kill provder
The owner or other selected user should be able to kill a provider. This can use common code associated with a new embedded object (Provider or Killable)

- For each provider type, miner, sharder, blobber, validator and authoriser, we need a new `kill_?` transaction. This will set a new `Killed` flag in the provider object stored in the MPT and mirrored in the event database. 
- If a provider is flagged as `killed` it will fail all validation to take on new provider commitments. It will also not be provided with any rewards or other paymetns.
- Killing a provider down is irreversible.
- There is a partial implementation for [miners](https://github.com/0chain/0chain/blob/staging/code/go/0chain.net/smartcontract/minersc/miner.go#L138) and [sharders](https://github.com/0chain/0chain/blob/staging/code/go/0chain.net/smartcontract/minersc/sharder.go#L158) which can be copied and updated to work the new `killed` flag https://github.com/0chain/0chain/issues/499.  
- For an overview see https://github.com/0chain/0chain/issues/1040.

### LINKED ISSUE #2122: Blobber slash should be taken from stake pool not rewards.
[stakePool.Slash](https://github.com/0chain/0chain/blob/staging/code/go/0chain.net/smartcontract/storagesc/stakepool.go#L215)
```go
	if err := edbSlash.Emit(event.TagStakePoolReward, balances); err != nil {
		return 0, err
	}
```
This takes slash from rewards. This could result in negative rewards. The slash value would come from the stake pool instead. We need to modify event.DelegatePool to allow moidfy balances.

This code will be modified in both https://github.com/0chain/0chain/pull/1465 and [Blobber and validator, health check kill and shut down](https://github.com/0chain/0chain/pull/1439) so we can do this change as part of one of these or afterwards.

Sriep:
## Fixes
Closes https://github.com/0chain/0chain/issues/1040, Closes https://github.com/0chain/0chain/issues/1044, Closes https://github.com/0chain/0chain/issues/1045, Closes https://github.com/0chain/0chain/issues/1046, Closes https://github.com/0chain/0chain/issues/1020
1. New `health_check_period` setting.
2. Providers can now be deactivated. While deactivated providers will not be available for further commitmetns.
3. Three endpoints for validators and blobbers.
* `Health check` A provider has temporarily deactivated if the time since the last health check is greater than the `health_check_period`.  Health checks are sent from the provider.
* `Shut down` Permianlty deactivates a provider. It can still receive rewards for existing commitments. Only the provider can deactivate itself.
* `Kill` Permianlty deactivates a provider with prejudice. It will no longer be able to receive rewards. Only the 0chain's owner can deactivate a provider. Stakeholders lose a percentage of their stake. New `sc.yaml` settings for this fraction.
4. Shut down and killed blobbers are no longer available for new allocation and are removed from `/blobbers` endpoint.
5. Killed blobbers are removed from partitions used for challenges and block rewards.
6. Killed validators are removed from the partition for challenge validation.
7. New fields to indicate killed and shut down added to providers and stake pools. 
8. Distribute rewards not carried out for killed providers.
9. get blobbers API modified to allow listing of killed or shut down blobbers.
10. get blobber and validator APIs modified to show status
## Tests

- blobber:
- gosdk: https://github.com/0chain/gosdk/pull/574
- system_test: https://github.com/0chain/system_test/pull/333
- zboxcli: https://github.com/0chain/zboxcli/pull/270
- zwalletcli:
- Other: ...
The goSDK and zoxcli PRs only provide access to the new smart contract kill and shut down endpoints. This could be useful for testing.


service-0chain:
| config  | value |
| tests skipped  | FALSE  |
| system-tests  | master  |
| gosdk  | NOT OVERRIDDEN  |
| zboxcli  | staging  |
| zwalletcli  | staging  |
| 0chain  | provider_changes  |
| blobber  | staging  |
| authorizer  | staging  |
| 0box  | staging  |
| 0block  | staging  |
| 0dns  | staging  |
| explorer  | master  |
| 0proxy  | staging  |
| 0search  | staging  |

service-0chain:
| config  | value |
| tests skipped  | FALSE  |
| system-tests  | master  |
| gosdk  | NOT OVERRIDDEN  |
| zboxcli  | staging  |
| zwalletcli  | staging  |
| 0chain  | provider_changes  |
| blobber  | staging  |
| authorizer  | staging  |
| 0box  | staging  |
| 0block  | staging  |
| 0dns  | staging  |
| explorer  | master  |
| 0proxy  | staging  |
| 0search  | staging  |

service-0chain:
| config  | value |
| tests skipped  | FALSE  |
| system-tests  | shutdown-kill-blobbers  |
| gosdk  | kill_and_shutdown  |
| zboxcli  | provider_kill  |
| zwalletcli  | staging  |
| 0chain  | provider_changes  |
| blobber  | staging  |
| authorizer  | staging  |
| 0box  | staging  |
| 0block  | staging  |
| 0dns  | staging  |
| explorer  | staging  |
| 0proxy  | staging  |
| 0search  | staging  |

service-0chain:
| config  | value |
| tests skipped  | FALSE  |
| system-tests  | master  |
| gosdk  | NOT OVERRIDDEN  |
| zboxcli  | staging  |
| zwalletcli  | staging  |
| 0chain  | provider_changes  |
| blobber  | provider_changes  |
| authorizer  | staging  |
| 0box  | staging  |
| 0block  | staging  |
| 0dns  | staging  |
| explorer  | master  |
| 0proxy  | staging  |
| 0search  | staging  |

service-0chain:
| config  | value |
| tests skipped  | FALSE  |
| system-tests  | shutdown-kill-blobbers  |
| gosdk  | kill_and_shutdown  |
| zboxcli  | provider_kill  |
| zwalletcli  | staging  |
| 0chain  | provider_changes  |
| blobber  | staging  |
| authorizer  | staging  |
| 0box  | staging  |
| 0block  | staging  |
| 0dns  | staging  |
| explorer  | staging  |
| 0proxy  | staging  |
| 0search  | staging  |

code/go/0chain.net/smartcontract/dbs/event/blobber.go
```diff
@@ -21,6 +21,12 @@ type Blobber struct {
 	BlobberID string `json:"id" gorm:"uniqueIndex"`
 	BaseURL   string `json:"url" gorm:"uniqueIndex"`
 
+	//provider
+	LastHealthCheck int64 `json:"last_health_check"`
+	IsKilled        bool  `json:"is_killed"`
```
dabasov, line 26: can the same blobber be killed and shutdown at the same time? if not then it'll be better to use one column with status instead

dabasov: most likely we don't need gorm.Model there, since we don't use deletedAt and other fields on here

code/go/0chain.net/smartcontract/dbs/event/blobber.go
```diff
@@ -132,13 +137,15 @@ func (edb *EventDb) BlobberTotalCapacity() (int64, error) {
 	var total int64
 	return total, edb.Store.Get().Model(&Blobber{}).
 		Select("SUM(capacity)").
+		Where("is_killed = ? AND is_shut_down = ?", false, false).
```
dabasov, line 140: these fields has low selectivity and selects might be slow with full scan. we can consider using binary index here, but should test it with benchmarks first

code/go/0chain.net/smartcontract/dbs/event/provider.go
```diff
@@ -0,0 +1,109 @@
...
+		LastHealthCheck int64     `json:"last_health_check"`
+	}
+	model, err := providerModel(providerType)
+	if err != nil {
+		return 0, 0, err
+	}
+	result := edb.Store.Get().
+		Model(&model).
+		Select("down_time, created_at, last_health_check").
+		Where(providerType.String()+"_id = ?", id).
```
dabasov, line 47: we need to remove gorm.Model form Blobber, Sharder etc, since this tables will have id field and ``providerType.String()+"_id`` string

code/go/0chain.net/smartcontract/provider/provider.go
```diff
@@ -0,0 +1,70 @@
...
+	Kill() error
+}
+
+type Provider struct {
+	LastHealthCheck common.Timestamp `json:"last_health_check"`
+	IsShutDown      bool             `json:"is_shut_down"`
+	IsKilled        bool             `json:"is_killed"`
+}
+
+func (p *Provider) Status(now, healthCheckPeriod common.Timestamp) (Status, string) {
```
dabasov, line 58: why do we need to return string here?

code/go/0chain.net/smartcontract/stakepool/stakepool.go
```diff
@@ -345,3 +349,26 @@ func (sp *StakePool) equallyDistributeRewards(coins currency.Coin, spUpdate *Sta
 
 	return nil
 }
+
+func (sp *StakePool) SlashFraction(
```
dabasov, line 351: is it part of another PR?

code/go/0chain.net/smartcontract/storagesc/allocation.go
```diff
@@ -1583,6 +1582,14 @@ func (sc *StorageSmartContract) finishAllocation(
 			return common.NewErrorf("fini_alloc_failed",
 				"blobber %s and %s don't match", b.ID, d.BlobberID)
 		}
+		if b.IsKilled {
```
dabasov: should we add this check to every allocation action (get blobbers, create allocation and so on?)

code/go/0chain.net/smartcontract/storagesc/blobber_reward_partition.go
```diff
@@ -37,14 +37,52 @@ func BlobberRewardKey(round int64) datastore.Key {
...
+
 // getActivePassedBlobberRewardsPartitions gets blobbers passed challenge from last challenge period
 func getActivePassedBlobberRewardsPartitions(balances c_state.StateContextI, period int64) (*partitions.Partitions, error) {
 	name := BlobberRewardKey(GetPreviousRewardRound(balances.GetBlock().Round, period))
-	return partitions.CreateIfNotExists(balances, name, blobberRewardsPartitionSize)
+	blobbers, err := partitions.CreateIfNotExists(balances, name, blobberRewardsPartitionSize)
+	if err != nil {
+		return nil, err
+	}
+	blobbers.SetCallback(passedCallback)
```
dabasov, line 65: why do we introduce side effect here in get method? we need to rename this method or remove side effect

code/go/0chain.net/smartcontract/storagesc/handler.go
```diff
@@ -92,7 +95,104 @@ func GetEndpoints(rh rest.RestHandlerI) []rest.Endpoint {
...
+// parameters:
+//    + name: id
+//      description: validator id
+//      in: query
+//      type: string
+//
+// responses:
+//  200: StatusInfo
+//  400:
+func (srh *StorageRestHandler) getValidatorStatus(w http.ResponseWriter, r *http.Request) {
```
dabasov, line 135: can we put this requests into the getProvider?

code/go/0chain.net/smartcontract/storagesc/health_check.go
```diff
@@ -0,0 +1,81 @@
...
+	balances cstate.StateContextI,
+) (string, error) {
+	var blobber *StorageNode
+	var err error
+	if blobber, err = ssc.getBlobber(t.ClientID, balances); err != nil {
+		return "", common.NewError("blobber_health_check_failed",
+			"can't get the blobber "+t.ClientID+": "+err.Error())
+	}
+	lastHealthCheck := blobber.LastHealthCheck
+	blobber.HealthCheck(t.CreationDate)
```
dabasov, line 24: use ``balances.Now`` instead

code/go/0chain.net/smartcontract/storagesc/health_check.go
```diff
@@ -0,0 +1,81 @@
...
+) (string, error) {
+	var validator = &ValidationNode{
+		ID: t.ClientID,
+	}
+	if err := balances.GetTrieNode(validator.GetKey(ssc.ID), validator); err != nil {
+		return "", common.NewError("blobber_health_check_failed",
+			"can't get the blobber "+t.ClientID+": "+err.Error())
+	}
+
+	validator.HealthCheck(t.CreationDate)
```
dabasov, line 67: use ``balances.Now()`` instead

code/go/0chain.net/smartcontract/storagesc/kill.go
```diff
@@ -0,0 +1,179 @@
...
+	if err := req.decode(input); err != nil {
+		return "", common.NewError("kill_blobber_failed", err.Error())
+	}
+
+	var blobber *StorageNode
+	if blobber, err = ssc.getBlobber(req.ID, balances); err != nil {
+		return "", common.NewError("kill_blobber_failed",
+			"can't get the blobber "+t.ClientID+": "+err.Error())
+	}
+	blobber.Kill()
```
dabasov: check if blobber is already killed

code/go/0chain.net/smartcontract/storagesc/kill.go
```diff
@@ -0,0 +1,179 @@
...
+	if err = emitUpdateBlobber(blobber, balances); err != nil {
+		return "", common.NewError("kill_blobber_failed", err.Error())
+	}
+
+	if blobber.LastRewardPartition.valid() {
```
dabasov: can you please elaborate on this logic

code/go/0chain.net/smartcontract/storagesc/kill.go
```diff
@@ -0,0 +1,179 @@
...
+				"cannot get all blobbers list: "+err.Error())
+		}
+		err = activePassedBlobberRewardPart.RemoveItem(balances, blobber.LastRewardPartition.Index, blobber.ID)
+		if err != nil {
+			return "", common.NewError("kill_blobber_failed",
+				"cannot remove blobber from active passed rewards partition: "+err.Error())
+		}
+	}
+
+	if blobber.RewardPartition.valid() {
```
dabasov: can you please elaborate on this logic

code/go/0chain.net/smartcontract/storagesc/kill.go
```diff
@@ -0,0 +1,179 @@
...
+
+	var validator = &ValidationNode{
+		ID: req.ID,
+	}
+
+	if err = balances.GetTrieNode(validator.GetKey(ssc.ID), validator); err != nil {
+		return "", common.NewError("kill_validator_failed",
+			"can't get the blobber "+t.ClientID+": "+err.Error())
+	}
+	validator.Kill()
```
dabasov: check if it is already killed

code/go/0chain.net/smartcontract/storagesc/kill.go
```diff
@@ -0,0 +1,179 @@
...
+	input []byte,
+	balances cstate.StateContextI,
+) (string, error) {
+	var err error
+	var conf *Config
+	if conf, err = ssc.getConfig(balances, true); err != nil {
+		return "", common.NewError("kill_validator_failed",
+			"can't get config: "+err.Error())
+	}
+	if err := smartcontractinterface.AuthorizeWithOwner("kill_validator_failed", func() bool {
```
dabasov: probably we can move common code from killBlobber and killValidator to some common func

code/go/0chain.net/smartcontract/storagesc/shut_down.go
```diff
@@ -0,0 +1,83 @@
...
+	if blobber, err = ssc.getBlobber(req.ID, balances); err != nil {
+		return "", common.NewError("shut_down_blobber_failed",
+			"can't get the blobber "+t.ClientID+": "+err.Error())
+	}
+	if t.ClientID != blobber.StakePoolSettings.DelegateWallet {
+		return "", common.NewError("shut_down_blobber_failed",
+			"access denied, allowed for delegate_wallet owner only")
+	}
+
+	blobber.ShutDown()
```
dabasov: check if it is already shut down

code/go/0chain.net/smartcontract/storagesc/shut_down.go
```diff
@@ -0,0 +1,83 @@
...
+			"can't save blobber: "+err.Error())
+	}
+	return "", nil
+}
+
+func (ssc *StorageSmartContract) shutDownValidator(
+	t *transaction.Transaction,
+	input []byte,
+	balances cstate.StateContextI,
+) (string, error) {
```
dabasov, line 49: generalize common code

code/go/0chain.net/smartcontract/storagesc/shut_down.go
```diff
@@ -0,0 +1,83 @@
+package storagesc
+
+import (
+	cstate "0chain.net/chaincore/chain/state"
+	"0chain.net/chaincore/transaction"
+	"0chain.net/core/common"
+)
+
+func (ssc *StorageSmartContract) shutDownBlobber(
```
dabasov: how is this status used

dabasov (changes requested):

code/go/0chain.net/smartcontract/dbs/event/blobber.go
Sriep:
```
can the same blobber be killed and shutdown at the same time? 
```
Unspecified, but as of this moment killed is a stronger version of shutdown and we could replace with an enumeration.
```
neither_shutdown_or_killed -> could be active or inactive
shutdown -> always inactive
killed -> always inactive
```
An issue here is that the designation that is used in the code is (active, inactive), which can change silently as global settings and provider settings change. So nothing quite fits. Which takes more space, two `bools` or an `int`?

Sriep:
```
most likely we don't need gorm.Model there, since we don't use deletedAt and other fields on here
```
We will need to refactor the event database in several ways. Hopefully, load tests will help indicate areas where we have problems. But this should be for a separate PR.

code/go/0chain.net/smartcontract/dbs/event/provider.go
Sriep: I created an issue https://github.com/0chain/0chain/issues/1644 for brainstorming ways to increase event database efficiency.

code/go/0chain.net/smartcontract/provider/provider.go
Sriep: The string gives the reason why the status is as it is.

We might have a separate function to get the reason, however in some cases finding the status could take some time, and I see no reason to repeat this.

I guess I think of Status here as a virtual function which is why I am keeping the same signature across all versions of Status. Maybe too much C++ in my youth.

code/go/0chain.net/smartcontract/stakepool/stakepool.go
Sriep: No. When we kill a provider, we slash half (or whatever) of the stake pool's funds.

code/go/0chain.net/smartcontract/storagesc/allocation.go
Sriep: Unspecified. 

I am not sure how we are going to support killing outside 0chian. For example, maybe we should always pair a kill_blobber transaction with a update_allocation transaction that removes the blobber. Which would make it up to the user to handle this, but what happens to the files? In this context, the user is always the chain owner.  Anyway, I am keeping it flexible and considering these changes iterative.

code/go/0chain.net/smartcontract/storagesc/blobber_reward_partition.go
Sriep: The reason for this is that we now want to remove blobbers from partitions if they have been killed. To do this, we have to keep the partition balanced, which means moving another blobber into the vacated place in the partition. For this all to work, we need to update the moved blobber with its new position in the partition so that it in turn, can be successfully killed if required.

This is part of partition design, that when we remove an item from a partition, we call a callback function to be called when an item changes position in the partition. So I do not see adding this callback as a side effect but part of the process of creating a partition.

code/go/0chain.net/smartcontract/storagesc/handler.go
Sriep: Providers are spread across three packages; storagesc, zcnsc and minersc. If you are suggesting global endpoints that do the same action for all providers, like `getProviderStatus`, then there is currently nowhere to put them and no directive to create a new place to put them.

I decided having a separate endpoint for each provider type was easiest as we were already most of the way down that path.

code/go/0chain.net/smartcontract/storagesc/health_check.go
Sriep: balance.Now is not available in StateContextI. We discussed this and concluded that t.CreationDate was ok.

code/go/0chain.net/smartcontract/storagesc/health_check.go
Sriep: balance.Now is not available in StateContextI. We discussed this and concluded that t.CreationDate was ok.

code/go/0chain.net/smartcontract/storagesc/kill.go
Sriep: We want to killed partitions to not receive rewards. So we remove them from the partition containing blobbers to be rewarded.

code/go/0chain.net/smartcontract/storagesc/kill.go
Sriep: Killed blobbers are not to receive any more rewards, so need to be removed from associated partitions.

code/go/0chain.net/smartcontract/storagesc/shut_down.go
Sriep: It is used to decide whether to use the provider for any new activities. In the case of a blobber it means no new allocations.
https://github.com/0chain/0chain/blob/provider_changes/code/go/0chain.net/smartcontract/storagesc/models.go#L867

code/go/0chain.net/smartcontract/storagesc/kill.go
Sriep: I can have another try, but I am not convinced that this makes the code more readable.

code/go/0chain.net/smartcontract/storagesc/shut_down.go
Sriep: The only common code is decoding the provided request. We could use more common code, but that would mean adding get and save provider methods. Which would mean using a common format for the MPT key. Maybe a good idea but should we do that here?

service-0chain:
| config  | value |
| tests skipped  | FALSE  |
| system-tests  | master  |
| gosdk  | blobber_kill_shutdown  |
| zboxcli  | staging  |
| zwalletcli  | staging  |
| 0chain  | provider_changes  |
| blobber  | staging  |
| authorizer  | staging  |
| 0box  | staging  |
| 0block  | staging  |
| 0dns  | staging  |
| explorer  | staging  |
| 0proxy  | staging  |
| 0search  | staging  |

service-0chain:
| config  | value |
| tests skipped  | FALSE  |
| system-tests  | master  |
| gosdk  | blobber_kill_shutdown  |
| zboxcli  | staging  |
| zwalletcli  | staging  |
| 0chain  | provider_changes  |
| blobber  | staging  |
| authorizer  | staging  |
| 0box  | staging  |
| 0block  | staging  |
| 0dns  | staging  |
| explorer  | staging  |
| 0proxy  | staging  |
| 0search  | staging  |

code/go/0chain.net/smartcontract/dbs/event/blobber.go
```diff
@@ -21,6 +21,12 @@ type Blobber struct {
 	BlobberID string `json:"id" gorm:"uniqueIndex"`
 	BaseURL   string `json:"url" gorm:"uniqueIndex"`
 
+	//provider
+	LastHealthCheck int64 `json:"last_health_check"`
+	IsKilled        bool  `json:"is_killed" gorm:"index"`
+	IsShutDown      bool  `json:"is_shut_down" gorm:"index"`
```
boddumanohar: postgresql doesn't support index on boolean field. It will fall back to sequential scan. If you would like to index this you might have to set the type as varchar without suported values as 't' or 'f'. that way the index/query works efficiently.

code/go/0chain.net/smartcontract/benchmark/main/config/benchmark.yaml
```diff
@@ -7,7 +7,7 @@ simulation:
   nun_active_sharders: 20
   num_allocations: 300
   num_blobbers: 100
-  num_validators: 10
+  num_validators: 100
```
boddumanohar, line 10: why increase the num_validator?

Sriep: I was not increasing the validator number rather I was equalising the blobber and validator numbers.

code/go/0chain.net/smartcontract/dbs/event/blobber.go
ra0o0f: does it make sense to index a low cardinality column?

service-0chain:
| config  | value |
| tests skipped  | FALSE  |
| system-tests  | master  |
| gosdk  | NOT OVERRIDDEN  |
| zboxcli  | staging  |
| zwalletcli  | staging  |
| 0chain  | provider_changes  |
| blobber  | staging  |
| authorizer  | staging  |
| 0box  | staging  |
| 0block  | staging  |
| 0dns  | staging  |
| explorer  | master  |
| 0proxy  | staging  |
| 0search  | staging  |

sagchain:
@Sriep What are the next steps and ETA on this? It seems there are many linked issues and this seems to be an important change, can you please provide an update here?

Sriep:
Waiting for me to be reassigned back. Actions left involve merging in changes from staging and resolving any review issues.

Sriep:
Replaced with https://github.com/0chain/0chain/pull/1946

PR #1441, opened 2022-06-27
Feat/search endpoint

### LINKED ISSUE #1480: Generic search API for clients
The API accepts query as parameter and return eventsdb data if query is one of the following:
* BlockHash
* Round no
* transaction_hash,
* file_name
* content hash
* Wallet address

Requirement for: https://github.com/0chain/web-apps/issues/473, https://github.com/0chain/web-apps/issues/472

lawRathod:
## Changes
* Adds new search endpoint on sharder for generic search query

Closes https://github.com/0chain/web-apps/issues/473
Closes https://github.com/0chain/web-apps/issues/472
Closes https://github.com/0chain/0chain/issues/1480

## Tests

- blobber:
- gosdk:
- system_test:
- zboxcli:
- zwalletcli:
- Other: ...

### DIFF
--- a/code/go/0chain.net/smartcontract/dbs/event/search.go
+++ b/code/go/0chain.net/smartcontract/dbs/event/search.go
@@ -1,10 +1,19 @@
 package event
 
-import "database/sql"
+import (
+	"database/sql"
+	"strconv"
+)
 
 // GetGenericSearchType returns kind of search query
 func (edb *EventDb) GetGenericSearchType(query string) (string, error) {
 	var queryType string
+	var round int
+
+	round, err := strconv.Atoi(query)
+	if err != nil {
+		round = -1
+	}
 
 	res := edb.Store.Get().Raw(
 		"SELECT "+
@@ -15,7 +24,7 @@ func (edb *EventDb) GetGenericSearchType(query string) (string, error) {
 			"THEN 'BlockHash' "+
 			"WHEN EXISTS (SELECT Id FROM USERS WHERE USER_ID = @query) "+
 			"THEN 'UserId' "+
-			"WHEN EXISTS (SELECT Id FROM BLOCKS WHERE ROUND = @query) "+
+			"WHEN EXISTS (SELECT Id FROM BLOCKS WHERE ROUND = @round) "+
 			"THEN 'BlockRound' "+
 			"WHEN EXISTS (SELECT Id FROM WRITE_MARKERS WHERE CONTENT_HASH = @query) "+
 			"THEN 'ContentHash' "+
@@ -24,6 +33,7 @@ func (edb *EventDb) GetGenericSearchType(query string) (string, error) {
 			"ELSE 'Not Found' "+
 			"END AS queryType",
 		sql.Named("query", query),
+		sql.Named("round", round),
 	).Scan(&queryType)
 
 	return queryType, res.Error
--- a/code/go/0chain.net/smartcontract/dbs/event/search.go
+++ b/code/go/0chain.net/smartcontract/dbs/event/search.go
@@ -2,6 +2,7 @@ package event
 
 import (
 	"database/sql"
+	"fmt"
 	"strconv"
 )
 
@@ -10,30 +11,42 @@ func (edb *EventDb) GetGenericSearchType(query string) (string, error) {
 	var queryType string
 	var round int
 
+	roundSearchQueryString := ""
 	round, err := strconv.Atoi(query)
-	if err != nil {
-		round = -1
+	if err == nil {
+		roundSearchQueryString = fmt.Sprintf(
+			`WHEN EXISTS (
+				SELECT Id FROM BLOCKS WHERE ROUND = %d
+			) THEN 'BlockRound'`, round,
+		)
 	}
 
 	res := edb.Store.Get().Raw(
-		"SELECT "+
-			"CASE "+
-			"WHEN EXISTS (SELECT Id FROM TRANSACTIONS WHERE HASH = @query) "+
-			"THEN 'TransactionHash' "+
-			"WHEN EXISTS (SELECT Id FROM BLOCKS WHERE HASH = @query) "+
-			"THEN 'BlockHash' "+
-			"WHEN EXISTS (SELECT Id FROM USERS WHERE USER_ID = @query) "+
-			"THEN 'UserId' "+
-			"WHEN EXISTS (SELECT Id FROM BLOCKS WHERE ROUND = @round) "+
-			"THEN 'BlockRound' "+
-			"WHEN EXISTS (SELECT Id FROM WRITE_MARKERS WHERE CONTENT_HASH = @query) "+
-			"THEN 'ContentHash' "+
-			"WHEN EXISTS (SELECT Id FROM WRITE_MARKERS WHERE NAME = @query) "+
-			"THEN 'FileName' "+
-			"ELSE 'Not Found' "+
-			"END AS queryType",
+		fmt.Sprintf(
+			`
+			SELECT 
+			CASE 
+			%s
+			WHEN EXISTS (
+				SELECT Id FROM USERS WHERE USER_ID = @query
+			) THEN 'UserId' 
+			WHEN EXISTS (
+				SELECT Id FROM TRANSACTIONS WHERE HASH = @query
+			) THEN 'TransactionHash' 
+			WHEN EXISTS (
+				SELECT Id FROM BLOCKS WHERE HASH = @query
+			) THEN 'BlockHash' 
+			WHEN EXISTS (
+				SELECT Id FROM WRITE_MARKERS WHERE CONTENT_HASH = @query
+			) THEN 'ContentHash' 
+			WHEN EXISTS (
+				SELECT Id FROM WRITE_MARKERS WHERE NAME = @query
+			) THEN 'FileName'
+			ELSE 'Not Found' 
+			END AS queryType
+		`, roundSearchQueryString,
+		),
 		sql.Named("query", query),
-		sql.Named("round", round),
 	).Scan(&queryType)
 
 	return queryType, res.Error
--- a/code/go/0chain.net/smartcontract/dbs/event/search.go
+++ b/code/go/0chain.net/smartcontract/dbs/event/search.go
@@ -1,0 +1,30 @@
+package event
+
+import "database/sql"
+
+// GetGenericSearchType returns kind of search query
+func (edb *EventDb) GetGenericSearchType(query string) (string, error) {
+	var queryType string
+
+	res := edb.Store.Get().Raw(
+		"SELECT "+
+			"CASE "+
+			"WHEN EXISTS (SELECT Id FROM TRANSACTIONS WHERE HASH = @query) "+
+			"THEN 'TransactionHash' "+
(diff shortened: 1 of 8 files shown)


code/go/0chain.net/chaincore/chain/state_handler.go
```diff
@@ -139,6 +141,29 @@ func (c *Chain) GetNodeFromSCState(ctx context.Context, r *http.Request) (interf
 	return retObj, nil
 }
 
+/*GetSearchHandler - Get result based on query, query could be blockHash, transactionHash, wallet address */
+func (c *Chain) GetSearchHandler(ctx context.Context, r *http.Request) (interface{}, error) {
```
dabasov: @lawRathod I meant this question, to define the type of value we request, then we can switch in golang and query needed entity, it will guarantee 2 queries no matter how much cases we have
```sql
SELECT 
  CASE 
  WHEN EXISTS (SELECT Id FROM TRANSACTIONS WHERE TRANSACTIONS.HASH = 'ab17d7fd0108b706e6f7e44707c4acbbba8ff50e55c5c91ac0552080b71c5f92')
       THEN 'T' 
  WHEN EXISTS (SELECT Id FROM BLOCKS WHERE BLOCKS.HASH = 'ab17d7fd0108b706e6f7e44707c4acbbba8ff50e55c5c91ac0552080b71c5f92')
       THEN 'B' 
  WHEN EXISTS (SELECT Id FROM CLIENTS WHERE CLIENTS.ID = 'ab17d7fd0108b706e6f7e44707c4acbbba8ff50e55c5c91ac0552080b71c5f92')
       THEN 'C' 
  END AS NewFiled  
  ```

**Review by dabasov — APPROVED** [review]
lgtm

**Review by dabasov — CHANGES_REQUESTED** [review]
lgtm, do not merge until clients table is added

File: code/go/0chain.net/chaincore/chain/state_handler.go (hunk 1 above)
Comment (lawRathod) [reply code/go/0chain.net/chaincore/chain/state_handler.go]: I have moved the endpoint location from `state_handler.go` to `storagesc/handler`, I think the search handler makes more sense on the sharder. What do you think?

File: code/go/0chain.net/smartcontract/dbs/event/search.go [hunk 2]
```diff
@@ -0,0 +1,40 @@
...
+			"WHEN EXISTS (SELECT Id FROM USERS WHERE USER_ID = @query) "+
+			"THEN 'UserId' "+
+			"WHEN EXISTS (SELECT Id FROM BLOCKS WHERE ROUND = @round) "+
+			"THEN 'BlockRound' "+
+			"WHEN EXISTS (SELECT Id FROM WRITE_MARKERS WHERE CONTENT_HASH = @query) "+
+			"THEN 'ContentHash' "+
+			"WHEN EXISTS (SELECT Id FROM WRITE_MARKERS WHERE NAME = @query) "+
+			"THEN 'FileName' "+
+			"ELSE 'Not Found' "+
+			"END AS queryType",
```
peterlimg: hmm, I'm not a fan of putting generic search in sharders, it should be a higher level function that should not be provided directly by the sharders,  otherwise it may cause performance issue for sharders, such as if I provide none exist query string every time, then the sharder will need to check all the tables to see if the query data exists. 

Instead, for the generic search function that the web app needed, we might need to parse the query string first to get the query type and data, and then call specific API to get the data respectively. 

The `query` format could be like the github filters that we can see when we search pull requests or issues, such as `is:open is:pr author:peterlimg`, and for us, the query string could be for example: `block:af204cfd0e3a7dea37d168acbccad252c59d335dabf5647cb55d35f29e5caba2`, `round:1000`, `txn:af204cfd0e3a7dea37d168acbccad252c59d335dabf5647cb55d35f29e5caba2`, etc.

peterlimg (changes requested):

code/go/0chain.net/smartcontract/storagesc/handler.go
```diff
@@ -2234,3 +2235,104 @@ func (srh StorageRestHandler) getBlobber(w http.ResponseWriter, r *http.Request)
...
+		return
+	}
+
+	queryType, err := edb.GetGenericSearchType(query)
+	if err != nil {
+		common.Respond(w, r, nil, common.NewErrInternal(err.Error()))
+		return
+	}
+
+	if queryType == "TransactionHash" {
```
ssardana08: at 1 time it cannot be more than one type
better we use switch case here..

ssardana08 (changes requested):

code/go/0chain.net/smartcontract/dbs/event/search.go
dabasov: > mm, I'm not a fan of putting generic search in sharders, it should be a higher level function that should not be provided directly by the sharders, otherwise it may cause performance issue for sharders, such as if I provide none exist query string every time, then the sharder will need to check all the tables to see if the query data exists.

not sure it is huge performance loss, it is 3 lookups to indexes, should be fast, much faster than aggregate functions for example. @peterlimg I agree with you that searching is not sharders responsibility and we might need some other indexing solution, such as elastic or similar, but this exact query will work fast

code/go/0chain.net/smartcontract/storagesc/handler.go
lawRathod: Done Sumit!

stewartie4 (commented):
are there any tests for this new endpoint? we shouldn't be merging new features without tests

lawRathod:
> are there any tests for this new endpoint? we shouldn't be merging new features without tests

No tests added here but I think there's sharder endpoints tests on system-tests, we can add it for this if we decide to actually keep it.

code/go/0chain.net/smartcontract/dbs/event/search.go
```diff
@@ -0,0 +1,40 @@
...
+
+	round, err := strconv.Atoi(query)
+	if err != nil {
+		round = -1
+	}
+
+	res := edb.Store.Get().Raw(
+		"SELECT "+
+			"CASE "+
+			"WHEN EXISTS (SELECT Id FROM TRANSACTIONS WHERE HASH = @query) "+
```
sum-it08, line 21: identifier truncated issue will be there as hashes are 64 chars

code/go/0chain.net/smartcontract/dbs/event/search.go
```diff
@@ -0,0 +1,40 @@
...
+			"THEN 'TransactionHash' "+
+			"WHEN EXISTS (SELECT Id FROM BLOCKS WHERE HASH = @query) "+
```
sum-it08, line 23: same here

code/go/0chain.net/smartcontract/dbs/event/search.go
```diff
@@ -0,0 +1,40 @@
...
+			"THEN 'BlockHash' "+
+			"WHEN EXISTS (SELECT Id FROM USERS WHERE USER_ID = @query) "+
```
sum-it08, line 25: same here

code/go/0chain.net/smartcontract/dbs/event/search.go
```diff
@@ -0,0 +1,40 @@
...
+			"THEN 'UserId' "+
+			"WHEN EXISTS (SELECT Id FROM BLOCKS WHERE ROUND = @round) "+
+			"THEN 'BlockRound' "+
+			"WHEN EXISTS (SELECT Id FROM WRITE_MARKERS WHERE CONTENT_HASH = @query) "+
```
sum-it08, line 29: same here

code/go/0chain.net/smartcontract/dbs/event/search.go
```diff
@@ -0,0 +1,40 @@
...
+			"THEN 'ContentHash' "+
+			"WHEN EXISTS (SELECT Id FROM WRITE_MARKERS WHERE NAME = @query) "+
```
ssardana08: this is expensive and inconsistent. something to think about

code/go/0chain.net/smartcontract/dbs/event/search.go
```diff
@@ -0,0 +1,40 @@
+package event
+
+import (
+	"database/sql"
+	"strconv"
+)
+
+// GetGenericSearchType returns kind of search query
+func (edb *EventDb) GetGenericSearchType(query string) (string, error) {
```
ssardana08, line 10: `query` is something like the entire command in sql domain..
maybe we should use a better name `searchString`

code/go/0chain.net/smartcontract/dbs/event/search.go
```diff
@@ -0,0 +1,40 @@
...
+	var queryType string
+	var round int
+
+	round, err := strconv.Atoi(query)
+	if err != nil {
```
ssardana08: if it errors, we should not even query in blocks table

code/go/0chain.net/smartcontract/dbs/event/user.go
```diff
@@ -53,17 +54,27 @@ func (edb *EventDb) addOrOverwriteUser(u User) error {
 	return result.Error
 }
 
+func (edb *EventDb) GetUserFromId(userId string) (User, error) {
+	user := User{}
+	return user, edb.Store.Get().Model(&User{}).Where(User{UserID: userId}).Scan(&user).Error
+
+}
+
+func (edb *EventDb) CreateUser(usr *User) error {
```
ssardana08: is this being used anywhere?

code/go/0chain.net/smartcontract/storagesc/handler.go
```diff
@@ -2253,3 +2254,96 @@ func (srh StorageRestHandler) getBlobber(w http.ResponseWriter, r *http.Request)
...
+		return
+	}
+
+	queryType, err := edb.GetGenericSearchType(query)
+	if err != nil {
+		common.Respond(w, r, nil, common.NewErrInternal(err.Error()))
+		return
+	}
+
+	switch queryType {
```
ssardana08, line 2272: there is no case 'FileName' here
which is there in gettingQueryType.. just to make sure we are covering all cases

ssardana08 (changes requested):

dabasov (approved):

code/go/0chain.net/smartcontract/dbs/event/search.go
lawRathod: Yes, I agree, renamed the param!

code/go/0chain.net/smartcontract/dbs/event/search.go
lawRathod: Done

code/go/0chain.net/smartcontract/dbs/event/user.go
lawRathod: I don't remember adding this, could be a result of bad merge/rebase. Removed it since couldn't find it being used anywhere.

code/go/0chain.net/smartcontract/storagesc/handler.go
lawRathod: Good catch! I missed it. Added it.

peterlimg (approved):
lgtm

ssardana08 (approved):

PR #1454, opened 2022-07-01
Miner sharder, kill and shut down

Sriep:
## Fixes
Adds kill and shutdown endpoints for miners and sharders. Related to https://github.com/0chain/0chain/pull/1439
issues: https://github.com/0chain/0chain/issues/1040, https://github.com/0chain/0chain/issues/1044, https://github.com/0chain/0chain/issues/1045, https://github.com/0chain/0chain/issues/1046
## Tests

- blobber:
- gosdk: https://github.com/0chain/gosdk/pull/502
- system_test:
- zboxcli:
- zwalletcli: https://github.com/0chain/zwalletcli/pull/154
- Other: ...


code/go/0chain.net/smartcontract/minersc/kill.go
```diff
@@ -0,0 +1,78 @@
...
+	input []byte,
+	gn *GlobalNode,
+	balances cstate.StateContextI,
+) (string, error) {
+	var id killInput
+	if err := id.decode(input); err != nil {
+		return "", common.NewError("kill-miner", err.Error())
+	}
+
+	mn, err := getMinerNode(id.ID, balances)
```
peterlimg, line 47: We will need to verify if the transaction creator has the right to kill the miner, so need to ensure the transaction.ClientID is the chain owner.

code/go/0chain.net/smartcontract/minersc/kill.go
```diff
@@ -0,0 +1,78 @@
...
+	input []byte,
+	gn *GlobalNode,
+	balances cstate.StateContextI,
+) (string, error) {
+	var id killInput
+	if err := id.decode(input); err != nil {
+		return "", common.NewError("kill-miner", err.Error())
+	}
+
+	sn, err := msc.getSharderNode(id.ID, balances)
```
peterlimg, line 94: same as the killMiner above

code/go/0chain.net/smartcontract/minersc/miner.go
```diff
@@ -15,20 +15,21 @@ import (
...
+) (bool, *MinerNode) {
 	mn := NewMinerNode()
 	err := balances.GetTrieNode(pkey, mn)
 	if err != nil {
 		if err != util.ErrValueNotPresent {
 			logging.Logger.Error("GetTrieNode from state context", zap.Error(err),
 				zap.String("key", pkey))
 		}
-		return false
+		return false, nil
```
peterlimg: The previous logic seems incorrect, if we got 'node not found' error, then we should stop and return instead of continue saving the miner.  

So perhaps we can return 3 variables like `*MinerNode, bool, error`. So if we find the miner, we return `minerNode, true, nil`, If we get 'value not present' error, we return `nil, false, nil`, otherwise, we return `nil, false, err`, and the caller should not continue the work if err is not nil.

code/go/0chain.net/smartcontract/minersc/miner.go
```diff
@@ -120,15 +121,22 @@ func (msc *MinerSmartContract) AddMiner(t *transaction.Transaction,
 
 		update = true
 	}
-
-	if !doesMinerExist(newMiner.GetKey(), balances) {
+	exists, mn := doesMinerExist(newMiner.GetKey(), balances)
+	if !exists {
```
peterlimg, line 122: As per above, if not exist, the error could be 'node not found', which means the MPT is not synced, we should not continue saving.

peterlimg (changes requested):

code/go/0chain.net/smartcontract/provider/provider.go
```diff
@@ -0,0 +1,70 @@
...
+	"0chain.net/core/common"
+)
+
+//go:generate msgp -io=false -tests=false -v
+
+// swagger:enum Status
+type Status int
+
+const (
+	Active Status = iota
```
peterlimg: Perhaps we need to start the Status with 1 to avoid a potential issue that if some day the Status is an event filter, the default value of 0 would return all providers instead of returning `Active` providers. Please check the similar issue we had before https://github.com/0chain/0chain/issues/1381

peterlimg (changes requested):

code/go/0chain.net/smartcontract/provider/provider.go
```diff
@@ -0,0 +1,70 @@
...
+	Kill() error
+}
+
+type Provider struct {
+	LastHealthCheck common.Timestamp `json:"last_health_check"`
+	IsShutDown      bool             `json:"is_shut_down"`
+	IsKilled        bool             `json:"is_killed"`
+}
+
+func (p *Provider) Status(now, healthCheckPeriod common.Timestamp) (Status, string) {
```
peterlimg, line 47: We can return Status only, as we can call `String()` to get string value.

code/go/0chain.net/smartcontract/provider/provider.go
```diff
@@ -0,0 +1,70 @@
...
+	if p.IsKilled {
+		return Killed, Killed.String()
+	}
+	if p.IsShutDown {
+		return ShutDown, ShutDown.String()
+	}
+	if p.LastHealthCheck < (now - healthCheckPeriod) {
+		return Inactive, fmt.Sprintf("\tfailed health check, last check %v", p.LastHealthCheck)
```
peterlimg, line 55: Mix the error message info with the Status string seems not a good idea. The caller can get the `p.LatestHealthCheck` directly, so if the `Inactive` is returned, it can construct the message itself.

peterlimg (changes requested):

code/go/0chain.net/smartcontract/provider/provider.go
Sriep: Weird, so gorm where clauses don't work for filters using the default value?

code/go/0chain.net/smartcontract/provider/provider.go
Sriep: The string value is not the string version of the status, but the reason why the status is not active. For example it could be inactive because
* health check failed
* killed or shout down
* write price is too high due to change in global max write price
* read price is too high due to change to global max read price
* capacity is too low due to change in min blobber capacity
(https://github.com/0chain/0chain/blob/provider_changes/code/go/0chain.net/smartcontract/storagesc/models.go#L386)

For example, once the user knows the reason why his provider has stopped being active, they can fix it.

code/go/0chain.net/smartcontract/provider/provider.go
Sriep: It's not really an error message, it's the reason why the provider is inactive. 

Status here is a kind of a virtual function that individual providers will want to overwrite with their own copy with reasons for inaction that are specific to that particular provider. https://github.com/0chain/0chain/blob/provider_changes/code/go/0chain.net/smartcontract/storagesc/models.go#L386 Its not clear if we need anything for miners here, do we want to inactivate a miner if the [max_charge](https://github.com/0chain/0chain/blob/staging/docker.local/config/sc.yaml#L60) changes and becomes lower than the miner's service charge?

Environmental reason for being inactive as discussed [here](https://github.com/0chain/0chain/issues/1040).

We could have another function, `ReasonForStatus()`, and have `Status()` just return the status?

code/go/0chain.net/smartcontract/minersc/fees.go
```diff
@@ -367,9 +367,7 @@ func (msc *MinerSmartContract) payFees(t *transaction.Transaction,
 			"saving generator node: %v", err)
 	}
 
-	if err = emitUpdateMiner(mn, balances, false); err != nil {
-		return "", common.NewErrorf("pay_fees", "saving generator node to db: %v", err)
-	}
+	emitUpdateMiner(mn, balances, false)
```
peterlimg, line 369: Perhaps log the error if any, and we may need to avoid emitting events for miners as only sharders will need to process the events.

code/go/0chain.net/smartcontract/minersc/miner.go
```diff
@@ -194,7 +175,11 @@ func (msc *MinerSmartContract) deleteNode(
 
 	usp, err := stakepool.GetUserStakePools(nodeType, deleteNode.Settings.DelegateWallet, balances)
 	if err != nil {
-		return nil, fmt.Errorf("can't get user pools list: %v", err)
+		logging.Logger.Info("piers deleteNode", zap.Int("len deletNode.Pools", len(deleteNode.Pools)), zap.Error(err))
```
peterlimg: Remove the log if it's for debugging

code/go/0chain.net/smartcontract/minersc/sc.go
```diff
@@ -129,7 +136,12 @@ func (msc *MinerSmartContract) Execute(t *transaction.Transaction,
 	if !found {
 		return common.NewErrorf("failed execution", "no miner smart contract method with name: %v", funcName).Error(), nil
 	}
-	return scFunc(t, input, gn, balances)
+	resp, err := scFunc(t, input, gn, balances)
+	if funcName != "payFees" {
```
peterlimg: remove later if it's for debugging

code/go/0chain.net/smartcontract/minersc/kill.go
```diff
@@ -0,0 +1,110 @@
...
+	return buff
+}
+
+func (msc *MinerSmartContract) killMiner(
+	txn *transaction.Transaction,
+	input []byte,
+	gn *GlobalNode,
+	balances cstate.StateContextI,
+) (string, error) {
+	logging.Logger.Info("piers killMiner")
```
peterlimg: remove log later if it's for debugging

peterlimg (changes requested):

code/go/0chain.net/smartcontract/minersc/fees.go
Sriep: There is no error as we are not marshalling anything.

code/go/0chain.net/smartcontract/stakepool/stakepool.go
```diff
@@ -326,3 +326,20 @@ func (sp *StakePool) equallyDistributeRewards(coins currency.Coin, spUpdate *Sta
...
 	return nil
 }
+
+func (sp *StakePool) SlashFraction(
+	fraction float64,
+	providerId string,
+	providerType spenum.Provider,
+	balances cstate.StateContextI,
+) error {
+	if fraction == 0.0 {
```
peterlimg, line 341: Perhaps check if the `fraction > 0.0 && fraction <= 1`, otherwise the balance would be negative or 0 after splash.

code/go/0chain.net/smartcontract/minersc/fees.go
```diff
@@ -369,17 +369,17 @@ func (msc *MinerSmartContract) payFees(t *transaction.Transaction,
...
-	if gn.RewardRoundFrequency != 0 && mb.Round%gn.RewardRoundFrequency == 0 {
-		var lfmb = balances.GetLastestFinalizedMagicBlock().MagicBlock
-		if lfmb != nil {
-			err = msc.viewChangePoolsWork(gn, lfmb, mb.Round, balances)
-			if err != nil {
-				return "", err
-			}
-		} else {
-			return "", common.NewError("pay fees", "cannot find latest magic bock")
+	//if gn.RewardRoundFrequency != 0 && mb.Round%gn.RewardRoundFrequency == 0 {
```
peterlimg: We'd better not change the reward round frequency checking, as it will make the miners to get AllMinersList from MPT every time this payFee transaction is executed in the `msc.viewChangePoolSwork`.

peterlimg (changes requested):

code/go/0chain.net/smartcontract/minersc/fees.go
Sriep: Ops! that was just debugging. I wanted to force stepping through the code using the benchmark and that seemed the easiest way.

code/go/0chain.net/smartcontract/minersc/handler.go
```diff
@@ -549,6 +554,12 @@ func (mrh *MinerRestHandler) getSharderList(w http.ResponseWriter, r *http.Reque
 	}
 
 	filter := event.SharderQuery{}
+
+	killed := len(r.URL.Query().Get("killed")) > 0
+	shutDown := len(r.URL.Query().Get("killed")) > 0
```
dabasov, line 559: should it be `shutDown`?

code/go/0chain.net/smartcontract/minersc/handler.go
```diff
@@ -662,8 +756,13 @@ func (mrh *MinerRestHandler) getMinerList(w http.ResponseWriter, r *http.Request
 		common.Respond(w, r, nil, err)
 		return
 	}
-
 	filter := event.MinerQuery{}
+
+	killed := len(r.URL.Query().Get("killed")) > 0
+	shutDown := len(r.URL.Query().Get("killed")) > 0
```
dabasov, line 762: is shutdown?

code/go/0chain.net/smartcontract/minersc/miner.go
```diff
@@ -15,20 +15,22 @@ import (
 	"go.uber.org/zap"
 )
 
-func doesMinerExist(pkey datastore.Key,
-	balances cstate.CommonStateContextI) bool {
-
+func doesMinerExist(
```
dabasov, line 18: getMiner is better, since we return miner instead of bool

dabasov (changes requested):

code/go/0chain.net/smartcontract/minersc/shut_down.go
```diff
@@ -0,0 +1,89 @@
...
+
+func (msc *MinerSmartContract) shutDownMiner(
+	txn *transaction.Transaction,
+	input []byte,
+	gn *GlobalNode,
+	balances cstate.StateContextI,
+) (string, error) {
+	var id idInput
+	if err := id.decode(input); err != nil {
+		return "", common.NewError("kill-miner", err.Error())
```
dabasov, line 19: shut-down-miner

code/go/0chain.net/smartcontract/stakepool/stakepool.go
```diff
@@ -324,3 +328,23 @@ func (sp *StakePool) equallyDistributeRewards(coins currency.Coin, spUpdate *Sta
...
+) error {
+	if fraction < 0.0 || fraction > 1.0 {
+		return fmt.Errorf("kill slash fraction %v, not in interval [0.1]", fraction)
+	}
+	if fraction == 0.0 {
+		return nil
+	}
+	for _, dp := range sp.Pools {
+		var dpSlash = currency.Coin(float64(dp.Balance) * fraction)
+		dp.Balance -= dpSlash
```
dabasov, line 346: should check if balance is not negative

dabasov (changes requested):

sagchain:
@Sriep are you just waiting for review to be done and closed on this?

Sriep:
Closed as requested will redo later

PR #1445, opened 2022-06-29
Add new `/v1/healthcheck` API to sharder

peterlimg:
## Fixes

- #1437 

## Changes

- Added a new `/v1/healthcheck` endpoint that returns basic sharder info, `uptime`, `build tag`, `latest finalized block summary`, etc. 
- Added API logger to log the received api requests and response status and process time.

- Changed the `/_health_check` to `/_healthcheck` to force gosdk users to upgrade sdk so that the new endpoint will be used. 

## Tests
Tasks to complete before merging PR:
- [ ]  Ensure system tests are passing. If not [Run them manually](https://github.com/0chain/0chain/actions/workflows/system_tests.yml) to check for any regressions :clipboard:
- [ ]  Do any new system tests need added to test this change? do any existing system tests need updated? If so create a PR at [0chain/system_test](https://github.com/0chain/system_test)
- [ ]  Merge your system tests PR to master AFTER merging this PR

## Associated PRs (Link as appropriate):
- blobber: https://github.com/0chain/blobber/pull/737
- gosdk: https://github.com/0chain/gosdk/pull/503
- system_test:
- zboxcli: https://github.com/0chain/zboxcli/pull/271
- zwalletcli: https://github.com/0chain/zwalletcli/pull/151
- Other: ...

PR #1456, opened 2022-07-02
Revert "Checking config on loading lfb when start"

peterlimg:
## Fixes

See comments: https://github.com/0chain/0chain/pull/1445#issuecomment-1172779702

This reverts commit 917c90c25f35f7e009227420595e843ab096d3e1 as the config is not saved in MPT, so it would never be able to find config from MPT when loading lfb, therefore it will cause the sharder roll back to genesis block on restarting, which is not what we expect.

## Tests

- blobber:
- gosdk:
- system_test:
- zboxcli:
- zwalletcli:
- Other: ...

### DIFF
--- a/code/go/0chain.net/chaincore/chain/entity.go
+++ b/code/go/0chain.net/chaincore/chain/entity.go
@@ -1335,7 +1335,7 @@ func (c *Chain) getClientState(pb *block.Block) (util.MerklePatriciaTrieI, error
 	return pb.ClientState, nil
 }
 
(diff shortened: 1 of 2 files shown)


dabasov:
I can't get why lfb can have no config node, as soon as config is first accessed it should be stored in the db, sharder should not forget nodes, so MPT should always contain the config. @peterlimg am I missing something?

dabasov (approved):
lgtm

cnlangzi (approved):

Issue #1210, opened 2022-04-27
Slow MPT iterating


dabasov:
This problem happens after hightps-2 worked more than a week, causing it to stuck
```
2022-04-27T13:33:50.616Z	ERROR	chain/worker.go:515	Sync round state abort, context timed out for checking missing nodes
0chain.net/chaincore/chain.(*Chain).syncRoundStateToStateDB
	/0chain/go/0chain.net/chaincore/chain/worker.go:515
0chain.net/chaincore/chain.(*Chain).SyncLFBStateWorker.func2
	/0chain/go/0chain.net/chaincore/chain/worker.go:465
2022-04-27T13:33:50.616Z	ERROR	util/merkle_patricia_trie.go:986	Find missing node with unexpected err	{"error": "context deadline exceeded"}
0chain.net/core/util.(*MerklePatriciaTrie).FindMissingNodes
	/0chain/go/0chain.net/core/util/merkle_patricia_trie.go:986
0chain.net/chaincore/chain.(*Chain).syncRoundStateToStateDB
	/0chain/go/0chain.net/chaincore/chain/worker.go:508
0chain.net/chaincore/chain.(*Chain).SyncLFBStateWorker.func2
	/0chain/go/0chain.net/chaincore/chain/worker.go:465
2022-04-27T13:33:50.616Z	ERROR	util/merkle_patricia_trie.go:808	iterate - child node	{"error": "context deadline exceeded"}
0chain.net/core/util.(*MerklePatriciaTrie).iterate
	/0chain/go/0chain.net/core/util/merkle_patricia_trie.go:808
0chain.net/core/util.(*MerklePatriciaTrie).Iterate
	/0chain/go/0chain.net/core/util/merkle_patricia_trie.go:292
0chain.net/core/util.(*MerklePatriciaTrie).FindMissingNodes
	/0chain/go/0chain.net/core/util/merkle_patricia_trie.go:980
0chain.net/chaincore/chain.(*Chain).syncRoundStateToStateDB
	/0chain/go/0chain.net/chaincore/chain/worker.go:508
0chain.net/chaincore/chain.(*Chain).SyncLFBStateWorker.func2
	/0chain/go/0chain.net/chaincore/chain/worker.go:465
```

peterlimg:
The iteration will always timeout when the database become large, we need to have a better way to detect missing states.

peterlimg:
[...]

PR #1457, opened 2022-07-03
Fix sharder stuck

peterlimg:
## Fixes

- Fix a hidden cause of MPT 'node not found' error that stuck the sharders. The MPT `FullNode` `ExtensionNode` does not include the version number when calculation the node hash key. Suppose we have a `FullNode` that was removed in round 100 (not removed from db yet, happens later in client state prune worker), then in round 200, we added a new full node that has the same value, children nodes, those two `FullNode` will have the same node hash key. So when the prune client state worker start to work, the `FullNode` will be removed from db, and we will see 'node not found' error if we try to get that node from the round 200. The fix is to include the version number when calculating the node hash, so the `FullNode` in round 200 will not have the same key as it in round 100.
- Fixed a bug in sharder sync process that stuck the chain moving. This happens when block process was waiting for LFB moving as it reached the LFB+ahead round, while finalize block process was waiting for new block processed and push to the finalize channel, therefore, a dead lock happened. Fixed this by pushing back the block that is ahead of LFB+5 to the process channel so that it has the chance to be processed later.
- Implemented the MPT state validation on starting of sharders, it would roll back from latest finalized block till the one that has validate full state. 

## Changes
- Use `dead_nodes` to do client state pruning process #1210 

## Tests

- blobber:
- gosdk:
- system_test:
- zboxcli:
- zwalletcli:
- Other: ...


code/go/0chain.net/chaincore/chain/entity.go
```diff
@@ -167,6 +167,8 @@ type Chain struct {
 	magicBlockSaver              MagicBlockSaver
 
 	pruneStats *util.PruneStats
+	// channel to trigger the worker to start prunce process immediately
```
dabasov: typo

code/go/0chain.net/chaincore/chain/protocol_round.go
```diff
@@ -136,9 +136,9 @@ func (c *Chain) FinalizeRoundImpl(r round.RoundI) {
 	if !r.SetFinalizing() {
 		logging.Logger.Debug("finalize_round: already finalizing",
 			zap.Int64("round", r.GetRoundNumber()))
-		if node.Self.Type == node.NodeTypeSharder {
-			return
-		}
+		//if node.Self.Type == node.NodeTypeSharder {
```
dabasov: dead code, can't get why we returfo sn harder previously

code/go/0chain.net/core/common/rate_limter.go
```diff
@@ -47,7 +48,11 @@ func UserRateLimit(handler ReqRespHandlerf) ReqRespHandlerf {
 		return Recover(handler)
 	}
 	return func(writer http.ResponseWriter, request *http.Request) {
-		tollbooth.LimitFuncHandler(userRateLimit.Limiter, Recover(handler)).ServeHTTP(writer, request)
+		if strings.HasPrefix(request.URL.Path, "/_") {
```
dabasov: is it for `_diagnostics` to bypass limiter?

dabasov:
just found that we don't use MPT.UpdateVersion(). I wonder how do we update version of node then?

code/go/0chain.net/core/common/rate_limter.go
peterlimg: Yes, was for bypassing the diagnostic page for debugging, can be removed.

code/go/0chain.net/chaincore/chain/protocol_round.go
peterlimg: > just found that we don't use MPT.UpdateVersion(). I wonder how do we update version of node then?

We don't need to update version of node now. We just record all the deleted nodes' `(key, version)`  in `dead_nodes` in PNode when finalize a block. When enough dead nodes are collected or the prune period hits, we will check the `dead_nodes` and remove nodes that has version below specific number from db.

dabasov (approved):
lgtm

PR #1458, opened 2022-07-03
added free storage recipient check

dabasov:
## Changes
Sender of transaction to create free storage can only be marker recipient


peterlimg (approved):
lgtm

PR #1476, opened 2022-07-07
Fix sharder blocks sync issue that caused the chain stuck

peterlimg:
## Fixes
- Re-finalize the rounds if the processing block round is >= LFB+aheadN when syncing blocks. The chain would not move on if the finalize process stuck due to some reason.
- OOM caused by the [healthcheck](https://github.com/0chain/0chain/blob/ecc4c2fc7ad1ed95249e10d505aabf342c03e1fe/code/go/0chain.net/sharder/health_check.go#L190) worker on sharders 

## Changes
- Save dead nodes in gorocksdb column familiy `dead_nodes` and use iterator to check the nodes that need to be removed so that we don't need to load all dead nodes in RAM at once. This also takes advantage of the gorocksdb's iterator feature that it is sorted naturally by the key, so we can store the key as the dead nodes round number, and value as the node keys in MPT, then we only need to iterator from the iterator.First() (with the oldest round number) to the round that is greater than the `LFB - PruneStateBelowCount()`. 
- Records the dead nodes with version number as of the finalizing block round number. Previously, we set it as the node's origin version, i.e the round number when the node was created, which is incorrect and will lead to state being removed even it is just one round below the current round. 
- Sharders sync LFB from remote on start to check if the local LFB is correct
- Sharders check if the local LFB could be rollback if its state is invalid and panic if rollback below `PruneBelowCount` as all the state below that round are not validated (was pruned).

## Tests

No


dabasov (approved):
lgtm, I didn't like this 50 rounds scroll back from the beginning, awesome job

PR #1483, opened 2022-07-08
Remove non-required blobber fetch and updating in commit_connection

ssardana08:
## Tests

- blobber:
- gosdk:
- system_test:
- zboxcli:
- zwalletcli:
- Other: ...

### DIFF
--- a/code/go/0chain.net/smartcontract/storagesc/blobber.go
+++ b/code/go/0chain.net/smartcontract/storagesc/blobber.go
@@ -661,19 +661,6 @@ func (sc *StorageSmartContract) commitBlobberConnection(
 	alloc.Stats.UsedSize += commitConnection.WriteMarker.Size
 	alloc.Stats.NumWrites++
 
-	// UpdateItem saved_data on storage node
-	var storageNode *StorageNode
-	if _, ok := alloc.BlobberAllocsMap[commitConnection.WriteMarker.BlobberID]; ok {
-		storageNode, err = sc.getBlobber(commitConnection.WriteMarker.BlobberID, balances)
-		if err != nil {
-			return "", common.NewError("commit_connection_failed",
-				"can't get blobber")
-		}
-	}
-	m.tick("get_writemarker_blobber")
-
-	storageNode.SavedData += alloc.Stats.UsedSize
-
 	// check time boundaries
 	if commitConnection.WriteMarker.Timestamp < alloc.StartTime {
 		return "", common.NewError("commit_connection_failed",
--- a/code/go/0chain.net/smartcontract/storagesc/blobber.go
+++ b/code/go/0chain.net/smartcontract/storagesc/blobber.go
@@ -4,6 +4,7 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"time"
 
 	cstate "0chain.net/chaincore/chain/state"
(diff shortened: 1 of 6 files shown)


code/go/0chain.net/smartcontract/storagesc/sc.go
```diff
@@ -183,7 +183,7 @@ func (sc *StorageSmartContract) Execute(t *transaction.Transaction,
 		}
 
 	case "commit_connection":
-		resp, err = sc.commitBlobberConnection(t, input, balances)
+		resp, err = sc.commitBlobberConnection(t, input, balances, nil)
```
Sriep, line 186: Can we keep the signature of endpoints consistent? So `(t *transaction.Transaction, input []byte, balances cstate.StateContextI) (string error)`. 

We can use a local function, if convenient, to add the extra parameter.

code/go/0chain.net/smartcontract/storagesc/blobber.go
```diff
@@ -563,25 +564,30 @@ func (sc *StorageSmartContract) commitMoveTokens(alloc *StorageAllocation,
 }
 
 func (sc *StorageSmartContract) commitBlobberConnection(
-	t *transaction.Transaction, input []byte, balances cstate.StateContextI) (
+	t *transaction.Transaction, input []byte, balances cstate.StateContextI, timings map[string]time.Duration) (
 	string, error) {
 
+	m := Timings{timings: timings, start: time.Now(), fnName: "commit_connection"}
```
Sriep, line 570: What are all the timings about?

Sriep (commented):
What is this PR doing? Does it do anything other than refactoring variable names and adding timmings?

code/go/0chain.net/smartcontract/storagesc/blobber.go
ssardana08: @Sriep this is for timing benchmarks

code/go/0chain.net/smartcontract/storagesc/blobber.go
```diff
@@ -651,18 +661,6 @@ func (sc *StorageSmartContract) commitBlobberConnection(
 	alloc.Stats.UsedSize += commitConnection.WriteMarker.Size
 	alloc.Stats.NumWrites++
 
-	// UpdateItem saved_data on storage node
```
ssardana08, line 654: @Sriep this is the only change we have in this PR.. rest are refactoring

ssardana08: this was a non-required step which is now deleted

peterlimg: So we will get the total saved data from event db? If the `SavedData` will be used by smart contracts for other calculation, then perhaps we need to keep it updated, otherwise, we can remove it.

code/go/0chain.net/smartcontract/storagesc/blobber.go
peterlimg: @sum-it08 let's not add too many of those benchmark time logs, this could be added when for debugging or doing benchmark tests but should not go to production.

peterlimg (changes requested):

ssardana08:
I'll raise a new PR with only dedup changes.
Closing this

PR #1501, opened 2022-07-15
WIP: Fix/tests with sharders sync fix

fabioDMFerreira:
Check whether sharder sync fixes no-view-change.fault-tolerance tests


fabioDMFerreira:
This was only an experiment to check whether sharder sync changes have fixed some tests. They did!

PR #1510, opened 2022-07-18
Fix get client state API failure on miners to fix fault tolerant byzantine error

peterlimg:
## Fixes

Miners and sharders would not be able to sync the transaction nonce after restarting, hence unable to create new transactions, generate blocks, etc and failed the fault tolerant byzantine tests.

## Changes

- Sync client nonce from sharders

- Extend the wait time out time of fault tolerant tests, miners need more time to recover and sync missing state before they could work.

## Tests

- blobber:
- gosdk:
- system_test:
- zboxcli:
- zwalletcli:
- Other: ...

### DIFF
--- a/code/go/0chain.net/chaincore/chain/protocol_view_change.go
+++ b/code/go/0chain.net/chaincore/chain/protocol_view_change.go
@@ -277,7 +277,7 @@ func (c *Chain) RegisterNode() (*httpclientutil.Transaction, error) {
 	mb := c.GetCurrentMagicBlock()
 	var minerUrls = mb.Miners.N2NURLs()
 	logging.Logger.Debug("Register nodes to", zap.Strings("urls", minerUrls))
-	err = httpclientutil.SendSmartContractTxn(txn, minersc.ADDRESS, 0, 0, scData, minerUrls)
+	err = httpclientutil.SendSmartContractTxn(txn, minersc.ADDRESS, 0, 0, scData, minerUrls, mb.Sharders.N2NURLs())
 	return txn, err
 }
 
@@ -306,7 +306,7 @@ func (c *Chain) RegisterSharderKeep() (result *httpclientutil.Transaction, err2 
 	txn.PublicKey = selfNode.PublicKey
 	mb := c.GetCurrentMagicBlock()
 	var minerUrls = mb.Miners.N2NURLs()
-	err := httpclientutil.SendSmartContractTxn(txn, minersc.ADDRESS, 0, 0, scData, minerUrls)
+	err := httpclientutil.SendSmartContractTxn(txn, minersc.ADDRESS, 0, 0, scData, minerUrls, mb.Sharders.N2NURLs())
 	return txn, err
 }
 
--- a/code/go/0chain.net/chaincore/chain/state_handler.go
+++ b/code/go/0chain.net/chaincore/chain/state_handler.go
@@ -1,12 +1,6 @@
 package chain
 
 import (
-	"0chain.net/smartcontract/faucetsc"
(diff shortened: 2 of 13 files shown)


dabasov:
@peterlimg why can't we switch getBalance completely to sharders?

peterlimg:
@dabasov the main reason is that I will need to change the [SendSmartContractTxn](https://github.com/0chain/0chain/blob/4f4f4af566256307530e45428c311810f4f0efdc/code/go/0chain.net/chaincore/httpclientutil/http_client_util.go#L731)  function signature to add `shardersUrl` param and change all the places where this function is called. Comparing to the easy way that just add a few line code for miners, I decided to go the easier way :)

Another reason is as I said above in the PR description, miners number is much larger than the sharders, and they needs to sync nonce each time before creating a transaction, so sharders will under DDoS for syncing nonce. Anyway, I checked the `MakeClientStateRequest` function again, seems it would cause the DDoS for miners as well with current implementation. So perhaps change to sharders would reduce the requests number. I'll change to sharders and fix the DDoS issue later.

code/go/0chain.net/miner/protocol_block.go
```diff
@@ -162,6 +162,8 @@ func (mc *Chain) validateTransaction(b *block.Block, bState util.MerklePatriciaT
 	}
 
 	if txn.Nonce-state.Nonce < 1 {
+		logging.Logger.Debug("past txn", zap.String("txn", txn.Hash),
```
dabasov: we probably need to check debugTxn for this log, since there can be a lot of such logs under the load

code/go/0chain.net/miner/protocol_block.go
```diff
@@ -635,11 +637,11 @@ func txnProcessorHandlerFunc(mc *Chain, b *block.Block) txnProcessorHandler {
 		switch err {
 		case PastTransaction:
 			tii.pastTxns = append(tii.pastTxns, txn)
-			if debugTxn {
-				logging.Logger.Info("generate block (debug transaction) error, transaction hash old nonce",
-					zap.String("txn", txn.Hash), zap.Int32("idx", tii.idx),
-					zap.Any("now", common.Now()), zap.Int64("nonce", txn.Nonce))
-			}
+			//if debugTxn {
```
dabasov: we better uncomment this check

code/go/0chain.net/miner/protocol_block.go
```diff
@@ -652,15 +654,17 @@ func txnProcessorHandlerFunc(mc *Chain, b *block.Block) txnProcessorHandler {
 				return list[i].Nonce < list[j].Nonce
 			})
 			tii.futureTxns[txn.ClientID] = list
+			logging.Logger.Debug("generate block (debug transaction) error - future txn",
```
dabasov: we can spam logs with this, better put it under debugTxn

code/go/0chain.net/miner/protocol_block.go
```diff
@@ -652,15 +654,17 @@ func txnProcessorHandlerFunc(mc *Chain, b *block.Block) txnProcessorHandler {
...
 			return false
 		case ErrNotTimeTolerant:
 			tii.invalidTxns = append(tii.invalidTxns, txn)
-			if debugTxn {
-				logging.Logger.Info("generate block (debug transaction) error - "+
-					"txn creation not within tolerance",
-					zap.String("txn", txn.Hash), zap.Int32("idx", tii.idx),
-					zap.Any("now", common.Now()))
-			}
+			//if debugTxn {
```
dabasov: uncomment the same is previous

dabasov (approved):
fix please logging

PR #1521, opened 2022-07-22
new free allocation mint logic

dabasov:
## Fixes
Minted tokens are defined with `free_allocation_settings.mint_amount` setting. This amount will be split by `free_allocation_settings.read_pool_fraction` between read and write pools and minted to the recipient of free_storage


code/go/0chain.net/smartcontract/storagesc/free_allocation.go
```diff
@@ -264,22 +264,24 @@ func (ssc *StorageSmartContract) freeAllocationRequest(
 			"marshal request: %v", err)
 	}
 
-	assigner.CurrentRedeemed += txn.Value
-	fTxnVal, err := txn.Value.Float64()
+	totalMint, _ := currency.ParseZCN(conf.FreeAllocationSettings.MintAmount)
```
peterlimg: We'd better check the error and return if any in case the conf.FreeAllocationSetting is invalid

code/go/0chain.net/smartcontract/storagesc/free_allocation.go
```diff
@@ -264,22 +264,24 @@ func (ssc *StorageSmartContract) freeAllocationRequest(
...
+	assigner.CurrentRedeemed += totalMint
+
 	if err != nil {
 		return "", common.NewErrorf("free_allocation_failed", "converting transaction value to float: %v", err)
 	}
-	readPoolTokens, err := currency.Float64ToCoin(fTxnVal * conf.FreeAllocationSettings.ReadPoolFraction)
+	f, _ := totalMint.Float64()
```
peterlimg: Check the error please

docker.local/config/sc.yaml
```diff
@@ -139,21 +139,19 @@ smart_contracts:
...
     free_allocation_settings:
-      # the free payment is split between read and write pool in the ratio
-      # read_pool_fraction : (1-read_pool_fraction)
-      read_pool_fraction: 0.2
-      data_shards: 10
-      parity_shards: 5
-      size: 10000000000
+      data_shards: 2
       duration: 50h
+      max_challenge_completion_time: 1m
```
peterlimg: This max_challenge_completion_time is not being used, we can remove it.

docker.local/config/sc.yaml
```diff
@@ -139,21 +139,19 @@ smart_contracts:
...
+      parity_shards: 2
```
peterlimg, line 142: Let's move the parity_shards above to make it closer to the data_shards.

peterlimg (changes requested):

docker.local/config/sc.yaml
ssardana08: @peterlimg max_challenge_completion_time is the cct set. it's the cct set for all blobbers..

dabasov: yeah thanks, got it

code/go/0chain.net/smartcontract/storagesc/free_allocation.go
```diff
@@ -264,13 +264,19 @@ func (ssc *StorageSmartContract) freeAllocationRequest(
...
 
-	totalMint, _ := currency.ParseZCN(conf.FreeAllocationSettings.MintAmount)
+	totalMint, err := currency.ParseZCN(conf.FreeAllocationSettings.MintAmount)
+	if err != nil {
+		return "", err
+	}
 	assigner.CurrentRedeemed += totalMint
 
 	if err != nil {
 		return "", common.NewErrorf("free_allocation_failed", "converting transaction value to float: %v", err)
```
peterlimg: redundant err check, we have checked above

peterlimg (changes requested):

code/go/0chain.net/smartcontract/storagesc/allocation.go
```diff
@@ -954,7 +953,7 @@ func (sc *StorageSmartContract) reduceAllocation(
 
 	// lock tokens if this transaction provides them
 	if txn.Value > 0 {
-		if err = alloc.addToWritePool(txn, false, balances); err != nil {
+		if err = alloc.addToWritePool(txn, txn.Value, balances); err != nil {
```
peterlimg: This will mint `txn.Value` new tokens, is this intended?, otherwise, we might need to refactor the `addToWritePool` to make it hard to be misused. So perhaps use optional function var, use it explicitly to mint when needed. For example,

```go
// when no need to min token
addToWriteTool(txn, balances)

// when need to mint token
addToWriteTool(txn, balances, OptionMintToken(token))
```

peterlimg (changes requested):

peterlimg (approved):
lgtm

PR #1547, opened 2022-07-31
hotfix failing tests

dabasov:
## Fixes
broken tests


lawRathod:
I have created a PR https://github.com/0chain/system_test/pull/365 to update system tests which is in need of review.

peterlimg:
The `expiration < 0` checking is commented out in staging, so we don't need this PR to do it again. Anyway, we will add it back with tests fixed.

PR #1562, opened 2022-08-03
remove invalid transactions on block finalization

dabasov:


code/go/0chain.net/chaincore/transaction/worker.go
```diff
@@ -91,8 +91,45 @@ func RemoveFromPool(ctx context.Context, txns []datastore.Entity) {
...
+			current, ok := qe.(*Transaction)
+			if !ok {
+				logging.Logger.Error("generate block (invalid entity)", zap.Any("entity", qe))
+				return true
+			}
+
+			list := mappedTxns[txn.GetKey()]
+			for _, fromList := range list {
+				if fromList.GetKey() == current.GetKey() {
+					if current.Nonce <= fromList.Nonce {
```
peterlimg: The txn.GetKey() returns the transaction hash, so the nonce would always be the same as they are the same transaction.

peterlimg (changes requested):

code/go/0chain.net/chaincore/transaction/worker.go
```diff
@@ -91,8 +91,45 @@ func RemoveFromPool(ctx context.Context, txns []datastore.Entity) {
...
+	var past []datastore.Entity
+	err := transactionEntityMetadata.GetStore().IterateCollection(cctx, transactionEntityMetadata, collectionName,
+		func(ctx context.Context, qe datastore.CollectionEntity) bool {
+			current, ok := qe.(*Transaction)
+			if !ok {
+				logging.Logger.Error("generate block (invalid entity)", zap.Any("entity", qe))
+				return true
+			}
+
+			list := mappedTxns[txn.GetKey()]
```
peterlimg: Seems the `txn` is an empty transaction, so I think the list is empty too.

code/go/0chain.net/chaincore/transaction/worker.go
```diff
@@ -91,8 +91,45 @@ func RemoveFromPool(ctx context.Context, txns []datastore.Entity) {
...
 
+	logging.Logger.Debug("cleaning invalid transactions")
+	mappedTxns := make(map[string][]*Transaction)
+	for _, e := range txns {
+		blockTx, ok := e.(*Transaction)
+		if !ok {
+			logging.Logger.Error("generate block (invalid entity)", zap.Any("entity", e))
+			continue
+		}
+		mappedTxns[blockTx.GetKey()] = append(mappedTxns[blockTx.GetKey()], blockTx)
```
peterlimg: What's the mappedTxns key, seems it's txn hash, but perhaps should be client id.

peterlimg (dismissed):

boddumanohar (approved):
LGTM

Issue #1575, opened 2022-08-06
Add challenge pools to event database


Sriep:
We should add the challenge pool to the event database. We can add to the [event.Allocation table](https://github.com/0chain/0chain/blob/staging/code/go/0chain.net/smartcontract/dbs/event/allocation.go#L16).

Change to [getChallengePoolStat](https://github.com/0chain/0chain/blob/staging/code/go/0chain.net/smartcontract/storagesc/handler.go#L670).

dabasov:
https://github.com/0chain/0chain/blob/a1c3312a7d5db50288780a0652c0474fc5694734/code/go/0chain.net/smartcontract/storagesc/handler.go#L773

dabasov:
eature/snapshots

PR #1628, opened 2022-08-31
ChallengePool table in eventsDB

iamrz1:
- Add challengePool table in eventsDB
- Move challengePool stat handler source to eventsDB

## Fixes
https://github.com/0chain/0chain/issues/1575

### DIFF
--- a/code/go/0chain.net/smartcontract/dbs/event/challenge_pool_test.go
+++ b/code/go/0chain.net/smartcontract/dbs/event/challenge_pool_test.go
@@ -1,0 +1,79 @@
+package event
+
+import (
+	"os"
+	"testing"
+	"time"
+
+	"0chain.net/chaincore/config"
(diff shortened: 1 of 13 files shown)


code/go/0chain.net/smartcontract/dbs/event/challenge_pool_test.go
```diff
@@ -0,0 +1 @@
+package event
```
dabasov, line 1: should we add tests?

dabasov (approved):

code/go/0chain.net/smartcontract/dbs/event/challenge_pool_test.go
iamrz1: Yes, we probably should add some unit tests here.

PR #1641, opened 2022-09-06
0xhrsh/zcnsc check sign threshold

### LINKED ISSUE #1589: zcnsc mint does not check if signatures number meet the threshold
The [zcnsc.Mint](https://github.com/0chain/0chain/blob/20ca71132a964d54c0773d2d76a882746fb7dc56/code/go/0chain.net/smartcontract/zcnsc/mint.go#L16) only checks if the signatures are correct, we should also check if the signatures number is >= threshold.

0xhrsh:
## Fixes
fixes #1589
## Tests

- blobber:
- gosdk:
- system_test:
- zboxcli:
- zwalletcli:
- Other: ...


code/go/0chain.net/smartcontract/storagesc/sc.go
```diff
@@ -125,16 +127,32 @@ func (ssc *StorageSmartContract) GetAddress() string {
 	return ADDRESS
 }
 
+func (ssc *StorageSmartContract) GetStat(name string) (int64, error) {
```
dabasov: you can't use metrics for it, the problem is that only MPT is protected by consensus, so we should store total number of authorizers in the different node with some hand crafted key for this value, similar to how we had ALL_BLOBBERS key before

dabasov (changes requested):

code/go/0chain.net/smartcontract/zcnsc/authorizer.go
```diff
@@ -116,7 +117,22 @@ func (zcn *ZCNSmartContract) AddAuthorizer(
...
-	return string(authorizer.Encode()), nil
+	err = updateAuthorizerCount(ctx)
+
+	return string(authorizer.Encode()), err
+}
+
+func updateAuthorizerCount(ctx cstate.StateContextI) (err error) {
+	var numAuth item
+	err = ctx.GetTrieNode(storagesc.ALL_AUTHORIZERS_KEY, &numAuth)
+	if err != nil {
```
dabasov, line 130: more precise is to check if ``err == NotPresent`` and return error in other case

code/go/0chain.net/smartcontract/zcnsc/context_helpers_test.go
```diff
@@ -15,11 +16,30 @@ import (
 	"0chain.net/core/datastore"
 	"0chain.net/core/encryption"
 	"0chain.net/smartcontract/dbs/event"
+	"0chain.net/smartcontract/storagesc"
 	. "0chain.net/smartcontract/zcnsc"
 	"github.com/0chain/common/core/util"
 	"github.com/stretchr/testify/mock"
 )
 
+type item struct {
```
dabasov: better use https://github.com/0chain/0chain/blob/79a0fd029fa398b87be187912d9345ce25635709/code/go/0chain.net/core/datastore/key.go#L62

code/go/0chain.net/smartcontract/zcnsc/mint.go
```diff
@@ -1,17 +1,37 @@
...
 	"0chain.net/chaincore/state"
 	"0chain.net/chaincore/transaction"
 	"0chain.net/core/common"
+	"0chain.net/smartcontract/storagesc"
 	"github.com/0chain/common/core/logging"
 	"github.com/pkg/errors"
 	"go.uber.org/zap"
 )
 
+type item struct {
```
dabasov: use ``NOIDField``

code/go/0chain.net/smartcontract/zcnsc/mint.go
```diff
@@ -86,11 +106,20 @@ func (zcn *ZCNSmartContract) Mint(trans *transaction.Transaction, inputData []by
...
 	if err != nil {
-		msg := fmt.Sprintf("failed to verify signatures with error: %v, %s", err, info)
-		err = common.NewError(code, msg)
+		err = common.NewError(code, fmt.Sprintf("failed to get number of authorizers: %s; %s", err.Error(), info))
+		return
+	}
+
+	// verify signatures of authorizers
+	count := payload.countValidSignatures(ctx)
+	if count < int(gn.PercentAuthorizers)*numAuth.Field {
```
dabasov: `` percent_authorizers: 0.7``
you should make ``math.RoundToEven(gn.PercentAuthorizers*numAuth.Field)``
int(gn.PercentAuthorizers) will cut the float.

dabasov (changes requested):

code/go/0chain.net/smartcontract/zcnsc/authorizer.go
peterlimg, line 130: Check if err is `util.ErrValueNotPresent` and return if it's not as we must return if the err is `ErrNodeNotFound`.

peterlimg (changes requested):

code/go/0chain.net/smartcontract/storagesc/models.go
```diff
@@ -35,6 +35,7 @@ const confMaxChallengeCompletionTime = "smart_contracts.storagesc.max_challenge_
 //go:generate msgp -io=false -tests=false -unexported -v
 
 var (
+	ALL_AUTHORIZERS_KEY              = ADDRESS + encryption.Hash("all_authorizers")
```
peterlimg: Use `AUTHORIZERS_COUNT_KEY` if it is just for counting the authorizers number.

code/go/0chain.net/smartcontract/zcnsc/mint.go
peterlimg: We don't have to use NOIDField as it's a MPT item.

peterlimg: This `item` is not a good name for saving authorizer count. It is usually be used as a temp local variable. So perhaps `authorizersCount`. The `Field` member may not necessary, could be `type authorizersCount int64`, but need to check if msgp could generate encode/decode methods for it.

code/go/0chain.net/smartcontract/zcnsc/mint.go
```diff
@@ -86,11 +106,20 @@ func (zcn *ZCNSmartContract) Mint(trans *transaction.Transaction, inputData []by
 		return
 	}
 
-	// verify signatures of authorizers
-	err = payload.verifySignatures(ctx)
+	var numAuth item
+	err = ctx.GetTrieNode(storagesc.ALL_AUTHORIZERS_KEY, &numAuth)
```
peterlimg: Please wrap the authorizersCount get/set in functions.

code/go/0chain.net/smartcontract/zcnsc/models.go
```diff
@@ -149,40 +149,38 @@ func (mp *MintPayload) GetStringToSign() string {
 	return encryption.Hash(fmt.Sprintf("%v:%v:%v:%v", mp.EthereumTxnID, mp.Amount, mp.Nonce, mp.ReceivingClientID))
 }
 
-func (mp *MintPayload) verifySignatures(state cstate.StateContextI) error {
+func (mp *MintPayload) countValidSignatures(state cstate.StateContextI) int {
```
peterlimg: Let's keep the original `verifySignatures()` method, we don't accept any invalid signatures. What we need to do are:
- validate all the signatures are valid
- there's no duplicate signatures
- all signatures number > threshold, which could be done out of this function

peterlimg (dismissed):

code/go/0chain.net/smartcontract/zcnsc/mint.go
dabasov: we do not need MPT item here

0xhrsh: i don't think we can directly use an int field here, i am getting the following error when i try to do so:
```
var numAuth int cannot use &numAuth (value of type *int) as util.MPTSerializable value in argument to ctx.InsertTrieNode: *int does not implement util.MPTSerializable
```

dabasov: yeah pure structs didn't work for me either

code/go/0chain.net/smartcontract/zcnsc/authorizer.go
dabasov: is it fixed?

dabasov (approved):
lgtm

code/go/0chain.net/smartcontract/zcnsc/mint.go
```diff
@@ -86,11 +87,20 @@ func (zcn *ZCNSmartContract) Mint(trans *transaction.Transaction, inputData []by
...
+	numAuth, err := getAuthorizerCount(ctx)
+
 	// verify signatures of authorizers
 	err = payload.verifySignatures(ctx)
 	if err != nil {
 		msg := fmt.Sprintf("failed to verify signatures with error: %v, %s", err, info)
 		err = common.NewError(code, msg)
+	}
+
+	if payload.countUniqueSignatures() < int(math.RoundToEven(gn.PercentAuthorizers*float64(numAuth))) {
```
peterlimg, line 99: Unique signatures checking should be done in `payload.verifySignatures` to avoid duplicate verification. The signature verify process is cpu immense, so if any of the signatures are duplicated, we should detect early.

code/go/0chain.net/smartcontract/zcnsc/models.go
```diff
@@ -185,6 +185,14 @@ func (mp *MintPayload) verifySignatures(state cstate.StateContextI) error {
 	return nil
 }
 
+func (mp *MintPayload) countUniqueSignatures() int {
+	uniqueSignatures := make(map[string]bool)
```
peterlimg, line 189: Please use `map[string]struct{}`.

code/go/0chain.net/smartcontract/zcnsc/models.go
```diff
@@ -299,3 +307,17 @@ func (ps *poolStat) encode() []byte {
...
+}
+
+func (ac *AuthCount) Encode() (data []byte, err error) {
+	data, err = json.Marshal(ac)
+	return
+}
+
+func (ac *AuthCount) Decode(input []byte) error {
+	err := json.Unmarshal(input, ac)
+	return err
```
peterlimg, line 322: Could be one line code of `return json.Unmarhsal(input, ac)`

code/go/0chain.net/smartcontract/zcnsc/mint.go
0xhrsh: hey @peterlimg, do we need to delete duplicate signatures from `payload.Signatures`?

code/go/0chain.net/smartcontract/zcnsc/models.go
0xhrsh: did all these changes in #1682

Issue #1550, opened 2022-08-01
[EventsDB] Finalized block flag


dabasov:
Add nullable timestamp field to record when this block was finalized

boddumanohar:
fixed by: https://github.com/0chain/0chain/pull/1656

PR #1656, opened 2022-09-11
Add a field to identify if the block is finalised or not

boddumanohar:
## Fixes
https://github.com/0chain/0chain/issues/1550

## Changes
throughout the lifetime of a block, we emit block events to events db at 2 places:
* when the block is notarized
* when the block is finalised

to differentiate between both the type of block events we add a boolean field that identify the finalised blocks

## Tests

- blobber:
- gosdk:
- system_test:
- zboxcli:
- zwalletcli:
- Other: ...

### DIFF
--- a/code/go/0chain.net/miner/protocol_block.go
+++ b/code/go/0chain.net/miner/protocol_block.go
@@ -624,6 +624,7 @@ func (mc *Chain) FinalizeBlock(ctx context.Context, b *block.Block) error {
 	for idx, txn := range b.Txns {
 		modifiedTxns[idx] = txn
 	}
+	b.SetBlockFinalised()
 	return mc.deleteTxns(modifiedTxns)
 }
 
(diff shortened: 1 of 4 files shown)


code/go/0chain.net/miner/protocol_block.go
```diff
@@ -624,6 +624,7 @@ func (mc *Chain) FinalizeBlock(ctx context.Context, b *block.Block) error {
 	for idx, txn := range b.Txns {
 		modifiedTxns[idx] = txn
 	}
+	b.FinalisedAt = common.Now()
```
peterlimg: Different sharders will have different value of `common.Now()`, which will result in different `block.FinalizedAt` value. Basically every sharder has its own finalized time. Why do we need to save it to db, as if not necessary, I would suggest to not save inconsistent data.

peterlimg (dismissed):

code/go/0chain.net/miner/protocol_block.go
boddumanohar: throughout the lifetime of a block, we emit block events to events db at 2 places:
1. when the block is notarized
2. when the block is finalised
to differentiate between both the type of block events we would like to have a field that identify the finalised blocks. We could also use a boolean field. What do you suggest? shall I make it a boolean field?

dabasov (approved):
lgtm

PR #1663, opened 2022-09-13
Switch cancelation log messages type to debug

fabioDMFerreira:
#1187 

* Also fixes building of conductor images caused by the new setup handlers


dabasov (changes requested):
lgtm

dabasov (approved):
lgtm

Issue #838, opened 2021-12-29
Limit number of signatures in sc.yaml


Dmdv:
```
type MintPayload struct {
	EthereumTxnID     string                 `json:"ethereum_txn_id"`
	Amount            state.Balance          `json:"amount"`
	Nonce             int64                  `json:"nonce"`
	Signatures        []*AuthorizerSignature `json:"signatures"`
	ReceivingClientID string                 `json:"receiving_client_id"`
}

```

ma2b0043:
@Dmdv when can we expect to close this issue?

boddumanohar:
~currently blocked by : https://github.com/0chain/0chain/issues/1602~

~so fixing that issue first.~

peterlimg:
closed via #1667

PR #1667, opened 2022-09-14
limit the number of signatures

boddumanohar:
## Fixes
https://github.com/0chain/0chain/issues/838

## Changes
limit the number of signatures

## Tests

- blobber:
- gosdk:
- system_test:
- zboxcli:
- zwalletcli:
- Other: ...

### DIFF
--- a/code/go/0chain.net/smartcontract/zcnsc/mint.go
+++ b/code/go/0chain.net/smartcontract/zcnsc/mint.go
@@ -48,6 +48,21 @@ func (zcn *ZCNSmartContract) Mint(trans *transaction.Transaction, inputData []by
 		return
 	}
 
+	numAuth, err := getAuthorizerCount(ctx)
+	threshold := int(math.RoundToEven(gn.PercentAuthorizers * float64(numAuth)))
+
+	// if number of slices exceeds limits the check only withing required range
+	if len(payload.Signatures) < threshold {
+		msg := fmt.Sprintf("no of signatures lesser than threshold %d: %v, %s", threshold, err, info)
+		err = common.NewError(code, msg)
+		return
+	}
+
+	if len(payload.Signatures) > numAuth {
+		logging.Logger.Info("no of signatures execeed the no of available authorizers", zap.Int("available", numAuth))
+		payload.Signatures = payload.Signatures[0:numAuth]
+	}
+
 	// ClientID - is a client who broadcasts this transaction to mint token
 	// ToClientID - is an address of the smart contract
 	if payload.ReceivingClientID != trans.ClientID {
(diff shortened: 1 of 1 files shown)


code/go/0chain.net/smartcontract/zcnsc/mint.go
```diff
@@ -41,6 +41,11 @@ func (zcn *ZCNSmartContract) Mint(trans *transaction.Transaction, inputData []by
 		return
 	}
 
+	// if number of slices exceeds limits the check only withing required range
+	if len(payload.Signatures) > gn.MaxSignatures {
```
peterlimg: Threshold should also be considered, imagine `gn.MaxSignatures < threshold`. Thiss could happen when we have lot of authorizers registered, suppose 200, with the threshold of 0.7, the required signatures would be 140, which is greater than the default max signatures of 100. So perhaps the max signatures config is not necessary, we can use threshold as the max number of signatures we're going to check.

peterlimg (changes requested):

code/go/0chain.net/smartcontract/zcnsc/mint.go
dabasov: we can use ``if len(payload.Signatures) > max(gn.MaxSignatures, threshold)`` but frankly I'd rejected all mint transactions with more than `gn.MaxSignatures` and didn't slice it even

dabasov (approved):

code/go/0chain.net/smartcontract/zcnsc/mint.go
peterlimg: The `gn.MaxSignatures` does not scale, as we can register new authorizers without limitation, though we will control how many authorizers would be allowed, but then it is the `MaxAuthorizers * threshold` that we should check instead of the `gn.MaxSignatures`.

peterlimg (changes requested):
Seems the MaxAuthorizers is not being used, we can remove it as the sc owner will take care of the number of authorizers. Otherwise, if we hit the max number, we will have to issue  transactions to update the config to extend it

boddumanohar:
@peterlimg  I've updated the PR to have the upper of the signatures as the no of authorizers

peterlimg (approved):
lgtm

boddumanohar:
@dabasov if these changes looks good, can we merge this PR? thanks

service-0chain:
| config  | value |
| tests skipped  | TRUE  |
| system-tests  | master  |
| gosdk  | NOT OVERRIDDEN  |
| zboxcli  | staging  |
| zwalletcli  | staging  |
| 0chain  | signatures-limit  |
| blobber  | staging  |
| authorizer  | staging  |
| 0box  | staging  |
| 0block  | staging  |
| 0dns  | staging  |
| explorer  | master  |
| 0proxy  | staging  |
| 0search  | staging  |

PR #1681, opened 2022-09-17
Feature/removed the usage of block size param

### LINKED ISSUE #1505: Remove `block.max_block_size` from 0chain.yaml
Block generation should only be only dependent on block cost: `max_block_cost`
and hence remove `max_block_size`.

and smartcontracts have cost to be used.

YarikRevich:
## Fixes

Removes usage of "block_size" param

Referes to issue #1505


code/go/0chain.net/sharder/chain.go
```diff
@@ -35,10 +34,7 @@ func SetupSharderChain(c *chain.Chain) {
 	sharderChain.RoundChannel = make(chan *round.Round, 1)
 	blockCacheSize := 100
 	sharderChain.BlockCache = cache.NewLRUCache(blockCacheSize)
-	transactionCacheSize := int(c.BlockSize()) * blockCacheSize
-	if transactionCacheSize > 5000 {
-		transactionCacheSize = 5000
-	}
+	transactionCacheSize := 5 * blockCacheSize
```
boddumanohar, line 37: irrespective of the blocksize, we are setting the transaction cache size as 5 times the block cache size (which is 500). Why 5 times?

YarikRevich: I discussed this with @dabasov. He said, that in general transactionCacheSize is not mandatory to be calculated in the way it did before. So there may be any constant

code/go/0chain.net/miner/miner/worker.go
```diff
@@ -54,35 +54,12 @@ func TransactionGenerator(c *chain.Chain, workdir string) {
...
-	}
-
 	// validate the maxFee and minFee, maxFee must > minFee, otherwise, will panic
 	if maxFee-minFee <= 0 {
 		logging.Logger.Panic(fmt.Sprintf("development.txn_generation.max_txn_fee must be greater than "+
 			"development.txn_generation.min_txn_fee, max_fee: %v, min_fee: %v", maxFee, minFee))
 	}
 
-	switch {
-	case blockSize <= 10:
```
dabasov, line 70: looks like we should calculate `blockSize` as `totalCost/txnCost`
`totalCost` is in config  `    max_block_cost: 10000 #equal to 100ms`
for txnCost it is a little bit more trickier, but looks like we only generate transfer transactions and now we have cost for it in configs `    transfer_cost: 10`

code/go/0chain.net/miner/protocol_block.go
```diff
@@ -787,12 +787,12 @@ func (tii *TxnIterInfo) checkForCurrent(txn *transaction.Transaction) {
 	}
 }
 
-func newTxnIterInfo(blockSize int32) *TxnIterInfo {
+func newTxnIterInfo() *TxnIterInfo {
 	return &TxnIterInfo{
 		clients:    make(map[string]*client.Client),
-		eTxns:      make([]datastore.Entity, 0, blockSize),
```
dabasov: we can estimate blockSize maximum value the same way

dabasov (changes requested):

code/go/0chain.net/miner/miner/worker.go
```diff
@@ -105,7 +95,7 @@ func TransactionGenerator(c *chain.Chain, workdir string) {
 	ts := rand.NewSource(time.Now().UnixNano())
 	trng := rand.New(ts)
 	for {
-		numTxns = trng.Int31n(blockSize)
+		numTxns = trng.Int31n(100)
```
peterlimg: Use the blockSize calculated above

code/go/0chain.net/miner/protocol_block.go
```diff
@@ -592,7 +592,9 @@ func (mc *Chain) updateFinalizedBlock(ctx context.Context, b *block.Block) {
 		txns = append(txns, txn)
 	}
 
-	tii := newTxnIterInfo(mc.BlockSize())
+	blockSize := mc.ChainConfig.MaxBlockCost() / mc.ChainConfig.TxnTransferCost()
```
peterlimg: blockSize here is the number of txns, so it is `len(b.Txns)`.

code/go/0chain.net/miner/protocol_block.go
```diff
@@ -873,10 +873,12 @@ func (mc *Chain) generateBlock(ctx context.Context, b *block.Block,
 		return ErrLFBClientStateNil
 	}
 
-	b.Txns = make([]*transaction.Transaction, 0, mc.BlockSize())
+	b.Txns = make([]*transaction.Transaction, 0, 100)
+
+	initialBlockSize := mc.ChainConfig.MaxBlockCost() / mc.ChainConfig.TxnTransferCost()
```
peterlimg: The cost of `transfer` transaction is not the average cost of all kinds of transactions. So let's not use it to calculate the block size. The initialBlockSize could be the same as `b.Txns` , it will be extended when the capacity is reached.

peterlimg (changes requested):

dabasov (approved):
lgtm

peterlimg (approved):
lgtm, please fix the unit tests

service-0chain:
| config  | value |
| tests skipped  | FALSE  |
| system-tests  | master  |
| gosdk  | NOT OVERRIDDEN  |
| zboxcli  | staging  |
| zwalletcli  | staging  |
| 0chain  | feature/remove_max_block_size  |
| blobber  | staging  |
| authorizer  | staging  |
| 0box  | staging  |
| 0block  | staging  |
| 0dns  | staging  |
| explorer  | master  |
| 0proxy  | staging  |
| 0search  | staging  |

PR #1611, opened 2022-08-22
adds config to chain during genesis

0xhrsh:
## Fixes
fixes #687 
## Tests
Tasks to complete before merging PR:
- [ ]  Ensure system tests are passing. If not [Run them manually](https://github.com/0chain/0chain/actions/workflows/system_tests.yml) to check for any regressions :clipboard:
- [ ]  Do any new system tests need added to test this change? do any existing system tests need updated? If so create a PR at [0chain/system_test](https://github.com/0chain/system_test)
- [ ]  Merge your system tests PR to master AFTER merging this PR

## Associated PRs (Link as appropriate):
- blobber:
- gosdk:
- system_test:
- zboxcli:
- zwalletcli:
- Other: ...

PR #1692, opened 2022-09-20
updates config functions

### LINKED ISSUE #1392: Load config from MPT on node starts and update on block finalization
To avoid loading config from MPT every time when need to access the config info. Such as the code is called each time when selecting blobbers for blobber challenge generation and get config REST handler.
https://github.com/0chain/0chain/blob/f858985cdc34c5ba14378afbe43fc26998c3149e/code/go/0chain.net/smartcontract/storagesc/handler.go#L669-L683

0xhrsh:
## Fixes
fixes #1392
## Tests

- blobber:
- gosdk:
- system_test:
- zboxcli:
- zwalletcli:
- Other: ...


code/go/0chain.net/chaincore/chain/state/state_context.go
```diff
@@ -60,6 +60,8 @@ type CommonStateContextI interface {
 	GetTrieNode(key datastore.Key, v util.MPTSerializable) error
 	GetBlock() *block.Block
 	GetLatestFinalizedBlock() *block.Block
+	GetConfig() *storagescConfig
```
peterlimg: The config could be different in different SCs, such as minersc, storagesc. So the `storagescConfig` may only work for storagesc. The issue of #1392 is not an easy task actually, we will need to identify what config are common global config from 0chain.yaml, and what need to store in separate SC packages. We will also need to maintain in memory config variables for them as caches so that we don't need load them from MPT each time when we need to read from them.

peterlimg (changes requested):

code/go/0chain.net/chaincore/chain/state/state_context.go
dabasov: add mapped configs by sc to the context and it might work

dabasov:
where are other configs?

0xhrsh:
> where are other configs?

fixing the tests before adding other configs

code/go/0chain.net/smartcontract/storagesc/handler.go
```diff
@@ -830,20 +830,28 @@ func (srh *StorageRestHandler) getReadPoolStat(w http.ResponseWriter, r *http.Re
...
+	if *balances.GetConfig() != nil {
+		updateConfig(balances)
+	}
+	var conf = (*balances.GetConfig()).(Config)
+	return &conf, nil
+}
+
+func updateConfig(balances cstate.CommonStateContextI) error {
 	var conf = &Config{}
 	err := balances.GetTrieNode(scConfigKey(ADDRESS), conf)
```
boddumanohar, line 870: it looks like the only place where we are updating the balances.config is in [getConfig](https://github.com/0chain/0chain/blob/4875edf53fa590cde20028fcb1a0d0ddd59ec696/code/go/0chain.net/smartcontract/storagesc/handler.go#L831-L832)

But the ticket also tells to the update the config on block finalization also. should we also update there?

peterlimg: @boddumanohar the SC will update the config if any, the 'update config on block finalization' is mean to update the config cache.

0xhrsh:
> > where are other configs?
> 
> fixing the tests before adding other configs

I think it would be better If i create another PR for the rest of the configs after closing #1611 
can we do that @dabasov @peterlimg?

0xhrsh:
> I think it would be better If i create another PR for the rest of the configs after closing #1611 

quite a bit of the code required here is already written there

code/go/0chain.net/chaincore/chain/state/state_context.go
```diff
@@ -73,6 +74,8 @@ type CommonStateContextI interface {
 	InsertTrieNode(key datastore.Key, v util.MPTSerializable) (datastore.Key, error)
 	GetBlock() *block.Block
 	GetLatestFinalizedBlock() *block.Block
+	GetConfig(smartcontract string) (*SCConfig, error)
```
peterlimg, line 77: SCConfig is an interface, so let's not use pointer.

code/go/0chain.net/chaincore/chain/state/state_context.go
```diff
@@ -137,6 +142,10 @@ type StateContext struct {
 	getSignature                  func() encryption.SignatureScheme
 	eventDb                       *event.EventDb
 	mutex                         *sync.Mutex
+	faucetscConfig                *SCConfig
+	storagescConfig               *SCConfig
+	vestingscConfig               *SCConfig
+	zcnscConfig                   *SCConfig
```
peterlimg, line 148: We should not couple the SCs configs and the state package. This would not scale, which means we need to change the StateContext each time a new config need to be added. 

We actually don't need to expose through StateContext, instead keep the configs in SCs packages themself. I have thought about the necessary to update configs on finalization, and I think we don't have to, as it may cause problem if we do that. So what we will need to do is update the config in time when the MPT config is updated in SC packages.

peterlimg (changes requested):

peterlimg:
Dead PR

Issue #1636, opened 2022-09-04
[Logging] Add trace level


dabasov:
We need trace level, to add logging for
- sql queries

```golang
db, err = gorm.Open(postgres.Open(fmt.Sprintf(
   "host=%v port=%v user=%v dbname=%v password=%v sslmode=disable",
   config.Host,
   config.Port,
   config.User,
   config.Name,
   config.Password)),
   &gorm.Config{
      Logger:                 logger.Default.LogMode(logger.Silent),
      SkipDefaultTransaction: true,
      PrepareStmt:            true,
   })
```

-  enable transaction logging for Trace
```golang
/*DebugTxn - is this a transaction that needs being debugged
- applicable only when running in test mode and the transaction_data string contains debug keyword somewhere in it
*/
func (t *Transaction) DebugTxn() bool {
	if !config.Development() {
		return false
	}
	return strings.Contains(t.TransactionData, "debug")
}

```

Issue #467, opened 2021-08-10
Add config for turn on/off debug/pprof 


cnlangzi:
debug pprof is a debugging tool. 
- It will take more resources to collection information for debugging. 
- it is a security issue leak sensitive information.

so it is better to turn off on production based on config file

ma2b0043:
@cnlangzi when can we expect to close this issue?

cnlangzi:
it should be reachable from localhost only on production.

PR #1701, opened 2022-09-25
Add switches for verbosity and profiling

iamrz1:
## Fixes
https://github.com/0chain/0chain/issues/1636
https://github.com/0chain/0chain/issues/467

### DIFF
--- a/code/go/0chain.net/miner/miner/main.go
+++ b/code/go/0chain.net/miner/miner/main.go
@@ -7,7 +7,7 @@ import (
 	"flag"
 	"fmt"
 	"net/http"
-	_ "net/http/pprof"
+	"net/http/pprof"
 	"os"
 	"path/filepath"
 	"runtime"
@@ -204,7 +204,28 @@ func main() {
 	initIntegrationsTests(node.Self.Underlying().GetKey())
 
 	var server *http.Server
+	var profServer *http.Server
+
+	mux := http.NewServeMux()
+	http.DefaultServeMux = mux
+
 	if config.Development() {
+		if viper.GetBool("development.pprof") {
+			// start pprof server
+			pprofMux := http.NewServeMux()
+			profServer = &http.Server{
+				Addr:           fmt.Sprintf(":%d", node.Self.Underlying().Port+10),
+				ReadTimeout:    30 * time.Second,
+				MaxHeaderBytes: 1 << 20,
+				Handler:        pprofMux,
+			}
+			initProfHandlers(pprofMux)
+			go func() {
+				err2 := profServer.ListenAndServe()
+				logging.Logger.Info("Http server shut down", zap.Error(err2))
+			}()
+		}
+
 		// No WriteTimeout setup to enable pprof
 		server = &http.Server{
 			Addr:           address,
@@ -328,6 +349,10 @@ func main() {
 	}
 
 	shutdown := common.HandleShutdown(server, []func(){shutdownIntegrationTests, done, chain.CloseStateDB})
+	if profServer != nil {
+		shutdownProf := common.HandleShutdown(profServer, nil)
+		<-shutdownProf
+	}
 	<-shutdown
 	time.Sleep(2 * time.Second)
 	logging.Logger.Info("0chain miner shut down gracefully")
@@ -463,4 +488,12 @@ func initWorkers(ctx context.Context) {
 	serverChain.SetupWorkers(ctx)
 	//miner.SetupWorkers(ctx)
 	transaction.SetupWorkers(ctx)
+}
+
+func initProfHandlers(mux *http.ServeMux) {
+	mux.HandleFunc("/debug/pprof/", pprof.Index)
+	mux.HandleFunc("/debug/pprof/cmdline", pprof.Cmdline)
+	mux.HandleFunc("/debug/pprof/profile", pprof.Profile)
+	mux.HandleFunc("/debug/pprof/symbol", pprof.Symbol)
+	mux.HandleFunc("/debug/pprof/trace", pprof.Trace)
 }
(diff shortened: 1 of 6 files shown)


code/go/0chain.net/miner/miner/main.go
```diff
@@ -204,7 +204,21 @@ func main() {
...
 	var server *http.Server
+
+	mux := http.NewServeMux()
+	http.DefaultServeMux = mux
+
 	if config.Development() {
+		if viper.GetBool("development.pprof") {
+			// start pprof server
+			go func() {
+				pprofMux := http.NewServeMux()
```
dabasov: we need a way to shut down it gracefully

dabasov (changes requested):

dabasov:
@iamrz1 I've found this, we should log txn also
https://github.com/0chain/0chain/blob/e26382b0541421adc4cc2e323022ac660ff53f9e/code/go/0chain.net/chaincore/state/state_debug.go#L32

code/go/0chain.net/miner/miner/main.go
iamrz1: We'll need to add more invasive changes to the current code then. because we have to shutdown both servers following the same signal.

iamrz1: Added code to shutdown prof server. Works with test code, please take a look yourself.

code/go/0chain.net/miner/miner/main.go
```diff
@@ -204,7 +204,28 @@ func main() {
 	initIntegrationsTests(node.Self.Underlying().GetKey())
 
 	var server *http.Server
+	var profServer *http.Server
+
+	mux := http.NewServeMux()
+	http.DefaultServeMux = mux
```
peterlimg, line 210: No need to set a new mux for DefaultServeMux, the pprof will use another mux anyway.

code/go/0chain.net/miner/miner/main.go
```diff
@@ -204,7 +204,28 @@ func main() {
...
+
 	if config.Development() {
+		if viper.GetBool("development.pprof") {
+			// start pprof server
+			pprofMux := http.NewServeMux()
+			profServer = &http.Server{
+				Addr:           fmt.Sprintf(":%d", node.Self.Underlying().Port+10),
```
peterlimg: Do not use this `+10` port, which will have ports confliction probelm when we have more than 10 miners. Define it in config file and use a specific port number, e.g 6060, etc.

peterlimg (changes requested):

code/go/0chain.net/miner/miner/main.go
iamrz1: Valid point, I will change the range. But this is for development only. And if you look at out docker-compose file for local deployment, it exposes ports on 717(*), which means exactly 9 ports. Thus a clash seems unlikely.

code/go/0chain.net/miner/miner/main.go
iamrz1: Prof is pre-initialized for default server mux. If we do not replace this, prof handlers will be available for main router, which is not expected behavior. I have checked it myself in a test setting.

iamrz1:
> @iamrz1 I've found this, we should log txn also
> 
> https://github.com/0chain/0chain/blob/e26382b0541421adc4cc2e323022ac660ff53f9e/code/go/0chain.net/chaincore/state/state_debug.go#L32

Added. Please check.

dabasov (approved):
lgtm

iamrz1:
I can't merge PRs on 0chain, please merge if there's no further reviews.

code/go/0chain.net/miner/miner/main.go
peterlimg: Ah, that makes sense. Then we can make the switch code simpler, i.e replace the default mux with new one for disabling the pprof, and use the default router when enable the pprof. In this way, we don't need to setup pprof handlers manually, also we don't need to change the port number. So the changes we need for switching is alike:

```go
if !enabelPProf() {
	http.DefaultServeMux = http.NewServeMux()
}
```

peterlimg:
@iamrz1 please check new comments

code/go/0chain.net/miner/miner/main.go
iamrz1: I have already had this discussion with @dabasov. We dont want pprof on the main router ever, so that someone doesn't enable it in production by mistake. So pprof will be switched on an off on a different router running on a different port.

peterlimg: Enable it on main router is not a problem, we may enable it to do [Continuous Profiling](https://medium.com/google-cloud/continuous-profiling-of-go-programs-96d4416af77b) in the future, and it should be safe to enable pprof in prod. The main point is to make the change/code simple, and avoid unnecessary changes on both code logic and deployment scripts. This change will fail the pprof on networks if we don't update the helm repo.

dabasov: all debug functionality should be removed to dev, it makes no sense for end user and regular miner will never switch on profiling, no matter how much resources is spent, simply no use-case here.Devs can always switch on it for devv networks, good separation of concerns, I don't see any reason to keep it in the same router

peterlimg: It doesn't make any difference actually. We can disable the pprof in production, and enable it in dev only. This is a simple feature, which only require 3 lines of code changes. But running it in another router would require changes in both the code, deployment scripts and 0helm config, and I actually don't see much benefit of doing it. Anyway, if you insist to do this, please remember to update 0helm to add a new ngix ingress router for pprof.

dabasov: @peterlimg I'm afraid that it will create the overhead of profiling functions, it might be not big, but nevertheless meaningful

peterlimg: Sure, let's merge this one and update the 0helm repo.

peterlimg (approved):
lgtm

PR #1709, opened 2022-09-29
add authorizer now works correctly

dabasov:
## Fixes
add-authorizers

### DIFF
--- a/code/go/0chain.net/smartcontract/zcnsc/authorizer.go
+++ b/code/go/0chain.net/smartcontract/zcnsc/authorizer.go
@@ -60,7 +60,10 @@ func (zcn *ZCNSmartContract) AddAuthorizer(
 		return "", err
 	}
 
-	publicKeyBytes, _ := hex.DecodeString(params.PublicKey)
+	publicKeyBytes, err := hex.DecodeString(params.PublicKey)
+	if err != nil {
+		return "", err
+	}
 	authorizerID = encryption.Hash(publicKeyBytes)
 	clientId = tran.ClientID
 	if params.StakePoolSettings.DelegateWallet == "" {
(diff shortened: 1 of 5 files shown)


uneet7 (approved):
lgtm

code/go/0chain.net/smartcontract/zcnsc/authorizer.go
```diff
@@ -53,24 +54,15 @@ func (zcn *ZCNSmartContract) AddAuthorizer(
...
-
-	authorizerID = params.ID
-
 	if params.PublicKey == "" {
 		err = common.NewError(code, "public key was not included with transaction")
 		Logger.Error("public key error", zap.Error(err))
 		return "", err
 	}
 
 	publicKeyBytes, _ := hex.DecodeString(params.PublicKey)
```
peterlimg: As commented in another PR, let's not ignore the decode error.

peterlimg (dismissed):

code/go/0chain.net/smartcontract/zcnsc/authorizer.go
dabasov: @peterlimg usually, we don't improve code we haven't touched in the review, it can slowdown the process severely

dabasov: @uneet7 I've fixed it, but please handle all errors you get especially in smart contracts

peterlimg: @dabasov it's a minor fix that does not worth a new PR, otherwise we might just ignore it and forget it. The process would only be slowdown when we let code go to production easily and then spend more time to debug/fix it later .

PR #1725, opened 2022-10-04
Created _diagnostics/txns_in_pool to show transactions currenlty in tx…

manasnisar:
…n_pool

## Changes

Added _diagnostics/txns_in_pool route to show transactions currenlty in txn_pool

## Tests

- blobber:
- gosdk:
- system_test:
- zboxcli:
- zwalletcli:
- Other: ...

### DIFF
--- a/code/go/0chain.net/chaincore/chain/handler.go
+++ b/code/go/0chain.net/chaincore/chain/handler.go
@@ -1622,17 +1622,11 @@ func (c *Chain) MinerStatsHandler(w http.ResponseWriter, r *http.Request) {
 	}
 }
 
-func txnIterHandlerFunc(c *Chain, w http.ResponseWriter) func(context.Context, datastore.CollectionEntity) bool {
+func txnIterHandlerFunc(c *Chain, w http.ResponseWriter, s *state.State) func(context.Context, datastore.CollectionEntity) bool {
 	return func(ctx context.Context, ce datastore.CollectionEntity) bool {
 		txn, ok := ce.(*transaction.Transaction)
 		if !ok {
 			logging.Logger.Error("generate block (invalid entity)", zap.Any("entity", ce))
-			return false
-		}
-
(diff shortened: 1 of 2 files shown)


code/go/0chain.net/chaincore/chain/handler.go
```diff
@@ -1585,6 +1622,62 @@ func (c *Chain) MinerStatsHandler(w http.ResponseWriter, r *http.Request) {
...
 
+func txnIterHandlerFunc(c *Chain, w http.ResponseWriter) func(context.Context, datastore.CollectionEntity) bool {
+	return func(ctx context.Context, ce datastore.CollectionEntity) bool {
+		txn, ok := ce.(*transaction.Transaction)
+		if !ok {
+			logging.Logger.Error("generate block (invalid entity)", zap.Any("entity", ce))
+			return false
+		}
+
+		s, err := c.GetStateById(c.GetLatestFinalizedBlock().ClientState, txn.ClientID)
```
peterlimg: We should get the latest finalized block once and use it instead of getting it on each iteration calling. The block will be changed overtime.

peterlimg (dismissed):
Generally, we should get all txns from pool first, then we can either render the data to html or json. In this case, we can only do html, but please avoid render the data while iterate the txns from pool.

code/go/0chain.net/chaincore/chain/handler.go
manasnisar: Sure. I'll make the changes

dabasov:
@peterlimg what's the problem with rendering HTML right in handler?

dabasov (approved):
lgtm

peterlimg:
@dabasov we should not render html while accessing dbs, no matter whether it is sql db or a redis. It's a basic MVC model design to read data from db first, then render it in different formats as needed unless there are significant performance gaining. Another concern is we'd better finish the iteration as soon as possible, so we'd better not doing the rendering while reading.

Issue #1723, opened 2022-10-03
Hash mismatch on demo


dabasov:
It can be reproduced on any network, sharders stuck on hash mismatch

```
2022-10-02T09:19:22.179Z	ERROR	sharder/protocol_round.go:83	AddNotarizedBlock failed to compute state	{"round": 2649143, "error": "state_mismatch: Computed state hash doesn't match with the state hash of the block"}
0chain.net/sharder.(*Chain).AddNotarizedBlock
	/0chain/code/go/0chain.net/sharder/protocol_round.go:83
0chain.net/sharder.(*Chain).processBlock
	/0chain/code/go/0chain.net/sharder/protocol_block.go:283
0chain.net/sharder.(*Chain).BlockWorker
	/0chain/code/go/0chain.net/sharder/worker.go:157
2022-10-02T09:19:22.179Z	ERROR	block/entity.go:988	compute state - state hash mismatch	{"minerID": "f17e8b0d117132111b30d95cc9a1362fc3f929e8afc692ba88c8d309f038e177", "round": 2649143, "block": "cee316a800eff51b5d1e58cc28fabbcfc07401fdbfa114eff40578b88ee503df", "block_size": 3, "changes": 81, "begin_client_state": "3f5f1e8848d712fd8134c7c2cb496b58b7e4ce3d0c2387063d0543d778b980d2", "computed_state_hash": "adf80865d9c3dd52a128492dc46d3d495d58cdba98c4423cb3dfe1f973f2fbdd", "block_state_hash": "a63211615763a9047eba1b8be1ced0d080c524c5e100ef77e807dc047e80eb7d", "prev_block": "634c38eff4e377b4fd98c40b2d70619e3c54ec6a926d79a6c3e225d57f82d986", "prev_block_client_state": "3f5f1e8848d712fd8134c7c2cb496b58b7e4ce3d0c2387063d0543d778b980d2"}
0chain.net/chaincore/block.(*Block).ComputeState
	/0chain/code/go/0chain.net/chaincore/block/entity.go:988
0chain.net/chaincore/chain.(*Chain).computeState
	/0chain/code/go/0chain.net/chaincore/chain/state.go:80
0chain.net/chaincore/chain.(*Chain).ComputeState.func1
	/0chain/code/go/0chain.net/chaincore/chain/state.go:49
0chain.net/chaincore/chain.(*Chain).ComputeBlockStateWithLock
	/0chain/code/go/0chain.net/chaincore/chain/worker.go:797
[...]

Issue #1688, opened 2022-09-19
Parallel loops, error messages need to be deterministic.


Sriep:
Examples of parallel for loops:
[fetchStorageNodePlusStake](https://github.com/0chain/0chain/blob/staging/code/go/0chain.net/smartcontract/storagesc/allocation.go#L423) [fetchPools](https://github.com/0chain/0chain/blob/staging/code/go/0chain.net/smartcontract/storagesc/allocation.go#L481) [getStakePoolsByIDs](https://github.com/0chain/0chain/blob/staging/code/go/0chain.net/smartcontract/storagesc/allocation.go#L727) [getAllocationBlobbers](https://github.com/0chain/0chain/blob/staging/code/go/0chain.net/smartcontract/storagesc/allocation.go#L767) [blobberBlockRewards](https://github.com/0chain/0chain/blob/staging/code/go/0chain.net/smartcontract/storagesc/block_reward.go#L81). There may be others.

If multiple errors are found in these loops, then the error is random. For example, we might have both `value not present` and other times `node not found`. We should sort the errors found and return an order list.

Probably be best if we could do this in common code to reduce the chance of this confusion being reintroduced in the future.

peterlimg:
Closed via #1729

PR #1729, opened 2022-10-05
Fix state hash mismatch

peterlimg:
## Fixes
- #1723 
- #1688

## Changes

- Fix concurrently MPT access caused state hash mismatch error
- Implemented common way to get items from MPT concurrently
- Store invalid state error 'node not found' in StateContext, it will be saved whenever it happens when read/write/delete node from MPT. Check the invalid state error first to ensure invalid state error would not be ignored, which would lead to state hash mismatch error.

No

## Tests

### DIFF
--- a/code/go/0chain.net/chaincore/chain/state/state_context.go
+++ b/code/go/0chain.net/chaincore/chain/state/state_context.go
@@ -3,6 +3,7 @@ package state
 import (
 	"fmt"
 	"sort"
+	"strings"
 	"sync"
 
 	"github.com/0chain/common/core/logging"
@@ -104,6 +105,7 @@ type StateContextI interface {
 	EmitEvent(event.EventType, event.EventTag, string, interface{}, ...Appender)
 	EmitError(error)
 	GetEvents() []event.Event // cannot use in smart contracts or REST endpoints
+	GetInvalidStateErrors() []error
 }
 
 // StateContext - a context object used to manipulate global state
@@ -122,6 +124,7 @@ type StateContext struct {
 	getSignature                  func() encryption.SignatureScheme
 	eventDb                       *event.EventDb
 	mutex                         *sync.Mutex
+	invalidStateErrors            []error
 }
 
 type GetNow func() common.Timestamp
@@ -334,6 +337,7 @@ func (sc *StateContext) getClientState(clientID string) (*state.State, error) {
 	err := sc.state.GetNodeValue(util.Path(clientID), s)
 	if err != nil {
 		if err != util.ErrValueNotPresent {
+			sc.addInvalidStateError(err)
 			return nil, err
 		}
 		return s, err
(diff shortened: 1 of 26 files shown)


code/go/0chain.net/chaincore/chain/state/state_context.go
```diff
@@ -400,3 +402,87 @@ func (sc *StateContext) SetStateContext(s *state.State) error {
...
+		// error data added to transaction output.
+		logging.Logger.Error("could not get items", zap.Any("errors", errIdxs))
+		retErr := errIdxs[0]
+		return nil, fmt.Errorf("could not get item %q: %v", ids[retErr.index], retErr.err)
+	}
+
+	//ensure original ordering
+	items := make([]T, len(ids))
+	for item := range itemC {
+		items[item.index] = item.item.(T)
```
dabasov: if ID assigned to different type is passed then this type cast can panic, we better check the cast with comma ok to prevent panic

code/go/0chain.net/smartcontract/storagesc/blobber.go
```diff
@@ -866,8 +866,14 @@ func (sc *StorageSmartContract) insertBlobber(t *transaction.Transaction,
 	balances cstate.StateContextI,
 ) (err error) {
 	savedBlobber, err := sc.getBlobber(blobber.ID, balances)
-	if err == nil {
+	switch err {
```
dabasov: previous version was more compact and in more golang style 
```go
if err == nil {
  return sc.updateBlobber(t, conf, blobber, savedBlobber, balances)
}

    return err
}
```

dabasov (changes requested):

code/go/0chain.net/chaincore/chain/state/state_context.go
```diff
@@ -400,3 +402,87 @@ func (sc *StateContext) SetStateContext(s *state.State) error {
...
+	// check internal error first
+	select {
+	case err := <-stateErrC:
+		return nil, err
+	default:
+	}
+
+	errIdxs := make([]errorIndex, 0, len(ids))
+	for ei := range errC {
+		errIdxs = append(errIdxs, ei)
```
iamrz1, line 528: If `we would only return one 'value not present' error (the first one)`, then shouldn't we just grab the error form the object with lowest index value (if len(errC) > 0 )? Do we need to create an array of object and then sort the object array?

code/go/0chain.net/chaincore/chain/state/state_context.go
peterlimg: The type cast would fail if the `GetItemFunc` callback function did not implement correctly, adding comma ok would make this kind of error goto transaction as a chargeable error, which is not what we wanted. So we will need to ensure the `getItem` returns the `T` type, I'll do a fix,  then compiler would help us to detect the type error, which means if we don't implement the `GetItemFunc` with correct type, it wouldn't be able to pass the build.

dabasov: the problem is that if there is a bug there and `GetItemFunc()` returns unexpected type, then malicious user can create transaction that will `panic()` on all miners and network will be dead (during block generation). if no panic happens, then this transaction will simply be marked as invalid and nothing strange will happen, even no restart will occur. panics are reasonable only during startups I guess or something really unrecoverable, but any tricky transactions should be processed as usual with chargeable or unchargeable errors

peterlimg: The `GetItemFunc()` can not return unexpected type, wrong type can't pass the compiler. The only possible unexpected value is `nil`, but we will fix it. Anyway, I'll do comma ok check to ensure that it would not panic.

code/go/0chain.net/chaincore/chain/state/state_context.go
peterlimg: @iamrz1 yes, we will only return the first err, but we would logs all the errors for debugging. And perhaps return all if needed in the future. So should be fine to do this unless the performance is affected a lot.

dabasov (approved):
lgtm

service-0chain:
| config  | value |
| tests skipped  | FALSE  |
| system-tests  | master  |
| gosdk  | NOT OVERRIDDEN  |
| zboxcli  | staging  |
| zwalletcli  | staging  |
| 0chain  | fix/state-hash-mismatch  |
| blobber  | staging  |
| authorizer  | staging  |
| 0box  | staging  |
| 0block  | staging  |
| 0dns  | staging  |
| explorer  | master  |
| 0proxy  | staging  |
| 0search  | staging  |

service-0chain:
| config  | value |
| tests skipped  | TRUE  |
| system-tests  | master  |
| gosdk  | NOT OVERRIDDEN  |
| zboxcli  | staging  |
| zwalletcli  | staging  |
| 0chain  | fix/state-hash-mismatch  |
| blobber  | staging  |
| authorizer  | staging  |
| 0box  | staging  |
| 0block  | staging  |
| 0dns  | staging  |
| explorer  | master  |
| 0proxy  | staging  |
| 0search  | staging  |

PR #1743, opened 2022-10-12
Fix sharder token locking issue

peterlimg:
## Fixes

- Lock token to provider with correct provider type, previously it was always miner.
- Update `getNodesList` to only use keys from `AllMinersKeys` and `AllShardersKeys`, then load items by ids concurrently from MPT to avoid sync issue. Because of the nodes were not synced, even we locked token to sharder, and added the pool id to the sharder's delegate pool, it would be overwritten to nil later when update rewards to sharders delegates in payFee txn. This is kind of a workaround fix, and we will need to refactor to only store node keys in `AllMinersKeys` and `AllShardersKeys`, or replace the nodes list with partitions.  

## Tests

No


dabasov (approved):
lgtm

PR #1759, opened 2022-10-22
Fix wrong block generation conductor tests

fabioDMFerreira:
* add conductor directive that allows setting magic blocks for different tests. This directive overrides the magic block configuration file specified in `0chain.yaml`.

* Generate new DKG files for the magic block `miners_3_sharder_1`.

* Change env variables names used to find keys and dkg files.

* Separate conductor client initialization from the node registration.


code/go/0chain.net/conductor/conductrpc/client.go
```diff
@@ -212,3 +212,11 @@ func (c *client) addBlockClientStats(req []byte) (err error) {
 	}
 	return
 }
+
+func (c *client) magicBlock() (configFile *string, err error) {
+	err = c.client.Call("Server.MagicBlock", &struct{}{}, &configFile)
+	for err == rpc.ErrShutdown {
```
dabasov, line 218: what's the reason for this loop?

code/go/0chain.net/conductor/conductrpc/entity.go
```diff
@@ -29,6 +29,18 @@ func (e *Entity) State() (state *State) {
...
 }
 
+// MagicBlock returns the location path of the magic block configuration.
+func (e *Entity) MagicBlock() (magicBlock string) {
+	magicBlockPtr, err := e.client.magicBlock()
+	if err != nil {
+		log.Fatalf("failed getting magic block: %v", err)
+	}
+
+	magicBlock = *magicBlockPtr
```
dabasov: if return unnamed parameter you don't need this line

code/go/0chain.net/conductor/conductrpc/server.go
```diff
@@ -392,6 +394,12 @@ func (s *Server) ShareOrSignsShares(soss *ShareOrSignsSharesEvent,
 	return
 }
 
+// magic block handler
+func (s *Server) MagicBlock(_ *struct{}, configFile *string) (err error) {
+	(*configFile) = s.magicBlock
```
dabasov, line 399: it doesn't make any sense, we can't pass entities by reference

dabasov (changes requested):

code/go/0chain.net/conductor/conductrpc/entity.go
fabioDMFerreira: What do you mean?

code/go/0chain.net/conductor/conductrpc/server.go
fabioDMFerreira: I am following the same convention as the other similar methods. What do you suggest doing?

code/go/0chain.net/conductor/conductrpc/client.go
fabioDMFerreira: This came from the state method that uses a polling strategy to fetch state data. Prob. the client connection drops after some time if it is not used. I couldn't find anything related to this on the web though.

dabasov: hm, weird, usually we want to resend requests every small period of time, not in the loop, but it maybe ok for now

code/go/0chain.net/conductor/conductrpc/entity.go
dabasov:
```go
func (e *Entity) MagicBlock() string {
	magicBlockPtr, err := e.client.magicBlock()
	if err != nil {
		log.Fatalf("failed getting magic block: %v", err)
	}

	return *magicBlockPtr
}
```

code/go/0chain.net/conductor/conductrpc/server.go
dabasov: ok, It works somehow and we change the value of *configFile here, it looks weird to be honest, made test here
https://go.dev/play/p/e_N_4lHLk8u

code/go/0chain.net/conductor/conductrpc/entity.go
fabioDMFerreira: I got to the named parameters that forgot the other way. Thanks

fabioDMFerreira: fixed by [4922e7c](https://github.com/0chain/0chain/pull/1759/commits/4922e7cb226587a2e44f518f693f73ffc5a0d1b2)

dabasov (approved):
lgtm

peterlimg (approved):
lgtm

PR #1776, opened 2022-10-27
Need to confirm it is needed. Order event processing by round

Sriep:
## Fixes
Prevents events from being processed out of order. It assumes that we always receive events in two packets.

It is not clear whether we need this. While events are processed out of order, they do it regularly:
```
n
n-4
n+1
n-3
n+2
n-2
n+3
n-1
n+4
n
```
So is it possible that we don't have to worry about conflicts ?

## Tests

- blobber:
- gosdk:
- system_test:
- zboxcli:
- zwalletcli:
- Other: ...

### DIFF
--- a/code/go/0chain.net/chaincore/chain/protocol_block.go
+++ b/code/go/0chain.net/chaincore/chain/protocol_block.go
@@ -377,7 +377,7 @@ func (c *Chain) finalizeBlock(ctx context.Context, fb *block.Block, bsh BlockSta
 	c.SetLatestOwnFinalizedBlockRound(fb.Round)
 	c.SetLatestFinalizedBlock(fb)
 
-	if len(fb.Events) > 0 && c.GetEventDb() != nil {
+	if c.GetEventDb() != nil {
 		wg.Run("finalize block - add events", fb.Round, func() {
 			if err := c.GetEventDb().ProcessEvents(ctx, fb.Events, fb.Round, fb.Hash, len(fb.Txns)); err != nil {
 				logging.Logger.Error("finalize block - add events failed",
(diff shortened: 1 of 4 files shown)


service-0chain:
| config  | value |
| tests skipped  | FALSE  |
| system-tests  | master  |
| gosdk  | NOT OVERRIDDEN  |
| zboxcli  | staging  |
| zwalletcli  | staging  |
| 0chain  | order_events  |
| blobber  | staging  |
| authorizer  | staging  |
| 0box  | staging  |
| 0block  | staging  |
| 0dns  | staging  |
| explorer  | master  |
| 0proxy  | staging  |
| 0search  | staging  |

peterlimg:
Events in the same round could be re ordered due to events merging, but they should work in order between rounds as events are processed round by round on block finalization. Say we are finalizing round 100, the events in round 101 would not be processed until the round 100 is finalized. So basically I'm not sure how do you see the events go out of order.

Sriep:
The add-block event and the other events are processed in two different packets. One of these lags the other by four rounds. As long as this is constant, then we should not have objects getting corrupted. However, the latest row in the block table won't match the other objects in the database, we should monitor to ensure that this will not cause a problem.

peterlimg:
@Sriep I think the code [here](https://github.com/0chain/0chain/blob/8c2a9e9d5e4b22593c53b21cd9f0bc5ad6832662/code/go/0chain.net/sharder/sharder/sharder.go#L160) will insert events to db out of order, which is used for adding blocks to event db when we receive them. But that should be fine as it is just for recording the blocks, we won't use any data or state from those blocks.

PR #1778, opened 2022-10-30
Add Validation to ProviderType Parameter

### LINKED ISSUE #1772: validate ProviderType parameter
This code do not validate incoming parameters, which lead to a problem
```go
	if sp, err = ssc.getStakePool(spr.ProviderType, spr.ProviderID, balances); err != nil {
		return "", common.NewErrorf("stake_pool_lock_failed",
			"can't get stake pool: %v", err)
	}
```
ProviderType is an enum with method String() which panics on invalid inputs
```go

type Provider int

const (
	Miner Provider = iota + 1
	Sharder
	Blobber
	Validator
	Authorizer
)

var providerString = []string{"unknown", "miner", "sharder", "blobber", "validator", "authorizer"}

func (p Provider) String() string {
	return providerString[p]
}
```
we have to 
- validate inputs for ProviderType 
- we have to return error from ``String()`` if it's not in the range to prevent panic

uneet7:
## Fixes
#1772 

## Tests

- blobber:
- gosdk:
- system_test:
- zboxcli:
- zwalletcli:
- Other: ...

### DIFF
--- a/code/go/0chain.net/smartcontract/minersc/handler.go
+++ b/code/go/0chain.net/smartcontract/minersc/handler.go
@@ -796,6 +796,10 @@ func (mrh *MinerRestHandler) getUserPools(w http.ResponseWriter, r *http.Request
 	ups := new(userPools)
 	ups.Pools = make(map[string][]*delegatePoolStat, len(minerPools)+len(sharderPools))
 	for _, pool := range minerPools {
+		if pool.Status >= spenum.PoolStatus(pool.Status).Size() {
+			common.Respond(w, r, nil, errors.New("invalid miner pool status"))
+			return
+		}
 		dp := delegatePoolStat{
 			ID:     pool.PoolID,
 			Status: spenum.PoolStatus(pool.Status).String(),
@@ -810,6 +814,10 @@ func (mrh *MinerRestHandler) getUserPools(w http.ResponseWriter, r *http.Request
 	}
 
 	for _, pool := range sharderPools {
+		if pool.Status >= spenum.PoolStatus(pool.Status).Size() {
+			common.Respond(w, r, nil, errors.New("invalid sharder pool status"))
+			return
+		}
 		dp := delegatePoolStat{
 			ID:     pool.PoolID,
 			Status: spenum.PoolStatus(pool.Status).String(),
--- a/code/go/0chain.net/smartcontract/minersc/handler.go
+++ b/code/go/0chain.net/smartcontract/minersc/handler.go
@@ -796,14 +796,9 @@ func (mrh *MinerRestHandler) getUserPools(w http.ResponseWriter, r *http.Request
 	ups := new(userPools)
 	ups.Pools = make(map[string][]*delegatePoolStat, len(minerPools)+len(sharderPools))
 	for _, pool := range minerPools {
-		poolStatusStr, err := spenum.PoolStatus(pool.Status).String()
-		if err != nil {
-			common.Respond(w, r, nil, err)
-			return
-		}
 		dp := delegatePoolStat{
 			ID:     pool.PoolID,
-			Status: poolStatusStr,
+			Status: spenum.PoolStatus(pool.Status).String(),
 		}
 		dp.Balance = pool.Balance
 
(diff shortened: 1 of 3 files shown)


code/go/0chain.net/smartcontract/stakepool/spenum/enums.go
```diff
@@ -32,5 +32,12 @@ const (
...
 
 func (p PoolStatus) String() string {
-	return poolString[p]
+	if int(p) < len(poolString) {
+		return poolString[p]
+	}
+	return ""
+}
+
+func (p PoolStatus) Size() int {
```
peterlimg: We don't need the Size() method here, the previous `String()` is enough.

code/go/0chain.net/smartcontract/stakepool/spenum/enums.go
```diff
@@ -32,5 +32,12 @@ const (
 var poolString = []string{"active", "pending", "inactive", "unstaking", "deleting", "deleted"}
 
 func (p PoolStatus) String() string {
-	return poolString[p]
+	if int(p) < len(poolString) {
+		return poolString[p]
+	}
+	return ""
```
peterlimg: Return value from poolString only when p is in [0, len(poolString)), so please also check if p is < 0. 

Return `unknown pool status` if p is invalid.

code/go/0chain.net/smartcontract/minersc/handler.go
```diff
@@ -796,6 +796,10 @@ func (mrh *MinerRestHandler) getUserPools(w http.ResponseWriter, r *http.Request
 	ups := new(userPools)
 	ups.Pools = make(map[string][]*delegatePoolStat, len(minerPools)+len(sharderPools))
 	for _, pool := range minerPools {
+		if pool.Status >= spenum.PoolStatus(pool.Status).Size() {
```
peterlimg: We don't need to do this size checking as the `String()` method will do it for us and return unknown status if it's invalid. We will return it along with other valid info to users

code/go/0chain.net/smartcontract/minersc/handler.go
```diff
@@ -810,6 +814,10 @@ func (mrh *MinerRestHandler) getUserPools(w http.ResponseWriter, r *http.Request
 	}
 
 	for _, pool := range sharderPools {
+		if pool.Status >= spenum.PoolStatus(pool.Status).Size() {
```
peterlimg: Same as above

code/go/0chain.net/smartcontract/storagesc/handler.go
```diff
@@ -1137,6 +1137,10 @@ func (srh *StorageRestHandler) getUserStakePoolStat(w http.ResponseWriter, r *ht
 	var ups = new(userPoolStat)
 	ups.Pools = make(map[datastore.Key][]*delegatePoolStat)
 	for _, pool := range pools {
+		if pool.Status >= spenum.PoolStatus(pool.Status).Size() {
```
peterlimg: Same as in minersc

peterlimg (changes requested):

peterlimg:
@uneet7 we don't need to return error for the `unknown pool status`. The `String()` method should only return a string so that it can be used as a `stringer` interface. So for all kinds of invalid status, we can just return "unknown pool status".

code/go/0chain.net/smartcontract/storagesc/handler.go
```diff
@@ -1137,10 +1137,15 @@ func (srh *StorageRestHandler) getUserStakePoolStat(w http.ResponseWriter, r *ht
 	var ups = new(userPoolStat)
 	ups.Pools = make(map[datastore.Key][]*delegatePoolStat)
 	for _, pool := range pools {
+		poolStatusStr, err := spenum.PoolStatus(pool.Status).String()
+		if err != nil {
+			common.Respond(w, r, nil, err)
```
peterlimg: We don't have to response error when the pool has invalid status and ignore all other info, instead we should return all to users, and let users to determine what they should do base on the status.

peterlimg (changes requested):
Please fix all the tests

peterlimg (approved):
lgtm

PR #1802, opened 2022-11-13
1769 minerssharders nodes could be added without limitation

### LINKED ISSUE #1769: Miners/Sharders nodes could be added without limitation
We don't have any limit for adding miners or sharders to the the nodes list, malicious users could attack the network by adding lots of faked nodes to the list.

devyetii:
## Changes
- Add NodeListLength type as a new type of nodes that can be saved in the MPT
- Add 2 new node keys with their accessors/mutators (MinerListLength/SharderListLength)
- Check if the new added miner/sharder exist in the active set (i.e. mentioned in the MagicBlock) -> Will be changed when view change feature is enabled.
- IDK
## Tests

- blobber:
- gosdk:
- system_test:
- zboxcli:
- zwalletcli:
- Other: ...


peterlimg:
@devyetii this would not be able to prevent the attack actually, suppose we set the max miners number as 100, attackers could occupy all or most of the positions by adding fake nodes to the list, then honest miners would just failed to join the network due to the max limit reached.

peterlimg:
One possible way to prevent this kind of attack before enabling VC is checking the magic block to see if the nodes are in the magic block nodes list. This fix would not work if we enable the VC later, but could be a shot term fix. 

After enabling the VC, perhaps we can add a minimal staking requirement when adding a new miner, so if it is the first time the miner come to the network, the add_miner txn the miner send must provide minimal staking tokens. In this way, we can prevent spam nodes in certain extent, and the VC process will then ensure that those nodes do not obey the rules get removed from the magic block. @guruhubb what do you think.

dabasov:
we discussed with @devyetii similar approach with staking just when node is added. I suggest to implement it together with view-change, as it is tightly coupled with this process, for now we can hard cap miners and check if it is the same set as in magic block, while we are not changing magic block it should work

service-0chain:
| config  | value |
| tests skipped  | FALSE  |
| system-tests  | increase_settingUpdateSleepTime  |
| gosdk  | NOT OVERRIDDEN  |
| zboxcli  | staging  |
| zwalletcli  | staging  |
| 0chain  | 1769-minerssharders-nodes-could-be-added-without-limitation  |
| blobber  | staging  |
| authorizer  | staging  |
| 0box  | staging  |
| 0block  | staging  |
| 0dns  | staging  |
| explorer  | master  |
| 0proxy  | staging  |
| 0search  | staging  |

devyetii:
@peterlimg After pushing a small change - just changing the waiting time between two lines ([link](https://github.dev/0chain/system_test/blob/360139c28b67d7d7397b5afef199a743d97c9e6c/tests/cli_tests/zwalletcli_storage_update_config_test.go#L90-L91), I ran the system tests branch I pushed to against this branch and it passed successfully : https://github.com/0chain/0chain/actions/runs/3524409881/jobs/5909774247

code/go/0chain.net/smartcontract/minersc/miner.go
```diff
@@ -171,6 +180,9 @@ func (msc *MinerSmartContract) DeleteMiner(
 		return "", common.NewError("delete_miner", err.Error())
 	}
 
+	lockAllMiners.Lock()
```
dabasov: why do we need this lock?

dabasov (changes requested):

code/go/0chain.net/smartcontract/minersc/models.go
```diff
@@ -911,6 +913,92 @@ func (dmn *DKGMinerNodes) GetHashBytes() []byte {
...
+func getMinerListLength(state cstate.QueryStateContextI) (*NodeListLength, error) {
+	length, err := getOrInitializeLengthNode(state, MinerListLengthKey);
+	if err != nil {
+		return nil, err
+	}
+
+	return length, nil
+}
+
+func IncrementMinerListLength(state cstate.QueryStateContextI) (*NodeListLength, error) {
```
dabasov: where is this func used?

code/go/0chain.net/smartcontract/minersc/models.go
```diff
@@ -911,6 +913,92 @@ func (dmn *DKGMinerNodes) GetHashBytes() []byte {
...
+
+	err = updateLengthNode(state, MinerListLengthKey, length)
+	if err != nil {
+		return nil, err
+	}
+
+	return length, nil
+}
+
+func DecrementMinerListLength(state cstate.QueryStateContextI) (*NodeListLength, error) {
```
dabasov: where is it used?

dabasov (changes requested):

service-0chain:
| config  | value |
| tests skipped  | FALSE  |
| system-tests  | master  |
| gosdk  | NOT OVERRIDDEN  |
| zboxcli  | staging  |
| zwalletcli  | staging  |
| 0chain  | 1769-minerssharders-nodes-could-be-added-without-limitation  |
| blobber  | staging  |
| authorizer  | staging  |
| 0box  | staging  |
| 0block  | staging  |
| 0dns  | staging  |
| explorer  | master  |
| 0proxy  | staging  |
| 0search  | staging  |

dabasov (approved):
lgtm

code/go/0chain.net/smartcontract/minersc/miner.go
```diff
@@ -47,6 +47,15 @@ func (msc *MinerSmartContract) AddMiner(t *transaction.Transaction,
 	lockAllMiners.Lock()
 	defer lockAllMiners.Unlock()
 
+	magicBlockMiners := balances.GetChainCurrentMagicBlock().Miners
+	
+	if magicBlockMiners.HasNode(newMiner.ID) == false {
```
peterlimg, line 52: Please do `if !magicBlockMiners.HasNode(newMiner.ID)`

PR #1823, opened 2022-11-21
Fix/staking

dabasov:
## Tests

- gosdk: https://github.com/0chain/gosdk/pull/662
- system_test: https://github.com/0chain/system_test/pull/487
- zwalletcli: https://github.com/0chain/zwalletcli/pull/199
- zboxcli: https://github.com/0chain/zboxcli/pull/333
- Other: ...


code/go/0chain.net/smartcontract/dbs/event/stakepool.go
```diff
@@ -15,6 +15,17 @@ import (
 	"0chain.net/smartcontract/dbs"
 )
 
+type Provider struct {
```
Sriep: Why not call this file provider.go to fit wither other tables? 

However, I don't like the name as this looks more like ProviderStakePoolSettings. We might want to reserve the Provider table name for actual providers.

code/go/0chain.net/smartcontract/dbs/event/blobber.go
```diff
@@ -19,6 +19,8 @@ const ActiveBlobbersTimeLimit = 5 * time.Minute // 5 Minutes
 
 type Blobber struct {
 	gorm.Model
+	*Provider
```
Sriep: I think the provider object here should hold stuff like

```go
type Provider struct {
  privider_id string
  provider_type enums.ProvidedrType
  Status bool // killed shutdown active etc..
  StakePoolSettings StakePoolSettings
  LastHealthCheck Timestamp
}
```

Basically, everything logically belonging to every provider.

code/go/0chain.net/smartcontract/stakepool/lock.go
```diff
@@ -107,3 +109,33 @@ func (sp *StakePool) LockPool(
 
 	return nil
 }
+
+func (sp *StakePool) StakeForProvider(providerType spenum.Provider, providerID string, balances cstate.StateContextI) error {
```
Sriep: This method only exists to emit events, so maybe rename it to something like `EmitStakeForProvider`.

code/go/0chain.net/smartcontract/stakepool/lock.go
```diff
@@ -107,3 +109,33 @@ func (sp *StakePool) LockPool(
...
+	staked, err := sp.stake()
+	if err != nil {
+		return err
+	}
+
+	logging.Logger.Info("emitting stake event")
+	switch providerType {
+	case spenum.Blobber:
+		tag, data := event.NewUpdateBlobberTotalStakeEvent(providerID, staked)
```
Sriep, line 123: Why couple `event` and `stakepool` packages in this way? 

Can't we just have a unified event, `TagUpdateTotalStakeEvent`? The event package can have a version of this switch statement to split the different provider types. By moving this logic into the `event` package we might do away with the switch altogether.

Sriep (commented):
It might be an idea to merge these changes with https://github.com/0chain/0chain/pull/1439 and https://github.com/0chain/0chain/pull/1454.

code/go/0chain.net/smartcontract/stakepool/lock.go
peterlimg: Yes, I agree, ideally the stakepool package should not know what provider types are using it.

code/go/0chain.net/smartcontract/stakepool/lock.go
```diff
@@ -119,3 +119,33 @@ func (sp *StakePool) LockPool(
 
 	return nil
 }
+
+func (sp *StakePool) StakeForProvider(providerType spenum.Provider, providerID string, balances cstate.StateContextI) error {
+	staked, err := sp.stake()
```
peterlimg, line 115: We should use the difference stakes between current total stake and the previous one, otherwise the provider's total stakes will be doubled each time when this function is called.  

Suppose a blobber now has 100 zcn total staked, a client staked 10 zcc to it, the total stake of the blobber will be 110, if the same client stake another 10, then the pool will have totally 20 zcn, which is the total staked value here, and after processing the total stake update event, the blobber's total stake will be 110+20 =  130, which is incorrect.

code/go/0chain.net/smartcontract/dbs/event/stakepool.go
peterlimg: Yes, agree, ProviderStakePoolSettings or StakePoolSettings would fit better, let's reserve the name of Provider

peterlimg (changes requested):

code/go/0chain.net/smartcontract/stakepool/lock.go
dabasov: how can we find out which provider was staked? for example, we have different aggregates for different provider types, we need to add provider type also in the event

code/go/0chain.net/smartcontract/stakepool/lock.go
dabasov: @peterlimg it was simply moved from previous implementation,I wonder, how it was all working before. @Sriep do you know if it is all tested?

code/go/0chain.net/smartcontract/dbs/event/blobber.go
dabasov: I can add other fields, but @peterlimg it is not settings, since `totalStake` is also here

code/go/0chain.net/smartcontract/dbs/event/stakepool.go
dabasov: I can add other fields, but @peterlimg it is not settings, since totalStake is also here

code/go/0chain.net/smartcontract/stakepool/lock.go
dabasov: about visibility I can create different common package to import to every sc, like `provider` but does it really make any difference with `stakepool`?

code/go/0chain.net/smartcontract/stakepool/lock.go
dabasov: I checked TotalOffers, doesn't look to be working correctly either
```
tag, data := event.NewUpdateBlobberTotalOffersEvent(providerID, sp.TotalOffers)
```

```
func withBlobberTotalOffersAdded() eventMergeMiddleware {
	return withEventMerge(func(a, b *Blobber) (*Blobber, error) {
		a.OffersTotal += b.OffersTotal
		return a, nil
	})
}
```
looks like we also summarize total values here if several events are emitted, cmiiw @peterlimg

code/go/0chain.net/smartcontract/dbs/event/blobber.go
dabasov: moved Id to provider and  removed model

code/go/0chain.net/smartcontract/stakepool/lock.go
dabasov: renamed

code/go/0chain.net/smartcontract/stakepool/lock.go
dabasov: changed for TotalStaked to be overwritten during merging

peterlimg: @dabasov it worked previously because we use txn hash as pool id, so basically we would not stake to existing pool, the `sp.Stake()` is always a fresh one, and safe to be added to blobber's total stake.

code/go/0chain.net/smartcontract/dbs/event/authorizer.go
```diff
@@ -86,15 +78,42 @@ func (a *Authorizer) exists(edb *EventDb) (bool, error) {
...
+	}).Create(&authorizer).Error
+}
+
+func mergeUpdateAuthorizerTotalStakesEvents() *eventsMergerImpl[Authorizer] {
+	return newEventsMerger[Authorizer](TagUpdateAuthorizerTotalStake, withAuthorizerTotalStakesAdded())
+}
+
+func withAuthorizerTotalStakesAdded() eventMergeMiddleware {
+	return withEventMerge(func(a, b *Authorizer) (*Authorizer, error) {
+		//use the newer one
```
peterlimg: Please use `withUniqueEventOverwrite` for the mergeUpdateAuthorizerTotalStakesEvents if we don't need to do addition. So  as other providers.

peterlimg (changes requested):

code/go/0chain.net/smartcontract/stakepool/lock.go
dabasov: aaah, got  it, it's tricky

service-0chain:
| config  | value |
| tests skipped  | FALSE  |
| system-tests  | master  |
| gosdk  | NOT OVERRIDDEN  |
| zboxcli  | staging  |
| zwalletcli  | staging  |
| 0chain  | fix/staking  |
| blobber  | staging  |
| authorizer  | staging  |
| 0box  | staging  |
| 0block  | staging  |
| 0dns  | staging  |
| explorer  | master  |
| 0proxy  | staging  |
| 0search  | staging  |

service-0chain:
| config  | value |
| tests skipped  | FALSE  |
| system-tests  | master  |
| gosdk  | NOT OVERRIDDEN  |
| zboxcli  | staging  |
| zwalletcli  | staging  |
| 0chain  | fix/staking  |
| blobber  | staging  |
| authorizer  | staging  |
| 0box  | staging  |
| 0block  | staging  |
| 0dns  | staging  |
| explorer  | master  |
| 0proxy  | staging  |
| 0search  | staging  |

Sriep:
You know in [Lock](https://github.com/0chain/0chain/blob/2149f19d476855ba6e54bf8f4c013d9aa51aac54/code/go/0chain.net/smartcontract/stakepool/lock.go#L59) wether you are adding a new stake pool or updating an existing one.
```go
	var newPoolId = txn.ClientID
	dp, ok := sp.Pools[newPoolId]
	if !ok {
		// new stake
		dp = &DelegatePool{
			Balance:      txn.Value,
			Reward:       0,
			Status:       status,
			DelegateID:   txn.ClientID,
			RoundCreated: balances.GetBlock().Round,
			StakedAt:     txn.CreationDate,
		}

		sp.Pools[newPoolId] = dp

	} else {
````
I think we should record whether we are adding or updating at this point, propagate the `ok` variable, then send either a `TagAddDelegatePoolEvent` or `TagUpdateDelegatePoolEvent` event depending on whether the client already has a stake pool.

Include the provider type in the object sent with the event. If a `switch` statement is needed to untangle which provider type the new stake pool belongs to, this can be done in the `dbs/event` package.

**Comment by service-0chain** [comment]
| config  | value |
| tests skipped  | FALSE  |
| system-tests  | master  |
| gosdk  | NOT OVERRIDDEN  |
| zboxcli  | fix/staking  |
| zwalletcli  | staging  |
| 0chain  | fix/staking  |
| blobber  | staging  |
| authorizer  | staging  |
| 0box  | staging  |
| 0block  | staging  |
| 0dns  | staging  |
| explorer  | master  |
| 0proxy  | staging  |
| 0search  | staging  |

**Comment by service-0chain** [comment]
| config  | value |
| tests skipped  | FALSE  |
| system-tests  | master  |
| gosdk  | NOT OVERRIDDEN  |
| zboxcli  | fix/staking  |
| zwalletcli  | staging  |
| 0chain  | fix/staking  |
| blobber  | staging  |
| authorizer  | staging  |
| 0box  | staging  |
| 0block  | staging  |
| 0dns  | staging  |
| explorer  | master  |
| 0proxy  | staging  |
| 0search  | staging  |

**Comment by service-0chain** [comment]
| config  | value |
| tests skipped  | FALSE  |
| system-tests  | fix/staking  |
| gosdk  | NOT OVERRIDDEN  |
| zboxcli  | staging  |
| zwalletcli  | fix/staking  |
| 0chain  | fix/staking  |
| blobber  | staging  |
| authorizer  | staging  |
| 0box  | staging  |
| 0block  | staging  |
| 0dns  | staging  |
| explorer  | master  |
| 0proxy  | staging  |
| 0search  | staging  |

**Comment by peterlimg** [comment]
> You know in [Lock](https://github.com/0chain/0chain/blob/2149f19d476855ba6e54bf8f4c013d9aa51aac54/code/go/0chain.net/smartcontract/stakepool/lock.go#L59) wether you are adding a new stake pool or updating an existing one.
> 
> ```go
> 	var newPoolId = txn.ClientID
> 	dp, ok := sp.Pools[newPoolId]
> 	if !ok {
> 		// new stake
> 		dp = &DelegatePool{
> 			Balance:      txn.Value,
> 			Reward:       0,
> 			Status:       status,
> 			DelegateID:   txn.ClientID,
> 			RoundCreated: balances.GetBlock().Round,
> 			StakedAt:     txn.CreationDate,
> 		}
> 
> 		sp.Pools[newPoolId] = dp
> 
> 	} else {
> ```
> 
> I think we should record whether we are adding or updating at this point, propagate the `ok` variable, then send either a `TagAddDelegatePoolEvent` or `TagUpdateDelegatePoolEvent` event depending on whether the client already has a stake pool.
> 
> Include the provider type in the object sent with the event. If a `switch` statement is needed to untangle which provider type the new stake pool belongs to, this can be done in the `dbs/event` package.

The upsert will create for new stakes and update if the item already exist, so we may not two tags for this.

File: code/go/0chain.net/smartcontract/stakepool/stakepool.go [hunk 7]
```diff
@@ -27,6 +28,15 @@ func stakePoolKey(p spenum.Provider, id string) datastore.Key {
 	return p.String() + ":stakepool:" + id
 }
 
+type AbstractStakePool interface {
```
peterlimg, line 36: It's not necessary to have this interface, we can pass the *StakePool to keep it simple (e.g, avoid save()) -> Save() changing). Interface is useful when doing mock tests and hide info, but I don't see it was used for these purpose and basically all places are using the raw StakePool except the StakePoolLock function below. The interface is also useful when it was used in a package that does not know the details of the data struct, but in this case, the StakePool is simply defined in the same package, so let's just use it.

code/go/0chain.net/smartcontract/stakepool/stakepool.go
```diff
@@ -482,3 +496,65 @@ func equallyDistributeRewards(coins currency.Coin, pools []*DelegatePool, spUpda
...
+type stakePoolRequest struct {
+	ProviderType spenum.Provider `json:"provider_type,omitempty"`
+	ProviderID   string          `json:"provider_id,omitempty"`
+}
+
+func (spr *stakePoolRequest) decode(p []byte) (err error) {
+	if err = json.Unmarshal(p, spr); err != nil {
+		return
+	}
+	return // ok
```
peterlimg: could be `return json.Unmarshal(p, spr)`

code/go/0chain.net/smartcontract/stakepool/stakepool.go
```diff
@@ -482,3 +496,65 @@ func equallyDistributeRewards(coins currency.Coin, pools []*DelegatePool, spUpda
...
+		return "", common.NewErrorf("stake_pool_lock_failed",
+			"invalid request: %v", err)
+	}
+
+	var sp AbstractStakePool
+	if sp, err = get(spr.ProviderType, spr.ProviderID, balances); err != nil {
+		return "", common.NewErrorf("stake_pool_lock_failed",
+			"can't get stake pool: %v", err)
+	}
+	if err != nil {
```
peterlimg: The err has already been checked

code/go/0chain.net/smartcontract/storagesc/blobber.go
```diff
@@ -502,32 +502,32 @@ func (sc *StorageSmartContract) commitBlobberRead(t *transaction.Transaction,
...
 	}
 
-	// save pools
-	err = sp.save(spenum.Blobber, commitRead.ReadMarker.BlobberID, balances)
+	// Save pools
+	err = sp.Save(spenum.Blobber, commitRead.ReadMarker.BlobberID, balances)
 	if err != nil {
 		return "", common.NewErrorf("commit_blobber_read",
-			"can't save stake pool: %v", err)
+			"can't Save stake pool: %v", err)
```
peterlimg: Let's not change the 'save' in error messages, so as the other places

code/go/0chain.net/smartcontract/stakepool/stakepool.go
```diff
@@ -52,6 +63,76 @@ type DelegatePool struct {
...
+	Status       string        `json:"status"`
+	RoundCreated int64         `json:"round_created"`
+}
+
+// swagger:model userPoolStat
+type UserPoolStat struct {
+	Pools map[datastore.Key][]*DelegatePoolStat `json:"pools"`
+}
+
+func ToProviderStakePoolStats(miner *event.Provider, delegatePools []event.DelegatePool) (*StakePoolStat, error) {
```
peterlimg: please change param `miner` to `provider`

code/go/0chain.net/smartcontract/dbs/event/blobber_test.go
```diff
@@ -287,42 +290,104 @@ func TestBlobberLatLong(t *testing.T) {
...
-		Name:            os.Getenv("POSTGRES_DB"),
-		User:            os.Getenv("POSTGRES_USER"),
-		Password:        os.Getenv("POSTGRES_PASSWORD"),
-		Host:            os.Getenv("POSTGRES_HOST"),
-		Port:            os.Getenv("POSTGRES_PORT"),
+		Name:            "events_db",
+		User:            "zchain_user",
+		Password:        "zchian",
+		Host:            "localhost",
+		Port:            "5432",
```
peterlimg: Please avoid exposing db username and pass, hard-coded data makes it unable to change if the tests evn use different username/pass

code/go/0chain.net/smartcontract/dbs/event/blobber_test.go
```diff
@@ -287,42 +290,104 @@ func TestBlobberLatLong(t *testing.T) {
...
+			LogoUrl:             "",
+			Description:         "",
+			ChallengesPassed:    0,
+			ChallengesCompleted: 0,
+			OpenChallenges:      0,
+			RankMetric:          0,
+			WriteMarkers:        nil,
+			ReadMarkers:         nil,
+			CreationRound:       0,
+			InactiveRounds:      0,
```
peterlimg: We don't need to set zero value explicitly

code/go/0chain.net/smartcontract/dbs/event/provider.go
```diff
@@ -0,0 +1,21 @@
...
+	CreatedAt      time.Time
+	UpdatedAt      time.Time
+	DelegateWallet string          `json:"delegate_wallet"`
+	MinStake       currency.Coin   `json:"min_stake"`
+	MaxStake       currency.Coin   `json:"max_stake"`
+	NumDelegates   int             `json:"num_delegates"`
+	ServiceCharge  float64         `json:"service_charge"`
+	UnstakeTotal   currency.Coin   `json:"unstake_total"`
+	TotalStake     currency.Coin   `json:"total_stake"`
+	Rewards        ProviderRewards `json:"rewards" gorm:"foreignKey:ProviderID;"`
```
peterlimg: do we need the semicolon in the gorm tag?

peterlimg (changes requested):

service-0chain:
| config  | value |
| tests skipped  | FALSE  |
| system-tests  | fix/staking  |
| gosdk  | fix/staking  |
| zboxcli  | staging  |
| zwalletcli  | fix/staking  |
| 0chain  | fix/staking  |
| blobber  | staging  |
| authorizer  | staging  |
| 0box  | staging  |
| 0block  | staging  |
| 0dns  | staging  |
| explorer  | staging  |
| 0proxy  | staging  |
| 0search  | staging  |

service-0chain:
| config  | value |
| tests skipped  | FALSE  |
| system-tests  | fix/staking  |
| gosdk  | NOT OVERRIDDEN  |
| zboxcli  | staging  |
| zwalletcli  | fix/staking  |
| 0chain  | fix/unstaking  |
| blobber  | staging  |
| authorizer  | staging  |
| 0box  | staging  |
| 0block  | staging  |
| 0dns  | staging  |
| explorer  | master  |
| 0proxy  | staging  |
| 0search  | staging  |

service-0chain:
| config  | value |
| tests skipped  | FALSE  |
| system-tests  | master  |
| gosdk  | NOT OVERRIDDEN  |
| zboxcli  | staging  |
| zwalletcli  | fix/s  |
| 0chain  | fix/staking  |
| blobber  | staging  |
| authorizer  | staging  |
| 0box  | staging  |
| 0block  | staging  |
| 0dns  | staging  |
| explorer  | master  |
| 0proxy  | staging  |
| 0search  | staging  |

code/go/0chain.net/smartcontract/stakepool/stakepool.go
dabasov: we have different saves in different implementations, how can we override them then? interface is useful when you can abstract things out, it is always good to provide it, at least we used to do it java all the time when we have different implementations and we have different implementations here Stakepool, MinerNode etc

dabasov: also it is a good thing to write unit tests with mock stakepool, we can make different issue for it

service-0chain:
| config  | value |
| tests skipped  | FALSE  |
| system-tests  | fix/staking  |
| gosdk  | NOT OVERRIDDEN  |
| zboxcli  | staging  |
| zwalletcli  | fix/staking  |
| 0chain  | fix/staking  |
| blobber  | staging  |
| authorizer  | staging  |
| 0box  | staging  |
| 0block  | staging  |
| 0dns  | staging  |
| explorer  | master  |
| 0proxy  | staging  |
| 0search  | staging  |

peterlimg (approved):
LGTM, please sort out the system test errors

code/go/0chain.net/smartcontract/dbs/event/enums.go
```diff
@@ -86,6 +86,9 @@ const (
 	TagAllocBlobberValueChange
 	TagUpdateBlobberOpenChallenges
 	TagUpdateValidatorStakeTotal
+	TagUpdateMinerTotalStake
+	TagUpdateSharderTotalStake
+	TagUpdateAuthorizerTotalStake
```
Sriep, line 92: Add new tags to initTagString below.

code/go/0chain.net/smartcontract/dbs/event/writemarker.go
```diff
@@ -49,7 +49,7 @@ type WriteMarker struct {
 //	if err := tx.Model(&Blobber{}).Clauses(clause.OnConflict{
 //		Columns:   []clause.Column{{Name: "blobber_id"}},
 //		DoUpdates: clause.Assignments(vs),
-//	}).Create(&Blobber{BlobberID: w.BlobberID, Used: w.Size, SavedData: w.Size}).Error; err != nil {
+//	}).Create(&Blobber{ID: w.ID, Used: w.Size, SavedData: w.Size}).Error; err != nil {
```
Sriep, line 52: Suggest removing commented-out code, not maintaining it.

dabasov: we might need it in the future, let's wait until all snapshot aggregates are tested

Sriep: Leaving commented-out code is fine during development but should be removed before finally merging into the main code stream. If necessary leave a commit hash or link for the old code.

service-0chain:
| config  | value |
| tests skipped  | FALSE  |
| system-tests  | fix/staking  |
| gosdk  | fix/staking  |
| zboxcli  | staging  |
| zwalletcli  | fix/staking  |
| 0chain  | fix/staking  |
| blobber  | staging  |
| authorizer  | staging  |
| 0box  | staging  |
| 0block  | staging  |
| 0dns  | staging  |
| explorer  | staging  |
| 0proxy  | staging  |
| 0search  | staging  |

service-0chain:
| config  | value |
| tests skipped  | FALSE  |
| system-tests  | fix/staking  |
| gosdk  | NOT OVERRIDDEN  |
| zboxcli  | fix/staking  |
| zwalletcli  | fix/staking  |
| 0chain  | fix/staking  |
| blobber  | staging  |
| authorizer  | staging  |
| 0box  | staging  |
| 0block  | staging  |
| 0dns  | staging  |
| explorer  | master  |
| 0proxy  | staging  |
| 0search  | staging  |

service-0chain:
| config  | value |
| tests skipped  | FALSE  |
| system-tests  | fix/staking  |
| gosdk  | NOT OVERRIDDEN  |
| zboxcli  | fix/staking  |
| zwalletcli  | fix/staking  |
| 0chain  | fix/staking  |
| blobber  | staging  |
| authorizer  | staging  |
| 0box  | staging  |
| 0block  | staging  |
| 0dns  | staging  |
| explorer  | master  |
| 0proxy  | staging  |
| 0search  | staging  |

PR #1834, opened 2022-11-24
Event Database settings

Sriep:
## Changes
Adds settings to the event database that mirror global settings in the blockchain.
```go
	Debug           bool
	AggregatePeriod int64
	PageLimit       int64
```
Moved global enums from `minersc` to `chaincore/config`.
## Tests

- blobber:
- gosdk:
- system_test:
- zboxcli:
- zwalletcli:
- actions: https://github.com/0chain/actions/pull/87
- Other: ...


code/go/0chain.net/chaincore/chain/config.go
```diff
@@ -530,7 +530,10 @@ func (c *ConfigImpl) FromViper() error {
 	conf.DbsEvents.MaxIdleConns = viper.GetInt("server_chain.dbs.events.max_idle_conns")
 	conf.DbsEvents.MaxOpenConns = viper.GetInt("server_chain.dbs.events.max_open_conns")
 	conf.DbsEvents.ConnMaxLifetime = viper.GetDuration("server_chain.dbs.events.conn_max_lifetime")
+
+	conf.DbsEvents.Debug = viper.GetBool("server_chain.dbs.events.debug")
 	conf.DbsEvents.AggregatePeriod = viper.GetInt64("server_chain.dbs.events.aggregate_period")
+	conf.DbsEvents.Debug = viper.GetBool("server_chain.dbs.events.page_limit")
```
peterlimg: PageLimit?

code/go/0chain.net/smartcontract/dbs/event/process.go
```diff
@@ -17,8 +18,16 @@ import (
...
+func (edb *EventDb) ProcessEvents(
+	ctx context.Context,
+	events []Event,
+	round int64,
+	block string,
+	blockSize int,
+	cfg config.DbAccess,
+) error {
 	ts := time.Now()
+	edb.updateSettings(cfg)
```
peterlimg: We'd better not update config every time when the ProcessEvents() is called, this function will be called frequently,  so we should update it when necessary, e.g config is changed or initialized, etc.

peterlimg (changes requested):

service-0chain:
| config  | value |
| tests skipped  | FALSE  |
| system-tests  | master  |
| gosdk  | NOT OVERRIDDEN  |
| zboxcli  | staging  |
| zwalletcli  | staging  |
| 0chain  | edb_settings  |
| blobber  | staging  |
| authorizer  | staging  |
| 0box  | staging  |
| 0block  | staging  |
| 0dns  | staging  |
| explorer  | master  |
| 0proxy  | staging  |
| 0search  | staging  |

code/go/0chain.net/chaincore/chain/config.go
Sriep: https://github.com/0chain/0chain/blob/staging/code/go/0chain.net/smartcontract/dbs/event/enums.go#L10
```go
const pageLimit = int64(50)
```
A constant local to the event package that we want to be able to configure. Although we could possibly pass it into event database methods that need it from the handlers.

code/go/0chain.net/smartcontract/dbs/event/process.go
Sriep: I will try adding a round last update field.

code/go/0chain.net/chaincore/chain/config.go
peterlimg: Sorry, I mean shouldn't it be assigned to `conf.DbsEvent.PageLimit` instead of `conf.DbsEvents.Debug`

code/go/0chain.net/chaincore/chain/config.go
```diff
@@ -412,8 +426,11 @@ type ConfigData struct {
 	RoundTimeoutSofttoMult int `json:"softto_mult"`        // multiplier of mean network time for soft timeout
 	RoundRestartMult       int `json:"round_restart_mult"` // multiplier of soft timeouts to restart a round
 
-	DbsEvents config.DbAccess `json:"dbs_event"`
-	TxnExempt map[string]bool `json:"txn_exempt"`
+	DbsEvents   config.DbAccess   `json:"dbs_event"`
+	DbsSettings config.DbSettings `json:"dbs_settings"`
+	TxnExempt   map[string]bool   `json:"txn_exempt"`
+
+	LastUpdateRound int64
```
peterlimg: I'm thinking about if this LastUpdateRound is needed, we will update configs only when finalizing blocks, and basically most of the events are processed when finalizing blocks, so the config should already have been updated then.

code/go/0chain.net/chaincore/chain/protocol_block.go
```diff
@@ -379,7 +379,9 @@ func (c *Chain) finalizeBlock(ctx context.Context, fb *block.Block, bsh BlockSta
 
 	if len(fb.Events) > 0 && c.GetEventDb() != nil {
 		wg.Run("finalize block - add events", fb.Round, func() {
-			if err := c.GetEventDb().ProcessEvents(ctx, fb.Events, fb.Round, fb.Hash, len(fb.Txns)); err != nil {
+			if err := c.GetEventDb().ProcessEvents(
+				ctx, fb.Events, fb.Round, fb.Hash, len(fb.Txns), c.ChainConfig,
```
peterlimg: Can we update the eventDb's config when c.ChainConfig is updated so that we don't have to check if need to update config on every ProcessEvents call

peterlimg (changes requested):

code/go/0chain.net/chaincore/chain/config.go
```diff
@@ -562,71 +564,71 @@ func (c *ConfigImpl) Update(fields map[string]string, version int64) error {
 	logging.Logger.Debug("Updating config", zap.Int64("old version", old), zap.Int64("new version", conf.version))
 
 	var err error
-	conf.IsStateEnabled, err = cf.GetBool(minersc.State)
+	conf.IsStateEnabled, err = cf.GetBool(enums.State)
```
peterlimg, line 567: Let's keep the settings in minersc package as they are highly cohesive, we should avoid putting everything together in the same place. We can move some of them to global config package later but not the `enums` package, the name of `enums` is just not good, we will rename it later.

peterlimg (changes requested):

code/go/0chain.net/chaincore/chain/config.go
Sriep: The logic here was to reduce the danger import cycles due to [(s *DbSettings) Update](https://github.com/0chain/0chain/blob/aa6873abd7b3a3e82f6ff2a82fbe6f1d51a9ff76/code/go/0chain.net/chaincore/config/config.go#L214). 
Ah! so move the enum package into minersc but keep them in a separate package.

peterlimg (approved):

Issue #1700, opened 2022-09-25
Investigate 0chain /storage_config issues


Sriep:
[storagesc.SettingNames ](https://github.com/0chain/0chain/blob/staging/code/go/0chain.net/smartcontract/storagesc/config_settigns.go#L127)max_write_price is repeated

When you do ./zwallet sc-config, which calls /storage-config the following are missing
missing field time_unit, index 1
missing field min_alloc_duration, index 3
missing field max_challenge_completion_time, index 4
missing field min_offer_duration, index 5
missing field stakepool.min_lock_period, index 10
missing field free_allocation_settings.duration, index 17
missing field challenge_enabled, index 31
missing field expose_mpt, index 47
missing field owner_id, index 48
missing field cost, index 49

iamrz1:
@Sriep From my testing only the `stakepool.min_lock_period` was missing because it was not in the hander response, so I have added it. 
I dont see `expose_mpt` and `cost` fields mentioned anywhere in latest staging branch. Could they have been removed?

Issue #1707, opened 2022-09-28
Add initial staking


dabasov:
Initial states should include initial stakes, that will lock ZCN to the stake pools directly bypassing balances in state record.

```yaml
initialStates:
  - id: 0330d8c6c37277b2b5dd112dd41f105703ee5723589d65079a7c13d8659125e0
    tokens: 1123123123
initialLocks:
  - id: 0330d8c6c37277b2b5dd112dd41f105703ee5723589d65079a7c13d8659125e0
    tokens: 11111111

```

PR #1837, opened 2022-11-27
Add missing config and initial staking

iamrz1:
## Fixes
https://github.com/0chain/0chain/issues/1700
https://github.com/0chain/0chain/issues/1707

## Tests

- blobber:
- gosdk:
- system_test:
- zboxcli:
- zwalletcli:
- Other: ...


code/go/0chain.net/chaincore/chain/entity.go
```diff
@@ -611,6 +618,69 @@ func (c *Chain) setupInitialState(initStates *state.InitStates) util.MerklePatri
...
 }
 
+func (c *Chain) addInitialStakes(mpt *util.MerklePatriciaTrie, stakes []state.InitStake) error {
+	var edbDelegatePools []event.DelegatePool
+	for _, v := range stakes {
+		key := v.ProviderType + ":stakepool:" + v.ProviderID
+		sp := stakepool.StakePool{}
+		if err := mpt.GetNodeValue(util.Path(encryption.Hash(key)), &sp); err != nil {
+			if err != util.ErrValueNotPresent {
+				continue
```
peterlimg: Please remember that 'node not found' error should always be reported, so let's return error instead of ignoring it.

code/go/0chain.net/chaincore/chain/entity.go
```diff
@@ -611,6 +618,69 @@ func (c *Chain) setupInitialState(initStates *state.InitStates) util.MerklePatri
...
+
+		_, ok := sp.Pools[v.ClientID]
+		if ok {
+			continue
+		}
+
+		sp.Pools[v.ClientID] = &stakepool.DelegatePool{
+			Balance:      v.Tokens,
+			Reward:       0,
+			Status:       0,
```
peterlimg: We do not need to set zero values explicitly

code/go/0chain.net/chaincore/chain/entity.go
```diff
@@ -611,6 +618,69 @@ func (c *Chain) setupInitialState(initStates *state.InitStates) util.MerklePatri
...
+			PoolID:       v.ClientID,
+			ProviderType: int(spenum.ToProviderID(v.ProviderType)),
+			ProviderID:   v.ProviderID,
+			DelegateID:   v.ClientID,
+			Balance:      v.Tokens,
+			Status:       0,
+			RoundCreated: 0,
+		})
+
+		if _, err := mpt.Insert(util.Path(encryption.Hash(key)), &sp); err != nil {
```
peterlimg: Please call  `sp.Save()` to save the stake pool

code/go/0chain.net/smartcontract/stakepool/spenum/enums.go
```diff
@@ -22,6 +22,15 @@ func (p Provider) String() string {
 	return providerString[p]
 }
 
+func ToProviderID(ps string) Provider {
```
peterlimg: It's provider type,  so should be `ToProviderType`

peterlimg (changes requested):

code/go/0chain.net/chaincore/chain/entity.go
iamrz1: I think the context here is a bit different. stakePool is with a certain key can exist or it can be a new stakepool to be inserted, both are expected in this case (since we are dealing with initial stake pools). The error that I'm skipping is anything except `node not found`, because i'm not sure whether I should report it back as an error, or just print as log. I'm open to suggestion about it.

code/go/0chain.net/chaincore/chain/entity.go
iamrz1: Sure, I can remove them.

code/go/0chain.net/chaincore/chain/entity.go
iamrz1: I am trying to use `sp.Save()`, but can't find a suitable value for this parameter `balances cstate.StateContextI`. Apparently at this point in code `c.GetStateContextI()` returns nil value and causes panic.

code/go/0chain.net/smartcontract/stakepool/spenum/enums.go
iamrz1: Sure.

code/go/0chain.net/chaincore/chain/entity.go
peterlimg: Yes, it's a bit different, but still we need to panic here instead of continue or logging only, it is one of the most dangerous behaviors to skip a 'node not found' error. So please do 
```go
if state.ErrInvalidState(err) {
  logging.Logger.Panic("could not initialize staking", zap.Err(err))
}
```

code/go/0chain.net/chaincore/chain/entity.go
peterlimg: See code in line 581, we can pass in the `stateCtx`

iamrz1: Great! Will do.

code/go/0chain.net/chaincore/chain/entity.go
```diff
@@ -618,43 +618,49 @@ func (c *Chain) setupInitialState(initStates *state.InitStates) util.MerklePatri
...
+	knownStakePoolKeys := map[string]bool{}
 	for _, v := range stakes {
-		key := v.ProviderType + ":stakepool:" + v.ProviderID
+		providerType := spenum.ToProviderType(v.ProviderType)
+		stakePoolKey := stakepool.StakePoolKey(providerType, v.ProviderID)
 		sp := stakepool.StakePool{}
-		if err := mpt.GetNodeValue(util.Path(encryption.Hash(key)), &sp); err != nil {
-			if err != util.ErrValueNotPresent {
-				continue
+		if knownStakePoolKeys[stakePoolKey] {
```
peterlimg: We don't need to create a map to check if the stakePool exist, checking the MPT by calling `sp.Get()` is enough, we won't re-enter this function later after we refactor it to only run genesis block once. So let's keep the changes simple, and avoid exposing the `stakepool.StakePoolKey()` function as well

peterlimg (changes requested):

code/go/0chain.net/chaincore/chain/entity.go
iamrz1: Got it. One of the reasons to add the map was to add up the tokens for stakes with same clientID and and provider. It would also reduce calls to mpt for each entry in stakepool. But if that's fine, I'll  remove it.

peterlimg: We don't need to add up the tokens for stakes, we would only initialize stakes **once** on genesis block, which is round 1. It's a deterministic result, and we shall never change the staked value on genesis block later again. So please remove the `add up token` code and only do existence checking.

iamrz1: okay, so removing code to support adding up stakes for cases as below:
```
  - provider_id: x
    provider_type: blobber
    client_id: y
    tokens: 50000000000

  - provider_id: x
    provider_type: blobber
    client_id: y
    tokens: 10000000000
```
resulting in tokens = 60000000000

peterlimg (approved):

peterlimg (approved):

service-0chain:
| config  | value |
| tests skipped  | TRUE  |
| system-tests  | master  |
| gosdk  | NOT OVERRIDDEN  |
| zboxcli  | staging  |
| zwalletcli  | staging  |
| 0chain  | 0chain-1755  |
| blobber  | staging  |
| authorizer  | staging  |
| 0box  | staging  |
| 0block  | staging  |
| 0dns  | staging  |
| explorer  | master  |
| 0proxy  | staging  |
| 0search  | staging  |

PR #1841, opened 2022-11-28
1797 gorm relationships not loading or loading nulls or empty values in some of the http endpoints

### LINKED ISSUE #1797: Gorm relationships not loading or loading nulls or empty values in some of the http endpoints
The origin was #1659 
We found that some of the endpoints that loads data from eventdb using gorm in the smart contracts return null or empty values for the fields representing relationships. Two possible solutions to this are:
1. Use preloading with these endpoints either by "Preload" or "Join" (https://gorm.io/docs/preload.html)
3. Use smart select (or a filtration function) to select the relevant fields only. (https://gorm.io/docs/advanced_query.html#Smart-Select-Fields)

devyetii:
## Fixes
Some endpoints return empty or null data for relationship fields.
## Changes
I checked all the APIs and fixed the ones needed to be fixed by either Preloading, Joins or selecting the fields that will be exposed to the API
No
## Tests

- blobber:
- gosdk:
- system_test:
- zboxcli:
- zwalletcli:
- Other: ...


code/go/0chain.net/smartcontract/storagesc/handler.go
```diff
@@ -1743,6 +1744,71 @@ type readMarkersCount struct {
...
+		// 	Change: rm.User.Change,
+		// 	Round: rm.User.Round,
+		// 	Nonce: rm.User.Nonce,
+		// },
+
+		// Allocation: &event.Allocation{
+		// 	AllocationID: rm.Allocation.AllocationID,
+		// 	AllocationName: rm.Allocation.AllocationName,
+		// 	TransactionID: rm.Allocation.TransactionID,
+		// }
```
peterlimg: The structs defined in event package should be responded directly, changing APIs response fields could lead to unexpected issues if we don't know whether those fields would be used somewhere by the clients. So please response ReadMarker if nothing was wrong with it.

code/go/0chain.net/smartcontract/storagesc/handler.go
```diff
@@ -2381,7 +2604,7 @@ func (srh *StorageRestHandler) getTransactionByHash(w http.ResponseWriter, r *ht
 		return
 	}
 
-	common.Respond(w, r, transaction, nil)
+	common.Respond(w, r, toAPITransaction(transaction), nil)
```
peterlimg: Please return the raw transaction, we can define different txn info level, but that's another feature, and we should carefully change the API behavior, we can't change it solely in 0chain repo, all client ends API and sdks should be updated as well.

code/go/0chain.net/smartcontract/storagesc/handler.go
```diff
@@ -2953,7 +3176,7 @@ func (srh StorageRestHandler) getSearchHandler(w http.ResponseWriter, r *http.Re
 			return
 		}
 
-		common.Respond(w, r, txn, nil)
+		common.Respond(w, r, toAPITransaction(txn), nil)
```
peterlimg: Return raw transaction

peterlimg (changes requested):

code/go/0chain.net/smartcontract/storagesc/handler.go
devyetii: I understand clearly that this is not a safe thing to do. However, ReadMarkers and WriteMarkers were always returned as "null" no matter what data in the DB, and that was the main issue.

This is what happens in all the other API endpoins for the txn that returns a collection of them. I tried to resolve the issue but in vain, looks like it's a problem with Gorm; but since no body is complaining about it, we can leave it as is for now till we have an open issue (not being lazy but since no body is having problem with it, this means RMs and WMs are not used by the clients of the API consuming these endpoints).

What I did here is that I fixed this behaviour for the endpoint returning a single TXN by hash, and that's by returning only 1 level of relationship from the txn (which is the txn itself, same as the struct, and the non-relational fields of it's direct children which are the RMs and WMs related to it). I want to stress on something that this very endpoint always returned null for RMs and WMs.

peterlimg:
@devyetii I checked the `event.Transaction` again, basically I don't know why we added the `RM` and `WM` to it as they are not being used at all, they are also not parts of a transaction.  And there just no usage of them and no values assigned to them, that's why the API always return null for them. So can you help to remove them, we don't need them.

devyetii:
> @devyetii I checked the `event.Transaction` again, basically I don't know why we added the `RM` and `WM` to it as they are not being used at all, they are also not parts of a transaction. And there just no usage of them and no values assigned to them, that's why the API always return null for them. So can you help to remove them, we don't need them.

@peterlimg  I think they were added to setup the foreignKey on `transaction_id` field of `read_markers` and `write_markers` table. I can remove them from the json returned at the endpoints but I guess it'll be good to keep them in the Model.

code/go/0chain.net/smartcontract/dbs/event/transaction.go
```diff
@@ -104,7 +132,9 @@ func (edb *EventDb) GetTransactionByBlockNumbers(blockStart, blockEnd int64, lim
 func (edb *EventDb) GetTransactionsForBlocks(blockStart, blockEnd int64) ([]Transaction, error) {
 	tr := []Transaction{}
 	res := edb.Store.Get().
-		Model(Transaction{}).
+		Model(&Transaction{}).
+		Preload("ReadMarkers").
+		Preload("WriteMarker").
```
peterlimg: Unnecessary to preload if these field are not being used

code/go/0chain.net/smartcontract/storagesc/handler.go
```diff
@@ -1805,7 +1870,12 @@ func (srh *StorageRestHandler) getReadMarkers(w http.ResponseWriter, r *http.Req
 		return
 	}
 
-	common.Respond(w, r, readMarkers, nil)
+	rms := make([]readMarkerResponse, len(readMarkers))
+	for _, rm := range readMarkers {
+		rms = append(rms, toReadMarkerResponse(rm))
+	}
```
peterlimg, line 1854: The writeMarkers query API should also be updated to to keep consistence.

peterlimg (changes requested):

peterlimg:
> > @devyetii I checked the `event.Transaction` again, basically I don't know why we added the `RM` and `WM` to it as they are not being used at all, they are also not parts of a transaction. And there just no usage of them and no values assigned to them, that's why the API always return null for them. So can you help to remove them, we don't need them.
> 
> @peterlimg I think they were added to setup the foreignKey on `transaction_id` field of `read_markers` and `write_markers` table. I can remove them from the json returned at the endpoints but I guess it'll be good to keep them in the Model.

The API handler is the only place that use the RM and WM in a transaction, since we don't return them, so no one will use them actually. Basically the RM and WM refs are for getting all the RMs and WMs within a transaction, if we don't need them, we can simply remove them. BTW, they are not being used since the beginning.

devyetii:
> 

@peterlimg I was just concerned with the foreign key on `read_markers.transaction_id` that I thought will be removed if I removed `Transaction.ReadMarkers` (same for WMs) since GORM auto-migrates the schema from Model definitions. I realized that we have `ReadMarker.TransactionID` which also defines that FK so it's safe to be removed. Thanks @peterlimg

code/go/0chain.net/smartcontract/storagesc/handler.go
```diff
@@ -2208,7 +2237,13 @@ func (srh *StorageRestHandler) getWriteMarker(w http.ResponseWriter, r *http.Req
 		common.Respond(w, r, nil, common.NewErrInternal(err.Error()))
 		return
 	}
-	common.Respond(w, r, rtv, nil)
+
+	wmrs := make([]WriteMarkerResponse, len(rtv))
```
peterlimg: Same as above

code/go/0chain.net/smartcontract/storagesc/handler.go
```diff
@@ -1870,12 +1848,12 @@ func (srh *StorageRestHandler) getReadMarkers(w http.ResponseWriter, r *http.Req
 		return
 	}
 
-	rms := make([]readMarkerResponse, len(readMarkers))
+	rmrs := make([]ReadMarkerResponse, len(readMarkers))
```
peterlimg: Should be `rmrs := make([]ReadMarkerResponse, 0, len(readMarkers))` if we use append to add items below.

peterlimg (changes requested):
A minor issue need to fix, all others looks good to me, thanks @devyetii

peterlimg (approved):
LGTM

service-0chain:
| config  | value |
| tests skipped  | FALSE  |
| system-tests  | readmarkers  |
| gosdk  | NOT OVERRIDDEN  |
| zboxcli  | staging  |
| zwalletcli  | staging  |
| 0chain  | 1797-gorm-relationships-not-loading-or-loading-nulls-or-empty-values-in-some-of-the-http-endpoints  |
| blobber  | staging  |
| authorizer  | staging  |
| 0box  | staging  |
| 0block  | staging  |
| 0dns  | staging  |
| explorer  | master  |
| 0proxy  | staging  |
| 0search  | staging  |

PR #1777, opened 2022-10-29
New Rewards table for event database

Sriep:
## Fixes
closes https://github.com/0chain/0chain/issues/831, closes https://github.com/0chain/0chain/issues/1758
## Changes
1. This adds new tables to the event database to hold each instance of a provider or delegate reward. This is primarily so that system tests can confirm token payments match expectations.
2. There is indications that the new table slows down event processing noticeably. To make this optional, a new debug option has been added to `0chain.yaml` and the event database. Having this set to false will disable saving each reward.
3. Two new minersc query endpoints have been added, `/provider-rewards` and `/delegate-rewards` to return information form the two tables.
## Tests
Tasks to complete before merging PR:
- [ ]  Ensure system tests are passing. If not [Run them manually](https://github.com/0chain/0chain/actions/workflows/system_tests.yml) to check for any regressions :clipboard:
- [ ]  Do any new system tests need added to test this change? do any existing system tests need updated? If so create a PR at [0chain/system_test](https://github.com/0chain/system_test)
- [ ]  Merge your system tests PR to master AFTER merging this PR

## Associated PRs (Link as appropriate):
- blobber:
- gosdk:
- system_test:
- zboxcli:
- zwalletcli:
- Other: ...

PR #1847, opened 2022-11-29
New reward table - requires Pr1834

Sriep:
## Changes
- This adds new tables to the event database to hold each instance of a provider or delegate reward. This is primarily so that system tests can confirm token payments match expectations.
- There is indications that the new table slows down event processing noticeably. To make this optional, a new debug option has been added to 0chain.yaml and the event database. Having this set to false will disable saving each reward.
- Two new minersc query endpoints have been added, /provider-rewards and /delegate-rewards to return information form the two tables.
## Tests

- blobber:
- gosdk:
- system_test:
- zboxcli:
- zwalletcli:
- Other: ...
- Requires https://github.com/0chain/0chain/pull/1834, add debug check in  rewardUpdate after merged.
- Replaces https://github.com/0chain/0chain/pull/1777


code/go/0chain.net/smartcontract/minersc/handler.go
```diff
@@ -58,9 +58,105 @@ func GetEndpoints(rh rest.RestHandlerI) []rest.Endpoint {
...
+	}
+	common.Respond(w, r, rtv, nil)
+}
+
+// swagger:route GET /v1/screst/6dba10422e368813802877a85039d3985d96760ed844092319743fb3a76712d7/provider-rewards provider-rewards
+// Gets list of provider rewards satisfying filter
+//
+// parameters:
+//
+//	+name: offset
```
boddumanohar, line 127: as metioned in [line 135](https://github.com/0chain/0chain/pull/1847/files#diff-7b517d8ed30356228fd95e0be8b591935a1291434d6cee2cc8e98935622d8862R136) the endpoint also takes `start` and `end` of the block as query parameters

code/go/0chain.net/smartcontract/minersc/handler.go
```diff
@@ -58,9 +58,105 @@ func GetEndpoints(rh rest.RestHandlerI) []rest.Endpoint {
 		rest.MakeEndpoint(miner+"/configs", common.UserRateLimit(mrh.getConfigs)),
 		rest.MakeEndpoint(miner+"/get_miner_geolocations", common.UserRateLimit(mrh.getMinerGeolocations)),
 		rest.MakeEndpoint(miner+"/get_sharder_geolocations", common.UserRateLimit(mrh.getSharderGeolocations)),
+		rest.MakeEndpoint(miner+"/provider-rewards", common.UserRateLimit(mrh.getProviderRewards)),
+		rest.MakeEndpoint(miner+"/delegate-rewards", common.UserRateLimit(mrh.getDelegateRewards)),
 	}
 }
 
+// swagger:route GET /v1/screst/6dba10422e368813802877a85039d3985d96760ed844092319743fb3a76712d7/delegate-rewards delegate-rewards
+// Gets list of delegate rewards satisfying filter
```
boddumanohar, line 68: please use the minerSC REST endpoint URL
`v1/screst/6dba10422e368813802877a85039d3985d96760ed844092319743fb3a76712d9/delegate-rewards`

Sriep: I don't understand, you are saying change so that the endpoints prefix is the storagesc address rather than the minersc address. In this case, we should move the endpoints to storagesc, not just change the endpoint address. 

This endpoint covers data from all providers, so it does not really belong anywhere. At some point, we should have a new address to cover handlers that service all providers. But that is beyond the scope of this PR.

Sriep:
Replace with https://github.com/0chain/0chain/pull/1870

PR #1870, opened 2022-12-06
Add delegates pools to getNodeStat

Sriep:
## Fixes
- This adds new tables to the event database to hold each instance of a provider or delegate reward. This is primarily so that system tests can confirm token payments match expectations.
- There is indications that the new table slows down event processing noticeably. To make this optional, a new debug option has been added to 0chain.yaml and the event database. Having this set to false will disable saving each reward.
- Two new minersc query endpoints have been added, /provider-rewards and /delegate-rewards to return information form the two tables.
- Allows stake pools to be returned with `nodeStat`, using  `include_delegates=true` parameter.
- Fix sharder reward payments in payFees. 
## Tests

- blobber:
- gosdk:
- system_test:
- zboxcli:
- zwalletcli:
- Other: ...
Replaces https://github.com/0chain/0chain/pull/1847


code/go/0chain.net/smartcontract/minersc/handler.go
```diff
@@ -292,10 +293,19 @@ func (mrh *MinerRestHandler) getNodeStat(w http.ResponseWriter, r *http.Request)
 		common.Respond(w, r, nil, common.NewErrInternal("cannot get latest finalised block"))
 		return
 	}
+	var err error
+	var delegates []event.DelegatePool
+	if includeDelegates {
+		delegates, err = edb.GetDelegatePools(id)
+		if err != nil {
+			common.Respond(w, r, nil, common.NewErrInternal("getting delegates"+err.Error()))
```
peterlimg, line 418: We should not return internal error If the err is something like gorm.ErrRecordNotFound, which means the provider has no delegates, we can just return empty pools with the node stat to user.

peterlimg (dismissed):

code/go/0chain.net/smartcontract/minersc/handler.go
Sriep: I tested on an empty table, no error was returned just an empty slice. So I don't think this is an issue.

service-0chain:
| config  | value |
| tests skipped  | FALSE  |
| system-tests  | block_rewards  |
| gosdk  | NOT OVERRIDDEN  |
| zboxcli  | staging  |
| zwalletcli  | staging  |
| 0chain  | add_delegates  |
| blobber  | staging  |
| authorizer  | staging  |
| 0box  | staging  |
| 0block  | staging  |
| 0dns  | staging  |
| explorer  | master  |
| 0proxy  | staging  |
| 0search  | staging  |

dabasov (approved):
lgtm

PR #1875, opened 2022-12-07
Restarting miners doesn't reinitialize state

### LINKED ISSUE #1755: Restart miner/node may re-Initial state on genesis block
See code below, the smart contract balance will be reset to the value in `init_state.yaml` when nodes (miner/sharder) get restarted, rather we should check if the balance node does exist first to avoid resetting on restarting.
https://github.com/0chain/0chain/blob/047a6fa6d55a7a8dff3e088668abfc633e604cf6/code/go/0chain.net/chaincore/chain/entity.go#L564-L568

K-Kumar-01:
## Changes
- check for the current round number or lfb
- initialize only when the round is zero or LFB is nil

Details about state can be mentioned.

## Tests

### DIFF
--- a/code/go/0chain.net/miner/miner/miner.go
+++ b/code/go/0chain.net/miner/miner/miner.go
@@ -148,8 +148,15 @@ func main() {
 	// TODO: put it in a better place
 	go mc.StartLFMBWorker(ctx)
 
-	gb := mc.SetupGenesisBlock(viper.GetString("server_chain.genesis_block.id"),
-		magicBlock, initStates)
+	var gb *block.Block
+
+	lfb := miner.GetLatestBlockFromSharders(ctx)
+
+	if mc.GetCurrentRound() == 0 || lfb == nil {
+		gb = mc.SetupGenesisBlock(viper.GetString("server_chain.genesis_block.id"),
+			magicBlock, initStates)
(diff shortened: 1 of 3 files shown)


code/go/0chain.net/miner/miner/miner.go
```diff
@@ -148,8 +148,15 @@ func main() {
 	// TODO: put it in a better place
 	go mc.StartLFMBWorker(ctx)
 
-	gb := mc.SetupGenesisBlock(viper.GetString("server_chain.genesis_block.id"),
-		magicBlock, initStates)
+	var gb *block.Block
+
+	lfb := miner.GetLatestBlockFromSharders(ctx)
+
+	if mc.GetCurrentRound() == 0 || lfb == nil {
```
peterlimg: We should only check the current round, the lfb would be nil before current round is >= 3, however we should only initialize genesis block once.

peterlimg (changes requested):

code/go/0chain.net/miner/miner/miner.go
K-Kumar-01: @peterlimg updated.

peterlimg (approved):
LGTM

PR #1888, opened 2022-12-12
Fix/snap create tim

manasnisar:
## Tests

- blobber:
- gosdk:
- system_test:
- zboxcli:
- zwalletcli:
- Other: ...

### DIFF
--- a/code/go/0chain.net/smartcontract/dbs/event/snapshot.go
+++ b/code/go/0chain.net/smartcontract/dbs/event/snapshot.go
@@ -42,6 +42,7 @@ type Snapshot struct {
 	UniqueAddresses      int64 `json:"unique_addresses"`      // Total unique address
 	BlockCount           int64 `json:"block_count"`           // Total number of blocks currently
 	AverageTxnFee        int64 `json:"avg_txn_fee"`           // Average transaction fee per block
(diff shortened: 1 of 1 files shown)


code/go/0chain.net/smartcontract/dbs/event/snapshot.go
```diff
@@ -42,6 +42,7 @@ type Snapshot struct {
 	UniqueAddresses      int64 `json:"unique_addresses"`      // Total unique address
 	BlockCount           int64 `json:"block_count"`           // Total number of blocks currently
 	AverageTxnFee        int64 `json:"avg_txn_fee"`           // Average transaction fee per block
+	CreatedAt            int64 `gorm:"autoCreateTime" json:"created_at"` // Snapshot creation date
```
boddumanohar, line 45: this will be the time when the snapshot is added to 0box DB. This might not reflect the exact time when the snapshot is created. You  might have to get the snapshot created time from sharders

dabasov: it's on sharders already, hence this time will be slightly different from round time it will be pretty close and we should never see the difference

dabasov (approved):

PR #1890, opened 2022-12-13
Fix chain slow on chaos and round timeout

peterlimg:
## Fixes
- Chain would stuck for a while till round timeout occurs on small network (3 miners and 3 sharders) when chaos is enabled. This happens when one of the generator is down, and the alive generator have 'node not found' error. We did quick node states syncing for block verification before, but didn't do it for block generation process. So when this happens, we will have to wait for the round restart to re-generate the block. The fix is simply do the quick state syncing and retry just like we did for the block verification.
- Fix VRF tickets cache related slow consensus issue. See details in commit description in [47826](https://github.com/0chain/0chain/pull/1890/commits/47826062e6e90b20c5c64ef0e6fe1370724b2ab4)

## Tests

- blobber:
- gosdk:
- system_test:
- zboxcli:
- zwalletcli:
- Other: ...

### DIFF
--- a/code/go/0chain.net/miner/protocol.go
+++ b/code/go/0chain.net/miner/protocol.go
@@ -1,8 +1,9 @@
 package miner
 
 import (
-	"0chain.net/core/datastore"
 	"context"
+
+	"0chain.net/core/datastore"
 
 	"0chain.net/chaincore/chain"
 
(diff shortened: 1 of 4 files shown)


dabasov (approved):
lgtm

PR #1903, opened 2022-12-16
Ensure blobbers and validators have different Ids

### LINKED ISSUE #1936: Providers should have unique ids.
0chian should force all providers to have unique ids.

Sriep:
## Fixes
- Prevents creating blobbers and validators with the same id. 
- Uses the same MPT key for blobbers and validators. Adds a type field to blobbers and validators to facilitate this.
- Adds a new package `provider` which just holds the MPT key for providers ATM. This will be expanded upon in the upcoming kill/shutdown PR.

## Tests

- blobber:
- gosdk:
- system_test: https://github.com/0chain/system_test/pull/535
- zboxcli:
- zwalletcli:
- Other: ...


code/go/0chain.net/smartcontract/storagesc/models.go
```diff
@@ -197,22 +199,23 @@ func (sc *StorageChallenge) Save(state cstate.StateContextI, scAddress string) e
 type ValidationNode struct {
 	ID                string             `json:"id"`
 	BaseURL           string             `json:"url"`
+	ProviderType      spenum.Provider    `json:"provider_type"`
```
dabasov: Validation node is always a validator, why do you need to add this field?

code/go/0chain.net/smartcontract/storagesc/models.go
```diff
@@ -335,6 +338,7 @@ type Info struct {
 type StorageNode struct {
 	ID                      string                 `json:"id"`
 	BaseURL                 string                 `json:"url"`
+	ProviderType            spenum.Provider        `json:"provider_type"`
```
dabasov: storage node is always a blobber, why to add this field?

dabasov (changes requested):

code/go/0chain.net/smartcontract/storagesc/models.go
Sriep: The main point is that we are using the same MPT key for all providers. 
```go
func GetKey(id string) datastore.Key {
	return "provider:" + id
}
```

So when the user tries to say add a validator with the same id as an already existing blobber, then the sc will get the blobber data and put it into a validator object. By the error message returned, we can tell whether `add_validator` should fail because there is already a validoator with the same id or a blobber with that id.

In that way, it is easy to ensure that all providers have different ids.

Later we might be only interested in looking at the Provider part but still want to know what we are looking at. The provider part being the same object for all providers..

code/go/0chain.net/smartcontract/storagesc/models.go
Sriep: As above

code/go/0chain.net/smartcontract/storagesc/blobber.go
```diff
@@ -31,6 +31,10 @@ func getBlobber(
 	if err != nil {
 		return nil, err
 	}
+	if blobber.ProviderType != spenum.Blobber {
+		logging.Logger.Info("piers getBlobber", zap.Any("blobber", blobber))
```
dabasov: change logging

code/go/0chain.net/smartcontract/storagesc/models.go
```diff
@@ -197,22 +199,23 @@ func (sc *StorageChallenge) Save(state cstate.StateContextI, scAddress string) e
...
+		return errors.New("invalid vali1dator base url")
 	}
 
 	return
 }
 
-func (sn *ValidationNode) GetKey(globalKey string) datastore.Key {
-	return datastore.Key(globalKey + "validator:" + sn.ID)
+func (sn *ValidationNode) GetKey(_ string) datastore.Key {
+	return provider.GetKey(sn.ID)
```
peterlimg, line 217: Please do not remove the globalKey(SC address), it helps to differ the keys between SCs. Removing it increase the possibility to get conflict keys.

code/go/0chain.net/smartcontract/storagesc/validator.go
```diff
@@ -19,12 +22,13 @@ func (sc *StorageSmartContract) addValidator(t *transaction.Transaction, input [
 	}
 	newValidator.ID = t.ClientID
 	newValidator.PublicKey = t.PublicKey
+	newValidator.ProviderType = spenum.Validator
```
peterlimg, line 25: We don't need to add this filed to differ the different provider type, instead it's the `GetKey()` method that matters. So the format of key can be `${SCAddress}:${provider_type}:${ID}` instead of `provider:${ID}`.   Otherwise, if blobber and validator got the same ID, getting node from the same `key` will cause problem.

peterlimg (changes requested):

code/go/0chain.net/smartcontract/storagesc/models.go
Sriep: The whole point is to increase the chance of conflicting keys, so we can detect different providers with the same id, and reject them before they get created.

code/go/0chain.net/smartcontract/storagesc/validator.go
Sriep: The idea is to prevent blobbers and validators from having the same id, so we are deliberately creating conflicts so we can prevent them from being accepted. Is there a reason why we want to allow blobbers and validators to have the same id?

code/go/0chain.net/smartcontract/storagesc/blobber.go
```diff
@@ -31,6 +31,10 @@ func getBlobber(
 	if err != nil {
 		return nil, err
 	}
+	if blobber.ProviderType != spenum.Blobber {
```
peterlimg, line 47: It's impossible to get a storageNode from MPT that has different provider type other than `spenum.Blobber`, if we want to check if the same ID is used for other providers, we will then need to check `ValidatorNode`, `MinerNode`, `AuthorizerNode`, etc. But that will make the code ugly. Anyway we can do it in the following way
1. Define a `Provider` struct in the `provider` package with struct like:
```go
type Provider struct {
    ID string `json:"id"`
    ProviderType spenum.Provider `json:"provider_type"`
}
```
2. Embed the `Provider` to all of those `StorageNode`, `ValidatorNode`...
3. When need to check provider type, we can do:
```go
var p provider.Provider
balances.GetTrieNode(provider.GetKey(id), &p)
```
Then check if the provider type is the correct one.

peterlimg (changes requested):

code/go/0chain.net/smartcontract/storagesc/blobber.go
Sriep: My aim was to reduce the number of MPT lookups. 

If I understand correctly the above suggests two lookups, first, check existence and type. If everything is ok, then do a second lookup to get the actual object.

I have been trying for a solution that only does one MPT access. On asking for a blobber it tries to read the provider into a StorageNode object. 
1. If it actually is a blobber everything ok.
2. If not as many as possable of the fields with matching `json` tags will be copied across. If we make sure all providers have a `json:"provider_type"` field, then this should be filled in. It then makes sense to do the above check:
```go
if blobber.ProviderType != spenum.Blobber {
```

If we are not actually looking at a blobber, the ProviderType field should still be populated as it has the saem `json` tag.

service-0chain:
| config  | value |
| tests skipped  | FALSE  |
| system-tests  | master  |
| gosdk  | NOT OVERRIDDEN  |
| zboxcli  | staging  |
| zwalletcli  | staging  |
| 0chain  | proivder_ids  |
| blobber  | staging  |
| authorizer  | staging  |
| 0box  | staging  |
| 0block  | staging  |
| 0dns  | staging  |
| explorer  | master  |
| 0proxy  | staging  |
| 0search  | staging  |

code/go/0chain.net/smartcontract/storagesc/validator.go
dabasov: @Sriep I agree with @peterlimg, we can simply generate key in a different way, for example
```go
validator.GetKey(ADDRESS)
```
can be 
```go

func (sn *ValidationNode) GetKey(_ string) datastore.Key {
	return provider.GetKey(sn.ID, spenum.Validator)
}
```
and then as peter wrote ``key := ${SCAddress}:${provider_type}:${ID} ``
as we fully control writings to the MPT it should be safe

code/go/0chain.net/smartcontract/storagesc/blobber.go
peterlimg: Okay, I got what you mean. You solution should be able to work, but keep an eye on the error that caused by incompatible provider objects on msgp decoding process, which could happen if two different providers has the same field name but with different data type. We may not encounter this error atm, but could have when developer add new fields with different data type to providers.

code/go/0chain.net/smartcontract/storagesc/models.go
```diff
@@ -195,24 +197,24 @@ func (sc *StorageChallenge) Save(state cstate.StateContextI, scAddress string) e
...
+	StakePoolSettings  stakepool.Settings `json:"stake_pool_settings"`
 }
 
 // validate the validator configurations
-func (sn *ValidationNode) validate(conf *Config) (err error) {
+func (sn *ValidationNode) validate(_ *Config) (err error) {
 	if strings.Contains(sn.BaseURL, "localhost") &&
 		node.Self.Host != "localhost" {
-		return errors.New("invalid validator base url")
+		return errors.New("invalid vali1dator base url")
```
peterlimg: should be `validator`

docker.local/config/0chain.yaml
```diff
@@ -141,7 +141,7 @@ server_chain:
       conn_max_lifetime: 20s
     settings:
       # event database settings blockchain
-      debug: false
+      debug: true
```
peterlimg: Change to false before merging

code/go/0chain.net/smartcontract/storagesc/models.go
```diff
@@ -369,8 +371,8 @@ func (sn *StorageNode) validate(conf *Config) (err error) {
 	return
 }
 
-func (sn *StorageNode) GetKey(globalKey string) datastore.Key {
-	return datastore.Key(globalKey + sn.ID)
+func (sn *StorageNode) GetKey(_ string) datastore.Key {
```
peterlimg: We should remove the param if it's not used.

code/go/0chain.net/smartcontract/storagesc/models.go
```diff
@@ -332,7 +334,7 @@ type Info struct {
 
 // StorageNode represents Blobber configurations.
 type StorageNode struct {
-	ID                      string                 `json:"id"`
+	*provider.Provider      `json:"provider"`
```
peterlimg: We don't have to use pointer embed because the provider field must not be nil, also use value embed could avoid the field changed mistakenly by reference. Same as other providers.

peterlimg (changes requested):

code/go/0chain.net/smartcontract/storagesc/models.go
Sriep: Using value embed will mean that any future `Provider` methods will likely need to be defined on `Provider` rather than `*Provider`, or we will need to be systematic about always using pointer objects for handling providers. Although using value embed might be the better evil?

code/go/0chain.net/smartcontract/storagesc/validator.go
dabasov: can we remove this field then?

service-0chain:
| config  | value |
| tests skipped  | FALSE  |
| system-tests  | temp_test  |
| gosdk  | NOT OVERRIDDEN  |
| zboxcli  | staging  |
| zwalletcli  | staging  |
| 0chain  | proivder_ids  |
| blobber  | staging  |
| authorizer  | staging  |
| 0box  | staging  |
| 0block  | staging  |
| 0dns  | staging  |
| explorer  | master  |
| 0proxy  | staging  |
| 0search  | staging  |

code/go/0chain.net/smartcontract/storagesc/validator.go
Sriep: We need the provider type field so we can tell what provider type an object is, as we can no longer deduce it from the MPT key.

`key := ${SCAddress}:${provider_type}:${ID}` This does not do what we want as different providers will have different keys due to ` ${SCAddress}:${provider_type}:` part being different for different provider types. 

For example, if a blobber and a miner have the same id we want the MPT key to be the same so a conflict is immediately flagged. We can then easily prevent the creation of a second provider with the same id as an existing provider.

Sriep:
```
 Otherwise, if blobber and validator got the same ID, getting node from the same key will cause problem.
```
This should be impossible, as this PR will prevent blobbers and validators from ever having the same id.

The whole point of this PR is to create this conflict so that we can prevent different providers from having the same id. 

When we try to create a second provider with the same id as an already existing provider, the conflict is immediately detected, and the second provider cannot be created, returning an error.

service-0chain:
| config  | value |
| tests skipped  | FALSE  |
| system-tests  | temp_test  |
| gosdk  | NOT OVERRIDDEN  |
| zboxcli  | staging  |
| zwalletcli  | staging  |
| 0chain  | proivder_ids  |
| blobber  | staging  |
| authorizer  | staging  |
| 0box  | staging  |
| 0block  | staging  |
| 0dns  | staging  |
| explorer  | master  |
| 0proxy  | staging  |
| 0search  | staging  |

service-0chain:
| config  | value |
| tests skipped  | FALSE  |
| system-tests  | proivder_ids  |
| gosdk  | NOT OVERRIDDEN  |
| zboxcli  | staging  |
| zwalletcli  | staging  |
| 0chain  | proivder_ids  |
| blobber  | staging  |
| authorizer  | staging  |
| 0box  | staging  |
| 0block  | staging  |
| 0dns  | staging  |
| explorer  | master  |
| 0proxy  | staging  |
| 0search  | staging  |

service-0chain:
| config  | value |
| tests skipped  | FALSE  |
| system-tests  | proivder_ids  |
| gosdk  | NOT OVERRIDDEN  |
| zboxcli  | staging  |
| zwalletcli  | staging  |
| 0chain  | proivder_ids  |
| blobber  | staging  |
| authorizer  | staging  |
| 0box  | staging  |
| 0block  | staging  |
| 0dns  | staging  |
| explorer  | master  |
| 0proxy  | staging  |
| 0search  | staging  |

service-0chain:
| config  | value |
| tests skipped  | FALSE  |
| system-tests  | master  |
| gosdk  | NOT OVERRIDDEN  |
| zboxcli  | staging  |
| zwalletcli  | staging  |
| 0chain  | proivder_ids  |
| blobber  | staging  |
| authorizer  | staging  |
| 0box  | staging  |
| 0block  | staging  |
| 0dns  | staging  |
| explorer  | master  |
| 0proxy  | staging  |
| 0search  | staging  |

peterlimg (approved):
LGTM

dabasov (approved):
lgtm

PR #1908, opened 2022-12-18
Fix initial_state.yaml 

Sriep:
## Fixes
Fix local chains stall at round 1.
## Tests

- blobber:
- gosdk:
- system_test:
- zboxcli:
- zwalletcli:
- Other: ...


Sriep:
Fix another way
```
removing the config is not the way to fix it as we need it, we need to figure out why it get stall. 
I checked the issue, it’s the different [creation date](https://github.com/0chain/0chain/blob/512ba4aeaedc0b5aa6e85833eccafb868119aa69/code/go/0chain.net/chaincore/chain/entity.go#L650) 
on miners that caused the genesis block state hash mismatch error. 
We need to remove the creation date assignment actually, can you do that or I’ll send a PR to fix it.
```

PR #1919, opened 2022-12-21
Fix unlock pool

peterlimg:
## Fixes

- Fix forever intermediate deleting status on unlocking pool

## Tests

- blobber:
- gosdk:
- system_test:
- zboxcli:
- zwalletcli:
- Other: ...

### DIFF
--- a/code/go/0chain.net/smartcontract/minersc/miner.go
+++ b/code/go/0chain.net/smartcontract/minersc/miner.go
@@ -48,13 +48,12 @@ func (msc *MinerSmartContract) AddMiner(t *transaction.Transaction,
 	defer lockAllMiners.Unlock()
 
 	magicBlockMiners := balances.GetChainCurrentMagicBlock().Miners
-	
+
 	if magicBlockMiners.HasNode(newMiner.ID) == false {
 		logging.Logger.Error("add_miner: Error in Adding a new miner: Not in magic block")
 		return "", common.NewErrorf("add_miner",
 			"failed to add new miner: Not in magic block")
 	}
-	
 
 	logging.Logger.Info("add_miner: try to add miner", zap.Any("txn", t))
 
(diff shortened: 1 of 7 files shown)


code/go/0chain.net/smartcontract/minersc/miner.go
```diff
@@ -193,28 +193,20 @@ func (msc *MinerSmartContract) deleteNode(
...
 	case NodeTypeSharder:
-		nodeType = spenum.Sharder
 	default:
 		return nil, fmt.Errorf("unrecognised node type: %v", deleteNode.NodeType.String())
 	}
 
-	for key, pool := range deleteNode.Pools {
+	for _, pool := range deleteNode.Pools {
 		switch pool.Status {
-		case spenum.Pending:
```
peterlimg, line 210: miner/sharder ndoes do not have pending status anymore, they will be in active status when staked

code/go/0chain.net/smartcontract/minersc/miner.go
```diff
@@ -193,28 +193,20 @@ func (msc *MinerSmartContract) deleteNode(
 ) (*MinerNode, error) {
 	var err error
 	deleteNode.Delete = true
-	var nodeType spenum.Provider
 	switch deleteNode.NodeType {
```
dabasov: it's better to have
```go
if deleteNode.NodeType != NodeTypeMiner || eleteNode.NodeType != NodeTypeSharder {
    return nil, fmt.Errorf("unrecognised node type: %v", deleteNode.NodeType.String())
}
```

code/go/0chain.net/smartcontract/minersc/miner.go
```diff
@@ -193,28 +193,20 @@ func (msc *MinerSmartContract) deleteNode(
...
-				pool.DelegateID, nodeType, key, balances)
-			if err != nil {
-				return nil, fmt.Errorf("error emptying delegate pool: %v", err)
-			}
 		case spenum.Active:
 			pool.Status = spenum.Deleting
-		case spenum.Deleting:
-		case spenum.Deleted:
+		case spenum.Pending,
+			spenum.Deleting,
```
dabasov: why can't we completely remove Deleting status? I don't see a lot of logic built on this status, most likely ``Unstaking`` is what we need for 2 phase unlocking and use. Deleting is used in ``payFees`` and ``MintRewards`` but is doesn't look as something important.

dabasov (changes requested):

code/go/0chain.net/smartcontract/minersc/miner.go
peterlimg: The switch case is more readable, also we can simply add new valid type with a case instead of adding the long `|| deleteNode.NodeType != something`.

code/go/0chain.net/smartcontract/minersc/miner.go
peterlimg: Yes, it should be safe to be removed completely

dabasov (approved):
lgtm

PR #1955, opened 2022-12-30
Fix sharder missing blocks

peterlimg:
## Fixes
- Set `chain.LatestFinalizedBlock` only when finalize block process all success
- Rewrite the `round.Clone()` method to avoid potential race condition caused by the copying of locks
- Return error instead  of continue if save changes failed, do stricter FB checking.

## Tests

- blobber:
- gosdk:
- system_test:
- zboxcli:
- zwalletcli:
- Other: ...


code/go/0chain.net/chaincore/chain/worker.go
```diff
@@ -370,7 +370,7 @@ func (c *Chain) finalizeBlockProcess(ctx context.Context, fb *block.Block, bsh B
 		}
 
 		prevBlockHash := pr.GetBlockHash()
-		if prevBlockHash == "" {
+		if prevBlockHash == "" || !pr.IsFinalized() {
```
dabasov, line 373: previous round is not always a finalized round, if we had more than one finalizations on the previous round, then it can't be finalized and rounds should be finalized by chain later on, when single distinct round will be chosen, it will be done in method `c.finalizeBlock(ctx, fb, bsh)`

code/go/0chain.net/sharder/chain.go
```diff
@@ -206,8 +206,7 @@ func (sc *Chain) setupLatestBlocks(ctx context.Context, bl *blocksLoaded) (
 	}
 
 	sc.SetRandomSeed(bl.r, bl.r.GetRandomSeed())
-	bl.r.Block = bl.lfb
-	bl.r.BlockHash = bl.lfb.Hash
+	bl.r.Finalize(bl.lfb)
```
dabasov, line 210: why should we take care of finalization here, is it because we could crash in some inconsistent (partial) state last time and started with not round, which processing was not finished properly?

dabasov (changes requested):

code/go/0chain.net/chaincore/chain/worker.go
peterlimg: I think what you mean is the [chain.FinalizeRound](https://github.com/0chain/0chain/blob/d6cb4dc775b7cdd9ec2f78920d04e18148ad3167/code/go/0chain.net/chaincore/chain/protocol_round_main.go#L10), we run it whenever a block is notarized, and we don't require its previous round is finalized. But this finalize process is the last step to finalize a round, there's no later on actually, on sharders, its previous round must have been finalized, otherwise, there will have FB gaps.

code/go/0chain.net/sharder/chain.go
peterlimg: It only set the block as finalized, nothing else actually. The reason to set it as finalized because otherwise the `if prevBlockHash == "" || !pr.IsFinalized() {` will fail for `round:1`.

code/go/0chain.net/chaincore/chain/worker.go
dabasov: ah you mean that we should have already processed this round before, just when when were processing frchain here https://github.com/0chain/0chain/blob/81ba565e6e5e6597e94c08bbe68c9093d771ec48/code/go/0chain.net/chaincore/chain/protocol_round.go#L325

code/go/0chain.net/sharder/chain.go
dabasov: ok, got it

code/go/0chain.net/chaincore/chain/worker.go
peterlimg: yes

service-0chain:
| config  | value |
| tests skipped  | FALSE  |
| system-tests  | master  |
| gosdk  | NOT OVERRIDDEN  |
| zboxcli  | staging  |
| zwalletcli  | staging  |
| 0chain  | fix/fb  |
| blobber  | staging  |
| authorizer  | staging  |
| 0box  | staging  |
| 0block  | staging  |
| 0dns  | staging  |
| explorer  | master  |
| 0proxy  | staging  |
| 0search  | staging  |

dabasov (approved):
lgtm

PR #1962, opened 2023-01-03
Added updater to the project

dabasov:
## Changes
added sqlite and query logging for tests

## Tests

- blobber:
- gosdk:
- system_test:
- zboxcli:
- zwalletcli:
- Other: ...


code/go/0chain.net/smartcontract/dbs/event/miner_test.go
```diff
@@ -112,20 +114,206 @@ func TestMiners(t *testing.T) {
...
+		t.Error(err)
+	}
+
+	miner, err := eventDb.GetMiner(mnMiner1.ID)
+	if err != nil {
+		t.Error(err)
+	}
+	i, _ := miner.TotalStake.Int64()
+	assert.Equal(t, int64(10), i)
+
```
peterlimg, line 222: Please also check mnMiner2

code/go/0chain.net/smartcontract/dbs/event/updater.go
```diff
@@ -0,0 +1,99 @@
...
+
+	return b
+}
+
+type Query struct {
+	Q string
+	V []interface{}
+}
+
+func (b *UpdateBuilder) Build() *Query {
```
peterlimg: We can make this `Build()` method private if it would not be used outside, so as the `Query` struct, seems they would only be used in the `Exec()` method.

peterlimg (changes requested):
Please write unit tests for the `updater.go` to make sure that it works as expected and all the edge cases are covered. It would also provide guidance to users on how to use it.

code/go/0chain.net/smartcontract/dbs/event/updater.go
```diff
@@ -0,0 +1,99 @@
...
+		b.sets = append(b.sets, ", ")
+	}
+	switch len(expr) {
+	case 0:
+		b.sets = append(b.sets, fmt.Sprintf(SetTemplate, column, column))
+	case 1:
+		b.sets = append(b.sets, fmt.Sprintf(ExprTemplate, column, expr[0]))
+	default:
+		logging.Logger.Warn("only one expr is supported, ignoring")
+		b.sets = append(b.sets, fmt.Sprintf(ExprTemplate, column, expr[0]))
```
peterlimg, line 72: If only one expr is supported, we should return error instead of continue as user may think the second expr will work

code/go/0chain.net/smartcontract/dbs/event/updater.go
```diff
@@ -0,0 +1,99 @@
...
+	}
+
+	atype, ok := typeToSQL[reflect.TypeOf(values)]
+
+	logging.Logger.Debug("type", zap.String("t", reflect.TypeOf(values).String()))
```
peterlimg: Let's not print logs here, tool package or function should only report error if any, and log the error on the upper caller.

code/go/0chain.net/smartcontract/dbs/event/updater.go
```diff
@@ -0,0 +1,99 @@
...
+	if !ok {
+		atype = typeToSQL[reflect.TypeOf([]string{})]
```
peterlimg, line 78: Why use type of `[]string{}` for none exist type, it may cause unexpected error. Let's return error to users or collect them in an `errs` variable and check it explicitly.

code/go/0chain.net/smartcontract/dbs/event/updater.go
```diff
@@ -0,0 +1,99 @@
...
+	}
+	unnests := ""
+	for _, u := range b.unnests {
+		unnests = unnests + u
+	}
+
+	return &Query{Q: fmt.Sprintf(QueryTemplate, b.update, sets, unnests, b.where), V: b.values}
+}
+
+func (b *UpdateBuilder) Exec(db *EventDb) *gorm.DB {
```
peterlimg, line 110: Perhaps pass an interface to make this easier to test. For example,
```go
type EventDBer interface{
     Exec(query string, values ...) error
}
```

Then we can easily mock it and check whether the sql meet the requirement. We can also connect to real DB if needed

peterlimg (changes requested):

peterlimg:
Please add unit tests for updater.go, I did a simple test and find an issue, please try run the tests below:

```go
func TestBuild(t *testing.T) {
	tests := []struct {
		name           string
		builder        *UpdateBuilder
		expectedQuery  string
		expectedValues []interface{}
	}{
		{
			name: "test with valid input",
			builder: &UpdateBuilder{
				sets:    []string{"col1 = t.col1", "col2 = t.col2"},
				unnests: []string{"unnest(?::int[]) AS col1", "unnest(?::text[]) AS col2"},
				update:  "UPDATE table SET",
				where:   "WHERE table.id = t.id",
				values:  []interface{}{[]interface{}{[]int{1, 2, 3}}, []interface{}{[]string{"val1", "val2", "val3"}}},
			},
			expectedQuery:  "UPDATE table SET col1 = t.col1, col2 = t.col2 FROM (SELECT unnest(?::int[]) AS col1, unnest(?::text[]) AS col2) AS t WHERE table.id = t.id",
			expectedValues: []interface{}{[]interface{}{[]int{1, 2, 3}}, []interface{}{[]string{"val1", "val2", "val3"}}},
		},
	}
	for _, test := range tests {
		t.Run(test.name, func(t *testing.T) {
			query := test.builder.build()
			if query.Q != test.expectedQuery {
				t.Errorf("expected query %s, but got %s", test.expectedQuery, query.Q)
			}
			if !reflect.DeepEqual(query.V, test.expectedValues) {
				t.Errorf("expected values %v, but got %v", test.expectedValues, query.V)
			}
		})
	}
}
```
We will see error:
`expected query UPDATE table SET col1 = t.col1, col2 = t.col2 FROM (SELECT unnest(?::int[]) AS col1, unnest(?::text[]) AS col2) AS t WHERE table.id = t.id, but got UPDATE table SET col1 = t.col1col2 = t.col2 FROM (SELECT unnest(?::int[]) AS col1unnest(?::text[]) AS col2) AS t WHERE table.id = t.id
`

peterlimg:
We should also validate the inputs such as empty table name or columns so that we can catch error before the sql hit db.

code/go/0chain.net/smartcontract/dbs/sqlite/sqlite.go
```diff
@@ -0,0 +1,57 @@
...
+		return nil, err
+	}
+	return db, nil
+}
+
+type SqliteStore struct {
+	db *gorm.DB
+}
+
+func (store *SqliteStore) Open(config config.DbAccess) error {
```
peterlimg: The `config` param passed in is not used

peterlimg (changes requested):

peterlimg:
Another issue I found when using the `updater` to update the blobber healthcheck is that, it will fail with error of `	ERROR: column reference "downtime" is ambiguous (SQLSTATE 42702)`. Please setup a postgres localy and run the test code below:
```go
func TestEventDb_updateProvidersHealthCheck(t *testing.T) {
	//t.Skip("only for local debugging, requires local postgresql")
	edb, closeDB := prepareEventDB(t)
	defer closeDB()

	//db, err := sqlite.GetSqliteDb()
	//require.NoError(t, err)
	//
	//edb := &EventDb{
	//	Store: db,
	//}

	config.Configuration().ChainConfig = &mockConfig{
		dbSettings: config.DbSettings{
			Debug:           false,
			AggregatePeriod: 10,
			PageLimit:       10,
		},
	}
	err := edb.AutoMigrate()
	require.NoError(t, err)

	err = edb.addBlobbers([]Blobber{
		{
			Provider: Provider{ID: "b1"},
			BaseURL:  "b1.com",
		}, {
			Provider: Provider{ID: "b2"},
			BaseURL:  "b2.com",
		},
	})
	require.NoError(t, err)

	bs, err := edb.GetBlobbers(common2.Pagination{Limit: 10})
	require.NoError(t, err)
	require.Len(t, bs, 2)
	//v, err := json.MarshalIndent(bs, " ", "")
	//require.NoError(t, err)
	//fmt.Println(string(v))

	up := dbs.DbHealthCheck{
		ID:              "b1",
		LastHealthCheck: common.Now(),
		Downtime:        uint64(10 * time.Second),
	}

	err = edb.updateProvidersHealthCheck([]dbs.DbHealthCheck{up}, BlobberTable)
	require.NoError(t, err)
}
```

Btw, the sqlite will give different error message, so not sure if sqlite could be used for testing.

dabasov:
> Please add unit tests for updater.go, I did a simple test and find an issue, please try run the tests below:
> 
> ```go
> func TestBuild(t *testing.T) {
> 	tests := []struct {
> 		name           string
> 		builder        *UpdateBuilder
> 		expectedQuery  string
> 		expectedValues []interface{}
> 	}{
> 		{
> 			name: "test with valid input",
> 			builder: &UpdateBuilder{
> 				sets:    []string{"col1 = t.col1", "col2 = t.col2"},
> 				unnests: []string{"unnest(?::int[]) AS col1", "unnest(?::text[]) AS col2"},
> 				update:  "UPDATE table SET",
> 				where:   "WHERE table.id = t.id",
> 				values:  []interface{}{[]interface{}{[]int{1, 2, 3}}, []interface{}{[]string{"val1", "val2", "val3"}}},
> 			},
> 			expectedQuery:  "UPDATE table SET col1 = t.col1, col2 = t.col2 FROM (SELECT unnest(?::int[]) AS col1, unnest(?::text[]) AS col2) AS t WHERE table.id = t.id",
> 			expectedValues: []interface{}{[]interface{}{[]int{1, 2, 3}}, []interface{}{[]string{"val1", "val2", "val3"}}},
> 		},
> 	}
> 	for _, test := range tests {
> 		t.Run(test.name, func(t *testing.T) {
> 			query := test.builder.build()
> 			if query.Q != test.expectedQuery {
> 				t.Errorf("expected query %s, but got %s", test.expectedQuery, query.Q)
> 			}
> 			if !reflect.DeepEqual(query.V, test.expectedValues) {
> 				t.Errorf("expected values %v, but got %v", test.expectedValues, query.V)
> 			}
> 		})
> 	}
> }
> ```
> 
> We will see error: `expected query UPDATE table SET col1 = t.col1, col2 = t.col2 FROM (SELECT unnest(?::int[]) AS col1, unnest(?::text[]) AS col2) AS t WHERE table.id = t.id, but got UPDATE table SET col1 = t.col1col2 = t.col2 FROM (SELECT unnest(?::int[]) AS col1unnest(?::text[]) AS col2) AS t WHERE table.id = t.id `

you can't put inputs directly, please use builder interface for he test. in your particular test unnest set is not in correct, check code for 
```go
func (b *UpdateBuilder) AddUpdate(column string, values interface{}, expr ...string) *UpdateBuilder {
	if b.sets != nil {
		b.sets = append(b.sets, ", ")
	}
	switch len(expr) {
	case 0:
		b.sets = append(b.sets, fmt.Sprintf(SetTemplate, column, column))
	case 1:
		b.sets = append(b.sets, fmt.Sprintf(ExprTemplate, column, expr[0]))
	default:
		logging.Logger.Warn("only one expr is supported, ignoring")
		b.sets = append(b.sets, fmt.Sprintf(ExprTemplate, column, expr[0]))
	}

	atype, ok := typeToSQL[reflect.TypeOf(values)]

	if !ok {
		atype = typeToSQL[reflect.TypeOf([]string{})]
	}

	if b.unnests != nil {
		b.unnests = append(b.unnests, ", ")
	}
	b.unnests = append(b.unnests, fmt.Sprintf(UnnestTemplate, atype, column))

	b.values = append(b.values, []interface{}{pq.Array(values)})

	return b
}

```


you add ", " to an end of unnest

dabasov:
> Another issue I found when using the `updater` to update the blobber healthcheck is that, it will fail with error of ` ERROR: column reference "downtime" is ambiguous (SQLSTATE 42702)`. Please setup a postgres localy and run the test code below:
> 
> ```go
> func TestEventDb_updateProvidersHealthCheck(t *testing.T) {
> 	//t.Skip("only for local debugging, requires local postgresql")
> 	edb, closeDB := prepareEventDB(t)
> 	defer closeDB()
> 
> 	//db, err := sqlite.GetSqliteDb()
> 	//require.NoError(t, err)
> 	//
> 	//edb := &EventDb{
> 	//	Store: db,
> 	//}
> 
> 	config.Configuration().ChainConfig = &mockConfig{
> 		dbSettings: config.DbSettings{
> 			Debug:           false,
> 			AggregatePeriod: 10,
> 			PageLimit:       10,
> 		},
> 	}
> 	err := edb.AutoMigrate()
> 	require.NoError(t, err)
> 
> 	err = edb.addBlobbers([]Blobber{
> 		{
> 			Provider: Provider{ID: "b1"},
> 			BaseURL:  "b1.com",
> 		}, {
> 			Provider: Provider{ID: "b2"},
> 			BaseURL:  "b2.com",
> 		},
> 	})
> 	require.NoError(t, err)
> 
> 	bs, err := edb.GetBlobbers(common2.Pagination{Limit: 10})
> 	require.NoError(t, err)
> 	require.Len(t, bs, 2)
> 	//v, err := json.MarshalIndent(bs, " ", "")
> 	//require.NoError(t, err)
> 	//fmt.Println(string(v))
> 
> 	up := dbs.DbHealthCheck{
> 		ID:              "b1",
> 		LastHealthCheck: common.Now(),
> 		Downtime:        uint64(10 * time.Second),
> 	}
> 
> 	err = edb.updateProvidersHealthCheck([]dbs.DbHealthCheck{up}, BlobberTable)
> 	require.NoError(t, err)
> }
> ```
> 
> Btw, the sqlite will give different error message, so not sure if sqlite could be used for testing.

we do not use sqlite for this test, as you see it is commented out

dabasov:
> We should also validate the inputs such as empty table name or columns so that we can catch error before the sql hit db.

I don't think we should validate inputs, most of the time you should know the contract of interface you use, it's up to developer to use it correctly, it's not clients input, so we can not write it defensively all the time. Imagine code where every func checks and validates its inputs, code will be ambiguous, all the time, so most of the time you should prepare data properly to use interface

code/go/0chain.net/smartcontract/dbs/event/updater.go
dabasov: no it's fine, only one will be used and we will ignore others, I don't see any problems here

code/go/0chain.net/smartcontract/dbs/event/updater.go
dabasov: string is a default type, I checked it works most of the time, since array is a json and types are converted to strings internally the same way

dabasov:
@peterlimg please check your setup, most likely you see errors locally, because it is broken somehow, this DEV network is deployed by this branch and is working fine. also system tests are passing

code/go/0chain.net/smartcontract/dbs/event/provider.go
```diff
@@ -68,15 +57,52 @@ func (p *Provider) BeforeCreate(tx *gorm.DB) (err error) {
 	return
 }
 
+func (edb *EventDb) updateProviderTotalStakes(providers []Provider, tablename string) error {
+	var ids []string
+	var stakes []int64
+	for _, m := range providers {
+		ids = append(ids, m.ID)
+		i, err := m.TotalStake.Int64()
```
peterlimg, line 68: We don't need to do the int64 converting if the updater support uint64, and it should be easy to support.

code/go/0chain.net/smartcontract/dbs/event/provider.go
```diff
@@ -68,15 +57,52 @@ func (p *Provider) BeforeCreate(tx *gorm.DB) (err error) {
...
+	return CreateBuilder(tablename, "id", ids).
+		AddUpdate("total_stake", stakes).Exec(edb).Error
+}
+
+func (edb *EventDb) updateProvidersTotalUnStakes(providers []Provider, tablename string) error {
+	var ids []string
+	var unstakes []int64
+	for _, m := range providers {
+		ids = append(ids, m.ID)
+		i, err := m.TotalStake.Int64()
```
peterlimg: It should be `m.UnstakeTotal`, we need tests to ensure the total unstakes is updated correctly

code/go/0chain.net/smartcontract/dbs/event/config.go
```diff
@@ -0,0 +1,258 @@
...
+
+import (
+	"time"
+
+	"0chain.net/chaincore/config"
+	"0chain.net/core/datastore"
+	"github.com/0chain/common/core/currency"
+)
+
+type TestConfig struct {
```
peterlimg, line 11: Test structs need to be put in `_test.go` file.

code/go/0chain.net/smartcontract/dbs/event/updater.go
```diff
@@ -0,0 +1,112 @@
...
+const SetTemplate = "%v = t.%v"
+const ExprTemplate = "%v = %v"
+const UnnestTemplate = "unnest(?::%v[]) AS %v"
+const UpdateTemplate = "UPDATE %v SET"
+const WhereTemplate = "WHERE %v.%v = t.%v"
+const QueryTemplate = "%v %v FROM (SELECT %v) AS t %v"
+
+var typeToSQL = map[reflect.Type]string{
+	reflect.TypeOf([]string{}):  "text",
+	reflect.TypeOf([]int64{}):   "bigint",
```
peterlimg, line 21: Add `[]uint64` would make it easier for 'Stake' value updating and also for those uin64 field such as the 'Downtime' for health check.

peterlimg (changes requested):

code/go/0chain.net/smartcontract/dbs/event/updater.go
dabasov: we will add it in the next issue, we need to merge it asap

code/go/0chain.net/smartcontract/dbs/event/provider.go
dabasov: we will add it in the next issue, we need to merge it asap

code/go/0chain.net/smartcontract/dbs/event/config.go
dabasov: it's not test struct, it is utility struct for testing

peterlimg (approved):
LGTM

PR #1975, opened 2023-01-07
Added ZCN Service Fee

### LINKED ISSUE #1922: ZCN Service Fee
This fee should be minted and distributed between authorizers that took part in proof creation. Mint amount should be split into actual amount to mint and service fee ``amount = mint + service_fee``

There is MaxFee value in configs, but it's not used anyhow
```go
	conf.MaxFee = currency.Coin(cfg.GetInt64(postfix(MaxFee)))
```

din-mukhammed:
## Tests

- blobber:
- gosdk:
- system_test:
- zboxcli:
- zwalletcli:
- Other: ...


code/go/0chain.net/smartcontract/zcnsc/mint.go
```diff
@@ -114,11 +113,43 @@ func (zcn *ZCNSmartContract) Mint(trans *transaction.Transaction, inputData []by
...
+	if err != nil {
+		err = errors.Wrap(err, fmt.Sprintf("%s, convert len signatures to coin, %s", code, info))
+		return
+	}
+	amount, err = currency.MinusCoin(payload.Amount, share*n)
+	if err != nil {
+		err = errors.Wrap(err, fmt.Sprintf("%s, payload.Amount - share * len(signatures), %s", code, info))
+		return
+	}
+	for _, sig := range payload.Signatures {
```
dabasov, line 137: checked how we process transfers [here](https://github.com/0chain/0chain/blob/e28bff708acbc89a5cee677ed082aa184a87756e/code/go/0chain.net/chaincore/chain/state.go#L420:L463)
and looks like if client has no balance transfer won't be possible with ``transfer amount - insufficient balance`` error. 
I think that we simply should mint share also, the same way we mint amount and we are fine

dabasov (changes requested):

dabasov (approved):
lgtm

code/go/0chain.net/smartcontract/zcnsc/mint.go
```diff
@@ -114,11 +113,43 @@ func (zcn *ZCNSmartContract) Mint(trans *transaction.Transaction, inputData []by
...
+	if err != nil {
+		err = errors.Wrap(err, fmt.Sprintf("%s, DistributeCoin operation, %s", code, info))
+		return
+	}
+	n, err = currency.Int64ToCoin(int64(len(payload.Signatures)))
+	if err != nil {
+		err = errors.Wrap(err, fmt.Sprintf("%s, convert len signatures to coin, %s", code, info))
+		return
+	}
+	amount, err = currency.MinusCoin(payload.Amount, share*n)
```
peterlimg, line 131: The `share*n` is the `gn.ZCNSconfig.MaxFee`, we don't need to recalculate it. 

So are we going to send `MaxFee` as service to all the authorizers?, if so we'd better rename the `MaxFee` to `AuthorizerServiceFee` or something like to avoid ambiguity.

code/go/0chain.net/smartcontract/zcnsc/mint.go
```diff
@@ -114,11 +113,43 @@ func (zcn *ZCNSmartContract) Mint(trans *transaction.Transaction, inputData []by
...
+		amount currency.Coin
+		n      currency.Coin
+		share  currency.Coin
+	)
+	share, _, err = currency.DistributeCoin(gn.ZCNSConfig.MaxFee, int64(len(payload.Signatures)))
+	if err != nil {
+		err = errors.Wrap(err, fmt.Sprintf("%s, DistributeCoin operation, %s", code, info))
+		return
+	}
+	n, err = currency.Int64ToCoin(int64(len(payload.Signatures)))
```
peterlimg, line 126: Unnecessary converting as we don't need the `n`

peterlimg (changes requested):
Please add unit tests to verify the service fee paying. So need to assert:
- service fee is paid to all authorizers correctly
- the mint tokens is correct

code/go/0chain.net/smartcontract/zcnsc/mint.go
dabasov: we can calculate product in uint64 and then convert it to coin, will it be any different?

code/go/0chain.net/smartcontract/zcnsc/mint.go
dabasov: `share*n` can be not equal to `n.ZCNSconfig.MaxFee`, imagine you can't divide maxfee evenly for all authorizers, so current calculation is correct

peterlimg: `share*n` can be different from `z.ZCNSconfig.MaxFee`, please check why there is a `share*n` here.

peterlimg: Yes, the `share*n` can be different from the `n.ZCNSconfig.MaxFee`, so we don't care about the remaining balance? Suppose the `MaxFee` is `10`, and we the `n=length(sigs)` is 3, then we will get share as `3`, and `n*share` would be `9`, then we just send 9 as the service fee?

peterlimg: I can kind of understand why we name it as `MaxFee` here, as we may send service fee less than it if the `MaxFee` can not be dividend evenly by the authrorizers or validators. If this is the design, I agree that the calculation is correct.

peterlimg:
@din-mukhammed please add the unit tests

peterlimg (approved):

din-mukhammed:
@peterlimg @dabasov added ut

code/go/0chain.net/smartcontract/zcnsc/mint.go
dabasov: @peterlimg we will leave the remaining for client, you can never divide Fee evenly, because you can send different number of signatures, not every time 2f+, it can be more sometimes, should't be an issue

dabasov (approved):
lgtm

code/go/0chain.net/smartcontract/zcnsc/mint_test.go
```diff
@@ -66,6 +67,39 @@ func Test_FuzzyMintTest(t *testing.T) {
...
+		require.Equal(t, mm[i].ToClientID, authorizersID[i])
+		require.Equal(t, mm[i].Amount, expectedShare)
+	}
+
+	rp := &MintPayload{}
+	require.NoError(t, rp.Decode([]byte(response)))
+	require.Equal(t, payload.ReceivingClientID, rp.ReceivingClientID)
+	expectedAmount, err := currency.MinusCoin(payload.Amount, maxFee)
+	require.NoError(t, err)
+	require.Equal(t, rp.Amount, expectedAmount)
```
peterlimg, line 104: I think this would fail if the MaxFee is 10, can you add another case for when it's 10, and the rp should be the amount-9.

din-mukhammed: Yes, thanks. expectedAmount should be calculated:
`expectedAmount = payload.Amount - share * len(signatures)`
Fixed in latest

PR #1980, opened 2023-01-09
updated dependency versions

dabasov:
upgraded lru cache

## Tests

- blobber:
- gosdk:
- system_test:
- zboxcli:
- zwalletcli:
- Other: ...


code/go/0chain.net/sharder/m_handler.go
```diff
@@ -153,7 +153,7 @@ func (sc *Chain) cacheProcessingBlock(hash string) bool {
 	_, err := sc.processingBlocks.Get(hash)
 	switch err {
 	case cache.ErrKeyNotFound:
-		if err := sc.processingBlocks.Add(hash, struct{}{}); err != nil {
+		if err := sc.processingBlocks.Add(hash, &block.Block{}); err != nil {
```
peterlimg: Let's keep using the`struct{}{}`, and define the `processingBlocks` as `cache.LRU[string, struct{}]`.

peterlimg (changes requested):

code/go/0chain.net/sharder/m_handler.go
dabasov: @peterlimg why we should do it, if they are blocks?

peterlimg: It's the same as why do `map[sting]stuct{}` instead of `map[string]*block.Block`. We actually don't care the about the value, we just use key to check duplication, for those case, we should also consider empty struct. For why use empty struct, please check https://dave.cheney.net/2014/03/25/the-empty-struct

peterlimg (approved):
LGTM

service-0chain:
| config  | value |
| tests skipped  | TRUE  |
| system-tests  | master  |
| gosdk  | NOT OVERRIDDEN  |
| zboxcli  | staging  |
| zwalletcli  | staging  |
| 0chain  | enh/lru  |
| blobber  | staging  |
| authorizer  | staging  |
| 0box  | staging  |
| 0block  | staging  |
| 0dns  | staging  |
| explorer  | master  |
| 0proxy  | staging  |
| 0search  | staging  |

PR #2025, opened 2023-01-17
Fix OpenChallenges become negative after some time

### LINKED ISSUE #1945: OpenChallenges are negative
OpenChallenges can't be negative, most likely update is broken
```
{"code":"internal_error","error":"internal_error: cannot get blobber listsql: Scan error on column index 31, name \"open_challenges\": converting driver.Value type int64 (\"-226\") to a uint64: invalid syntax"}
```
related commit https://github.com/0chain/0chain/commit/4db016c625600e1188c69a3660a325210e8a585e

> The 1-expiredN means add one new open challenge and remove expiredN open challenges, it’s not the final result, and should be added to the current open challenge number, but the TagUpdateBlobberOpenChallenges process code just simply set the total open challenge as the 1-expiredN , which is wrong.

related discussion
https://0chain.slack.com/archives/C03788G8N2D/p1671805735260109

devyetii:
## Fixes
- blobbers.open_challenges become negative after some time. The problem is [here](https://github.com/0chain/0chain/blob/2c5acc7e172835bf84e3f5b255a29357e5b56a95/code/go/0chain.net/smartcontract/storagesc/challenge.go#L1110) the count of expired challenges is calculated for the allocation of the challenge regardless of which blobber this challenge is tied to, then sent to the sharder to be subtracted from open_challenges of a single blobber (the one the challenge is tied to i.e. challenge.BlobberID). Thus, if we have say 6 expired challenges for the allocation, 3 for B1 and 3 for B2, B1 has total of 4 challenges, the event is sent with `openDelta = -5 (1 challenge to add and 6 expired ones to remove)` so it brings B2's open_challenges down to -1.
- Also the same function produce duplicated expiredIDs, so will remove the duplicates from the count.
## Changes
- Regard the blobber in counting expired challenges
- Subtract the count of the duplicates

## Tests

- blobber:
- gosdk:
- system_test:
- zboxcli:
- zwalletcli:
- Other: ...


devyetii:
@peterlimg @dabasov Kindly review

peterlimg (changes requested):
@devyetii good findings, the reason for the negative value is correct, but we need to tune a bit for the fixing. As you have found that the `expiredIDs` belongs to the allocation rather than a single blobber, so what we need to do in the `addChallenge` method is to emit the open challenge updates for all the related blobbers. To achieve this, the `expiredIDs ` returned from the `removeExpiredChallenge` itself would not be sufficient, instead we need to return a map of the expired challenge ID and blobber ID, or a simple struct with those two data. then in the `addChallenge` method we emit the updates for all blobbers. Update for single blobber would make the open challenges stat inconsistent with the allocation.

code/go/0chain.net/smartcontract/storagesc/challenge.go
```diff
@@ -1116,7 +1117,11 @@ func (sc *StorageSmartContract) addChallenge(alloc *StorageAllocation,
 	for _, id := range expiredIDs {
 		_, err := balances.DeleteTrieNode(storageChallengeKey(sc.ID, id))
 		if err != nil {
-			return common.NewErrorf("add_challenge", "could not delete challenge node: %v", err)
+			if err == util.ErrValueNotPresent {
+				lenExpiredIDs--
```
peterlimg: We don't need to do this check, we should not have ErrValueNotPresent otherwise we need to return error instead of continue.

code/go/0chain.net/smartcontract/storagesc/models.go
```diff
@@ -1346,13 +1346,14 @@ func getMaxChallengeCompletionTime() time.Duration {
...
 func (sa *StorageAllocation) removeExpiredChallenges(allocChallenges *AllocationChallenges,
-	now common.Timestamp) ([]string, error) {
+	now common.Timestamp, blobberID string) ([]string, error) {
 	var (
 		expiredChallengeIDs = make([]string, 0, len(allocChallenges.OpenChallenges))
 	)
 
 	cct := getMaxChallengeCompletionTime()
 	for _, oc := range allocChallenges.OpenChallenges {
+		if oc.BlobberID != blobberID { continue }
```
peterlimg: We need to remove all expired open challenges, not just the one connect to the blobber

peterlimg (changes requested):

code/go/0chain.net/smartcontract/storagesc/challenge.go
devyetii: When this happen it blocks challenge generation where expiredIDs contain some duplicates (still trying to figure out why this happens).

code/go/0chain.net/smartcontract/storagesc/challenge.go
```diff
@@ -1107,20 +1107,28 @@ func (sc *StorageSmartContract) addChallenge(alloc *StorageAllocation,
 	}
 
 	// remove expired challenges
-	expiredIDs, err := alloc.removeExpiredChallenges(allocChallenges, challenge.Created, challenge.BlobberID)
-	lenExpiredIDs   := len(expiredIDs)
+	expiredIDsMap, totalExpiredCount, err := alloc.removeExpiredChallenges(allocChallenges, challenge.Created, challenge.BlobberID)
```
peterlimg: Why not just return the map of open challenge id and blobber id,  i.e. `map[string]string` instead of using blobber id as key. The process will be simplified, we can also check the duplication if needed when collecting the expired challenges. And we wouldn't need the second `totalExpiredCount` returned value.

code/go/0chain.net/smartcontract/storagesc/challenge.go
```diff
@@ -1107,20 +1107,28 @@ func (sc *StorageSmartContract) addChallenge(alloc *StorageAllocation,
...
+		if _, ok := expiredCountMap[bid]; !ok {
+			expiredCountMap[bid] = len(IDsPerBlobber)
+		}
+		for _, id := range IDsPerBlobber {
 		_, err := balances.DeleteTrieNode(storageChallengeKey(sc.ID, id))
 		if err != nil {
 			if err == util.ErrValueNotPresent {
-				lenExpiredIDs--
+					expiredCountMap[bid]--
+					totalExpiredCount--
```
peterlimg: Do not change the expired number as they should be what they are when collected, otherwise the stats will be inconsisted with the stat in allocation and each blobber in MPT. 

If we think there could be duplicate open challenges that could cause the second time of node deleting, then we need to add logs to check and debug to find out why there are duplicate items, not change it here as this will hide the bug if it exists. We can fix it if any in another thread, the easy way to debug is log the open challenge ID when removing it, and log the ValueNotPresent error, grafana will persist logs for us so that when the error happens, we can see when was the first time the challenge get deleted.

peterlimg (changes requested):

code/go/0chain.net/smartcontract/storagesc/challenge.go
devyetii: Actually I was trying to figure out the cause of this duplication but since the negative open challenges problem is now solved, I think we can address this issue in another PR. I find the logging solution you suggested pretty enough for now until the duplication problem is actually solved.

devyetii:
@peterlimg I shifted duplicate checking to `removeExpiredChallenges` so that we don't have data inconsistency in stats in case of duplicates. Kindly review.

peterlimg (approved):
LGTM, @devyetii please add unit tests next time to ensure that the changes work as expected. I added the tests so that we can close this PR fast.

service-0chain:
| config  | value |
| tests skipped  | TRUE  |
| system-tests  | master  |
| gosdk  | NOT OVERRIDDEN  |
| zboxcli  | staging  |
| zwalletcli  | staging  |
| 0chain  | 1945-openchallenges-are-negative  |
| blobber  | staging  |
| authorizer  | staging  |
| 0box  | staging  |
| 0block  | staging  |
| 0dns  | staging  |
| explorer  | master  |
| 0proxy  | staging  |
| 0search  | staging  |

PR #2036, opened 2023-01-20
Fixes issue where Mint returns error "not enough valid signatures" in wrong scenario

areebusmani:
## Fixes
1. Return appropriate error when there is an error in finding number of authorisers.
2. Return from Mint function once there is error returned from verifySignatures function

## Changes
1. Add unit test case for covering the scenario where number of authorisers are 0 but signatures are non empty.

## Tests

- blobber:
- gosdk:
- system_test:
- zboxcli:
- zwalletcli:
- Other: ...


code/go/0chain.net/smartcontract/zcnsc/mint_test.go
```diff
@@ -81,6 +81,22 @@ func Test_EmptySignaturesShouldFail(t *testing.T) {
 	require.Error(t, err)
 }
 
+func Test_EmptyAuthorizersShouldFail(t *testing.T) {
```
dabasov: add test where number of signatures are not empty

dabasov (dismissed):
lgtm

code/go/0chain.net/smartcontract/zcnsc/mint_test.go
areebusmani: Done

code/go/0chain.net/smartcontract/zcnsc/mint.go
```diff
@@ -49,6 +49,12 @@ func (zcn *ZCNSmartContract) Mint(trans *transaction.Transaction, inputData []by
 	}
 
 	numAuth, err := getAuthorizerCount(ctx)
+	if err != nil {
```
peterlimg, line 51: The err is nil even the numAuth does not exist. So we should also check if numAuth is 0.

code/go/0chain.net/smartcontract/zcnsc/mint_test.go
```diff
@@ -81,6 +81,45 @@ func Test_EmptySignaturesShouldFail(t *testing.T) {
...
+	ctx.authorizers = nil
+
+	contract := CreateZCNSmartContract()
+	payload, err := CreateMintPayload(ctx, defaultClient)
+	require.NoError(t, err)
+
+	transaction, err := CreateTransaction(defaultClient, "mint", payload.Encode(), ctx)
+	require.NoError(t, err)
+
+	_, err = contract.Mint(transaction, payload.Encode(), ctx)
```
peterlimg, line 153: The err is payload decoding error, please check the err message. If this is an intended test,  then we need to rename the Test to `Test_EmptySignaturehouldFail`.

code/go/0chain.net/smartcontract/zcnsc/mint_test.go
```diff
@@ -81,6 +81,45 @@ func Test_EmptySignaturesShouldFail(t *testing.T) {
...
+	var signatures []*AuthorizerSignature
+	for i := 0; i < 10; i++ {
+		signatures = append(signatures, &AuthorizerSignature{})
+	}
+	payload.Signatures = signatures
+
+	transaction, err := CreateTransaction(defaultClient, "mint", payload.Encode(), ctx)
+	require.NoError(t, err)
+
+	_, err = contract.Mint(transaction, payload.Encode(), ctx)
```
peterlimg, line 175: The `numAuth` is not empty in the `Mint` contract, the error is returned when verifying signatures, please check the place where the error is returned and make sure the tests do test the code we want.

peterlimg (changes requested):

code/go/0chain.net/smartcontract/zcnsc/mint.go
dabasov: I don't think we should check it, since we should invalidate any number of signatures in case we have 0 authorizers. In general, it should not differ how much authorizers we do ave 0, 1 or k, it should work the same way

peterlimg: > since we should invalidate any number of signatures. in case we have 0 authorizers. 

Yes, that's what I mean, `numAuth` is the number of authorizers, and we should stop the Mint if it is 0, otherwise the zero division panic would happen as we have encountered.

code/go/0chain.net/smartcontract/zcnsc/mint_test.go
areebusmani: Removed this test cases as the scenario where number of authorizers are 0 is already covered in another test (Test_EmptyAuthorizersNonemptySignaturesShouldFail) that I have added in this PR.

code/go/0chain.net/smartcontract/zcnsc/mint_test.go
areebusmani: Fixed

areebusmani: The error will still be "signatures not found" because since numAuth is zero so signatures will be sliced to an empty array.

https://github.com/0chain/0chain/blob/staging/code/go/0chain.net/smartcontract/zcnsc/mint.go#L62

code/go/0chain.net/smartcontract/zcnsc/mint.go
areebusmani: Actually, the panic does not happen in the unit test case that I have added in this PR. 

Since number of authorizers is 0 so the signatures are sliced to an empty array. 
https://github.com/0chain/0chain/blob/staging/code/go/0chain.net/smartcontract/zcnsc/mint.go#L62

Then, while trying to verify the signatures the method is returning an error saying that there are no signatures.

https://github.com/0chain/0chain/blob/staging/code/go/0chain.net/smartcontract/zcnsc/mint.go#L99
https://github.com/0chain/0chain/blob/31c960fcf9d2ebc1e160e8810b28927132f2ce90/code/go/0chain.net/smartcontract/zcnsc/models.go#L154

Hence, we don't get to the point where we divide the share to all authorizers.

Previously, we might have been getting panic because of the missing return statement here.
https://github.com/0chain/0chain/blob/staging/code/go/0chain.net/smartcontract/zcnsc/mint.go#L102

code/go/0chain.net/smartcontract/zcnsc/context_helpers_test.go
```diff
@@ -239,6 +239,192 @@ func MakeMockStateContext() *mockStateContext {
 	return ctx
 }
 
+func MakeMockStateContextWithNoAutorizers() *mockStateContext {
```
areebusmaniblr, line 242: @peterlimg Since golang neither supports optional parameters nor method overloading so I was finding it difficult to add modify MakeMockStateContext function to conditional add or not add the authorizers. So I copied the method to a new method MakeMockStateContextWithNoAutorizers and removed the part responsible for adding autorizers.

Can you suggest a better way of achieving the same thing?

code/go/0chain.net/smartcontract/zcnsc/mint.go
peterlimg: @areebusmaniblr yes, that should work to fix the panic (thought this branch would never panic as it does not have the latest code from staging), but we'd better check the 0 number of authorizer explicitly instead of letting the empty sliced signatures' verification code to do the 0 checking, in any chance if the `verifySignature` does not do the empty checking, we will panic here https://github.com/0chain/0chain/blob/59c0b3a1481d8e3c776c612987f494270d256e61/code/go/0chain.net/smartcontract/zcnsc/mint.go#L121. Anyway, I'll do a quick commit and merge this PR.

peterlimg (approved):
LGTM

PR #2039, opened 2023-01-22
Fix genesis block re-entry problem

### LINKED ISSUE #1949: Genesis block re-entry problem still exist
Checking the current round does not work as it's always value of 0 when node starts, we will need to check the LFB. So if LFB > 0, then we don't need to generate genesis block, otherwise, we need to check the latest round, current round may not work at the moment, need to check the code to see if we can get something from persistent storage.

din-mukhammed:
## Tests

- blobber:
- gosdk:
- system_test:
- zboxcli:
- zwalletcli:
- Other: ...


code/go/0chain.net/sharder/sharder/sharder.go
```diff
@@ -224,9 +224,9 @@ func main() {
 	go sc.StartLFMBWorker(ctx)
 
 	setupBlockStorageProvider(mConf, workdir)
-	if sc.GetCurrentRound() == 0 {
-		sc.SetupGenesisBlock(viper.GetString("server_chain.genesis_block.id"),
-			magicBlock, initStates)
+	latestFinalizedBlock := sc.GetLatestFinalizedBlock()
+	if latestFinalizedBlock == nil || latestFinalizedBlock.Round == 0 {
+		sc.SetupGenesisBlock(viper.GetString("server_chain.genesis_block.id"), magicBlock, initStates)
```
peterlimg, line 229: LatestFinalizedBlock() might work for sharder(need to confirm), but we also need to check the miners, whose LatestFinalizedBlock() seems always nil when on starting.

peterlimg (changes requested):

code/go/0chain.net/sharder/sharder/sharder.go
din-mukhammed: Added LFB check for miners also

peterlimg: @din-mukhammed LFB is also always 0 on start for miners, it's actually not easy to fix. Anyway, I'll fix it in another PR for fixing chain stuck and close this one.

peterlimg:
Closed as https://github.com/0chain/0chain/pull/2039#discussion_r1092710726

PR #2042, opened 2023-01-23
Manage error handling for new test endpoint

Kishan-Dhakan:
## Changes
- Add provider_type param to differentiate between miner, sharder
- Enhances error logic to avoid returning unreasonable/incorrect errors

## Tests

- blobber:
- gosdk:
- system_test:
- zboxcli:
- zwalletcli:
- Other: ...

### DIFF
--- a/code/go/0chain.net/smartcontract/dbs/event/miner.go
+++ b/code/go/0chain.net/smartcontract/dbs/event/miner.go
@@ -73,7 +73,7 @@ func (edb *EventDb) GetMinerWithDelegatePools(id string) (Miner, []DelegatePool,
 	}
 	m = minerDps[0].Miner
 	if id != minerDps[0].ProviderRewards.ProviderID {
-		return m, nil, fmt.Errorf("mismatched miner; want id %s but have id%s in provider rewrards",
+		return m, nil, fmt.Errorf("mismatched miner; want id %s but have id %s in provider rewrards",
 			id, minerDps[0].ProviderRewards.ProviderID)
 	}
 	m.Rewards = minerDps[0].ProviderRewards
--- a/code/go/0chain.net/smartcontract/dbs/event/miner.go
+++ b/code/go/0chain.net/smartcontract/dbs/event/miner.go
@@ -74,7 +74,7 @@ func (edb *EventDb) GetMinerWithDelegatePools(id string) (Miner, []DelegatePool,
 	m = minerDps[0].Miner
 	if id != minerDps[0].ProviderRewards.ProviderID {
 		return m, nil, fmt.Errorf("mismatched miner; want id %s but have id%s in provider rewrards",
-			id, minerDps[0].Miner.ID)
+			id, minerDps[0].ProviderRewards.ProviderID)
 	}
 	m.Rewards = minerDps[0].ProviderRewards
 	for i := range minerDps {
--- a/code/go/0chain.net/smartcontract/minersc/handler.go
+++ b/code/go/0chain.net/smartcontract/minersc/handler.go
@@ -406,16 +406,16 @@ type nodeStat struct {
 //	484:
 func (mrh *MinerRestHandler) testNodeStat(w http.ResponseWriter, r *http.Request) {
 	var (
-		provider_id      = r.URL.Query().Get("provider_id")
+		providerId      = r.URL.Query().Get("provider_id")
 		includeDelegates = strings.ToLower(r.URL.Query().Get("include_delegates")) == "true"
 	)
-	providerTypeString := r.URL.Query().Get("provider_type")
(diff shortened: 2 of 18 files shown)


code/go/0chain.net/smartcontract/minersc/handler.go
```diff
@@ -406,10 +406,16 @@ type nodeStat struct {
 //	484:
 func (mrh *MinerRestHandler) testNodeStat(w http.ResponseWriter, r *http.Request) {
 	var (
-		id               = r.URL.Query().Get("id")
+		provider_id      = r.URL.Query().Get("provider_id")
 		includeDelegates = strings.ToLower(r.URL.Query().Get("include_delegates")) == "true"
```
peterlimg, line 410: Use camel format, i.e `providerID`

code/go/0chain.net/smartcontract/minersc/handler.go
```diff
@@ -406,10 +406,16 @@ type nodeStat struct {
...
 	)
-	if id == "" {
+	providerTypeString := r.URL.Query().Get("provider_type")
```
peterlimg: Keep the variable name short, could be `providerTypeStr`, or `pTypeStr`

code/go/0chain.net/smartcontract/minersc/handler.go
```diff
@@ -418,34 +424,46 @@ func (mrh *MinerRestHandler) testNodeStat(w http.ResponseWriter, r *http.Request
...
+		} else {
+			miner, err = edb.GetMiner(provider_id)
+		}
+		if err == nil {
+			common.Respond(w, r, nodeStat{
+				NodeResponse: minerTableToMinerNode(miner, dps),
+				TotalReward:  int64(miner.Rewards.TotalRewards),
+			}, nil)
+		} else {
+			common.Respond(w, r, nil, common.NewErrBadRequest(err.Error()))
```
peterlimg, line 440: In Go, we prefer to do err checking first and return error early rather than the `if else` format, so let's do
```go
if err != nil {
    common.Response(w, r, nil, common.NewErrBadRequest(err.Error)))
    return
}

common.Response(...)
return
```

Same for the sharder process below

code/go/0chain.net/smartcontract/dbs/event/miner.go
```diff
@@ -74,7 +74,7 @@ func (edb *EventDb) GetMinerWithDelegatePools(id string) (Miner, []DelegatePool,
 	m = minerDps[0].Miner
 	if id != minerDps[0].ProviderRewards.ProviderID {
 		return m, nil, fmt.Errorf("mismatched miner; want id %s but have id%s in provider rewrards",
```
peterlimg: The second `id%s` should have a space, please also help to fix it.

peterlimg (changes requested):

code/go/0chain.net/smartcontract/minersc/handler.go
```diff
@@ -406,10 +406,16 @@ type nodeStat struct {
 //	484:
 func (mrh *MinerRestHandler) testNodeStat(w http.ResponseWriter, r *http.Request) {
 	var (
-		id               = r.URL.Query().Get("id")
+		providerId      = r.URL.Query().Get("provider_id")
```
Sriep, line 409: We should keep the interface consistent with the [pre-existing nodeStat endpoint](https://github.com/0chain/0chain/blob/staging/code/go/0chain.net/smartcontract/minersc/handler.go#L467), one way or the other.

id  changed to proivider_id and the addition of proivider_type.

Sriep (dismissed):

code/go/0chain.net/smartcontract/minersc/handler.go
dabasov: we used to use snake case for url parameters, no?

code/go/0chain.net/smartcontract/minersc/handler.go
peterlimg: Since we have new param `provider_type`, so I think use `provider_id` would be better to match it.

peterlimg (approved):
LGTM, please fix the conflicts

peterlimg (approved):
LGTM

code/go/0chain.net/smartcontract/minersc/handler.go
Sriep: This is an extension of the nodeStat endpont so it should use the same parameters. Hense use `id` instead of `provider_id` and no `provider_type` field.

We should keep our endpoint parameters self-consistent where possible. The only effect of this PR is to make our input parameters inconsistent, which will confuse only users.

PR #2044, opened 2023-01-24
Enhancement/blocks management

lpoli:
## Tests

- blobber:
- gosdk:
- system_test:
- zboxcli:
- zwalletcli:
- Other: ...


code/go/0chain.net/sharder/blockstore/cache.go
```diff
@@ -0,0 +1,296 @@
...
+type cache struct {
+	// path to store uncompressed blocks
+	path string
+	// sizeLimit limit of total blocks size in the path
+	sizeLimit int64
+
+	// size current total blocks size in the path
+	size int64
+
+	lru      *lru
```
dabasov: use cache.LRU as main implementation

code/go/0chain.net/sharder/blockstore/fs_store.go
```diff
@@ -2,275 +2,245 @@ package blockstore
...
-func (fbs *FSBlockStore) DeleteBlock(b *block.Block) error {
-	fileName := fbs.getFileName(b.Hash, b.Round)
-	err := os.Remove(fileName)
-	if err != nil {
-		return err
+// Init checks for minimum disk size, inodes requirement and assigns
+// block storer to a variable. If any error occurs during initialization
+// it will panic.
+// It will register read and write function based on whether cache config is provided or not.
+func Init(ctx context.Context, sViper *viper.Viper) {
```
dabasov: ctx is not used

code/go/0chain.net/sharder/blockstore/fs_store.go
```diff
@@ -2,275 +2,245 @@ package blockstore
...
+	switch {
+	default:
+		bStore.write = func(bStore *BlockStore, hash string, b *block.Block) error {
+			return bStore.writeToDisk(hash, b)
+		}
+
+		bStore.read = func(bStore *BlockStore, hash string) (*block.Block, error) {
+			return bStore.readFromDisk(hash)
+		}
+	case cViper != nil:
```
dabasov: why can't you write as 
```go
if cViper != nil {
...
}
//default block
```

code/go/0chain.net/sharder/blockstore/fs_store.go
```diff
@@ -2,275 +2,245 @@ package blockstore
...
+		bStore.write = func(bStore *BlockStore, hash string, b *block.Block) error {
```
dabasov: you better always use cache, but initialize it with no-op cache in case that cache is not used

code/go/0chain.net/sharder/blockstore/fs_store.go
```diff
@@ -2,275 +2,245 @@ package blockstore
...
+			go func() {
+				if err := bStore.writeBlockToCache(hash, b); err != nil {
+					logging.Logger.Error(err.Error())
+				}
+			}()
+
+			return nil
+		}
+
+		bStore.read = func(bStore *BlockStore, hash string) (*block.Block, error) {
```
dabasov: the same use no-op cache in case you don't need it, then you can define `read` in struct as a func

code/go/0chain.net/sharder/blockstore/fs_store.go
```diff
@@ -2,275 +2,245 @@ package blockstore
...
+			return bStore.readFromDisk(hash)
+		}
+	case cViper != nil:
+		bStore.write = func(bStore *BlockStore, hash string, b *block.Block) error {
+			err := bStore.writeToDisk(hash, b)
+			if err != nil {
+				return err
+			}
+
+			go func() {
```
dabasov: what is the reason to do it in different routine? usually cache write/read is something quick, with double linked list and map index it is O(1) for both read and write

dabasov (changes requested):

code/go/0chain.net/sharder/blockstore/cache.go
lpoli: I checked the cache.LRU if you mean from `cache` package in `core` directory.
It is limiting for the required implementation and cannot be modified without affecting other packages that requires it.

code/go/0chain.net/sharder/blockstore/fs_store.go
lpoli: Removing it. Context is not required as well.

code/go/0chain.net/sharder/blockstore/fs_store.go
lpoli: The cache is SSD disk. Maybe you are assuming cache is RAM.

code/go/0chain.net/sharder/blockstore/fs_store.go
lpoli: In this case not other than readability.

code/go/0chain.net/sharder/blockstore/cache.go
lpoli: Cache is SSD disk. To manage blocks in a directory cache.LRU is not enough.
Which is why local implementation is used.

dabasov: your LRu implementation is in memory LRU cache , am I missing something?

dabasov: https://github.com/0chain/0chain/blob/2260920e4ee7e530258dc50c8681c1062d4aa198/code/go/0chain.net/sharder/blockstore/cache.go#L182:L211

code/go/0chain.net/sharder/blockstore/fs_store.go
dabasov: ah, it's cacher not lru, well, anyway, why do we want to write to it concurrently?

code/go/0chain.net/sharder/blockstore/cache.go
lpoli: The main cache is SSD disk which is managed by LRU cache in Ram memory.
Managing two caches requires low level access to the lru and cache.LRU cannot provide that facility.

Check [this use case](https://github.com/0chain/0chain/blob/2260920e4ee7e530258dc50c8681c1062d4aa198/code/go/0chain.net/sharder/blockstore/cache.go#L147).

code/go/0chain.net/sharder/blockstore/fs_store.go
lpoli: Block is primarily stored into main Disk and we waited for it to complete.
Now to write into cache its not necessary to wait because block is safely stored in primary storage. 
The cache write is still a disk write.

code/go/0chain.net/sharder/blockstore/cache.go
dabasov: I can't get you, we already have LRU in the project please use it, what is wrong with it?

lpoli: @dabasov Don't understand what you are not getting.
First thing is that sharder can specify any amount of SSD cache size(better maybe to have it between 500MB-2GB). Minimum I've set is 500MB. Note that this is SSD disk space where the blocks are stored uncompressed.

The size of block is not fixed. Some blocks can have maximum number of transactions and some can have minimum number thus making block size different from each other.

The LRU cache in memory depends on the size of block as well.
So if SSD cache size is 100 MB and block size is for example constant to 1MB then LRU cache can have maximum of 100 key-value data.
Same SSD cache size with block size of 500 KB, LRU cache  can have maximum of 200 key-value data.

The LRU that you are suggesting to use has `size` parameter to initialize cache but size cannot be determined beforehand.

The other thing is; When sharder tries to push a block to SSD cache and finds that SSD cache is full then it will have to remove last used blocks. So we need to have access to doubly-linked-list to get block-id(hash) so as to delete last used block. For example, if sharder is trying to push block of size 1MB and last block is 500KB size and second last block is 600 KB then it will remove two blocks that were last used. This feature is not possible in the LRU that you mentioned.

cache.LRU is independent cache for itself.
The LRU cache that I wrote is co-dependent on SSD cache size and block size.

FYI Before writing own LRU, I've thoroughly checked if I could use existing LRU logic but it did not fit the requirement.

dabasov: calculate size and initialize cache with it, is it a problem? if you need different one, reinitialize it. please, do not repeat already implemented code, we try to use common code as much as possible

```
The LRU that you are suggesting to use has size parameter to initialize cache but size cannot be determined beforehand.
```
please reference where it is dependent? I'll think how to rewrite it

code/go/0chain.net/sharder/blockstore/fs_store.go
dabasov: then use context and timeout writings, if writes will hang, you will cause go routine leakage in current implementation

code/go/0chain.net/sharder/blockstore/cache.go
lpoli: If common LRU cache does not preoccupy memory then it is fine. We can calculate cache size considering minimum size of block.
The other important requirements are:

1. To be able get last element in the cache so as to delete it in LRU cache and SSD cache as well.
2. When new key-value is inserted into LRU cache it should return if [insertion is new](https://github.com/0chain/0chain/blob/enhancement%2Fblocks-management/code/go/0chain.net/sharder/blockstore/cache.go#L208-L218)(a boolean value).

dabasov (approved):
lgtm

service-0chain:
| config  | value |
| tests skipped  | FALSE  |
| system-tests  | master  |
| gosdk  | staging  |
| zboxcli  | staging  |
| zwalletcli  | staging  |
| 0chain  | enhancement/blocks-management  |
| blobber  | staging  |
| authorizer  | staging  |
| 0box  | staging  |
| 0dns  | staging  |

service-0chain:
| config  | value |
| tests skipped  | FALSE  |
| system-tests  | master  |
| gosdk  | staging  |
| zboxcli  | staging  |
| zwalletcli  | staging  |
| 0chain  | enhancement/blocks-management  |
| blobber  | staging  |
| authorizer  | staging  |
| 0box  | staging  |
| 0dns  | staging  |

PR #2048, opened 2023-01-25
Fix testNodeStat

Sriep:
## Fixes
Fix testNodeStat.
* Change to support miners and sharders with no stake pools. SQL + gorm returns an empty delegate pool object when the object has no stake pools. So we now test for this.
* Improve error message to show the errors from event database API.
## Tests

- blobber:
- gosdk:
- system_test:
- zboxcli:
- zwalletcli:
- Other: ...

### DIFF
--- a/code/go/0chain.net/smartcontract/dbs/event/miner.go
+++ b/code/go/0chain.net/smartcontract/dbs/event/miner.go
@@ -73,11 +73,12 @@ func (edb *EventDb) GetMinerWithDelegatePools(id string) (Miner, []DelegatePool,
 	}
 	m = minerDps[0].Miner
 
-	if "" == minerDps[0].ProviderRewards.ProviderID {
-		return m, nil, fmt.Errorf("cannot find provider rewards table for miner %v. rewards %v",
-			id, minerDps[0].ProviderRewards)
-	}
+	//if "" == minerDps[0].ProviderRewards.ProviderID {
+	//	return m, nil, fmt.Errorf("cannot find provider rewards table for miner %v. rewards %v",
+	//		id, minerDps[0].ProviderRewards)
+	//} todo remove
 	m.Rewards = minerDps[0].ProviderRewards
+	m.Rewards.ProviderID = id
 	if len(minerDps) == 1 && minerDps[0].DelegatePool.ProviderID == "" {
 		return m, nil, nil
 	}
--- a/code/go/0chain.net/smartcontract/dbs/event/miner.go
+++ b/code/go/0chain.net/smartcontract/dbs/event/miner.go
@@ -81,7 +81,8 @@ func (edb *EventDb) GetMinerWithDelegatePools(id string) (Miner, []DelegatePool,
 	//} todo
 	for i := 0; i < len(minerDps); i++ {
 		logging.Logger.Info("GetMinerWithDelegatePools Results",
-			zap.Any("delegate pool", minerDps[i].ProviderRewards),
+			zap.Int("index", i),
+			zap.Any("provider_rewards", minerDps[i].ProviderRewards),
 			zap.Any("miner", minerDps[i].Miner),
 			zap.Any("delegate pool", minerDps[i].DelegatePool),
 		)
--- a/code/go/0chain.net/smartcontract/dbs/event/miner.go
+++ b/code/go/0chain.net/smartcontract/dbs/event/miner.go
@@ -5,8 +5,6 @@ import (
 	"errors"
 	"fmt"
 	"github.com/0chain/common/core/currency"
-	"github.com/0chain/common/core/logging"
-	"go.uber.org/zap"
 	"gorm.io/gorm/clause"
 
 	"0chain.net/smartcontract/dbs"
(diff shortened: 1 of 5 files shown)


code/go/0chain.net/smartcontract/dbs/event/miner.go
```diff
@@ -72,11 +72,16 @@ func (edb *EventDb) GetMinerWithDelegatePools(id string) (Miner, []DelegatePool,
...
-			id, minerDps[0].Miner.ID)
-	}
+
+	//if "" == minerDps[0].ProviderRewards.ProviderID {
+	//	return m, nil, fmt.Errorf("cannot find provider rewards table for miner %v. rewards %v",
+	//		id, minerDps[0].ProviderRewards)
+	//} todo remove
 	m.Rewards = minerDps[0].ProviderRewards
+	m.Rewards.ProviderID = id
+	if len(minerDps) == 1 && minerDps[0].DelegatePool.ProviderID == "" {
```
dabasov: why is it possible to have empty provider id?

dabasov (dismissed):

code/go/0chain.net/smartcontract/dbs/event/miner.go
Sriep: I don't know. I am just fixing symptoms at this point.

But if there is no matching row in the ProviderReward table then there will be no error just a default ProviderReward object. Same for delegate pools.

So if the result-slice has only one object and the delegate_pool.provider_id is the default string, then the implication is that there are no delegate pools.

'No delegate pools' is a valid situation, but there should always be a matching row from the ProviderReward table.

peterlimg: I would spent some time to investigate how could the ProviderReward to be empty if it should not be, otherwise we would just hide the issue and if other APIs or functions need to touch the data, they will get invalid data too. Anyway, I checked how the data went to ProviderRewards table, and then found this https://github.com/0chain/0chain/issues/2055. Please have a check and fix.

peterlimg (changes requested):

code/go/0chain.net/smartcontract/dbs/event/miner.go
Sriep: @peterlimg  https://github.com/0chain/0chain/issues/2055 I don't think this is a problem. The Global Node values are read from [readConfig](https://github.com/0chain/0chain/blob/staging/code/go/0chain.net/smartcontract/minersc/models.go#L261) In particular, the GlobalNode.BlockReward is multiplied by 1e10 using ParseZCN, so as far as I can tell the correct value is calculated in `DistributeRewardRandN`.

Sriep: What I noticed when I ran tests with the commented-out lines which printed the ProviderRewards table
```go 
if "" == minerDps[0].ProviderRewards.ProviderID {
		return m, nil, fmt.Errorf("cannot find provider rewards table for miner %v. rewards %v",
			id, minerDps[0].ProviderRewards)
}
```
Is that we were getting a ProviderRewards object with the right data except for a `""` ProviderRewards.ProviderId field.  
```
{"code":"invalid_request","error":"invalid_request: no matching provider for id c6f4b8ce5da386b278ba8c4e6cf98b24b32d15bc675b4d12c95e082079c91937, miner not found: cannot find provider rewards table for miner c6f4b8ce5da386b278ba8c4e6cf98b24b32d15bc675b4d12c95e082079c91937. 
rewards {{0 0001-01-01 00:00:00 +0000 UTC 0001-01-01 00:00:00 +0000 UTC}  2276400000000 2276400000000 4354}, and sharder not found: get sharder c6f4b8ce5da386b278ba8c4e6cf98b24b32d15bc675b4d12c95e082079c91937 found no records"}
```
So the ProviderRewards object.
```
{{0 0001-01-01 00:00:00 +0000 UTC 0001-01-01 00:00:00 +0000 UTC}  2276400000000 2276400000000 4354}
```
provider id is "" so it only shows as an extra space between `}` and `2`. The 2276400000000 and 4354 entries appear to be correct.

I could not understand why this sometimes happens, but I am testing to see if I force the provider id to be correct does everything else work. The block rewards system test is quite sensitive, so incorrect rewards values should be quickly noticed.

service-0chain:
| config  | value |
| tests skipped  | FALSE  |
| system-tests  | block_rewards  |
| gosdk  | NOT OVERRIDDEN  |
| zboxcli  | staging  |
| zwalletcli  | staging  |
| 0chain  | fix_testnodestat  |
| blobber  | staging  |
| authorizer  | staging  |
| 0box  | staging  |
| 0block  | staging  |
| 0dns  | staging  |
| explorer  | master  |
| 0proxy  | staging  |
| 0search  | staging  |

code/go/0chain.net/smartcontract/dbs/event/miner.go
peterlimg: @Sriep I left a comment in issue #2055, it might be able to fix the issue.

peterlimg (approved):
lgtm

service-0chain:
| config  | value |
| tests skipped  | FALSE  |
| system-tests  | master  |
| gosdk  | NOT OVERRIDDEN  |
| zboxcli  | areeb/allocname  |
| zwalletcli  | staging  |
| 0chain  | staging  |
| blobber  | staging  |
| authorizer  | smart-contract-areeb  |
| 0box  | areeb/allocname  |
| 0block  | staging  |
| 0dns  | staging  |
| explorer  | master  |
| 0proxy  | staging  |
| 0search  | staging  |

service-0chain:
| config  | value |
| tests skipped  | FALSE  |
| system-tests  | master  |
| gosdk  | NOT OVERRIDDEN  |
| zboxcli  | areeb/allocname  |
| zwalletcli  | staging  |
| 0chain  | staging  |
| blobber  | staging  |
| authorizer  | staging  |
| 0box  | areeb/allocname  |
| 0block  | staging  |
| 0dns  | staging  |
| explorer  | master  |
| 0proxy  | staging  |
| 0search  | staging  |

service-0chain:
| config  | value |
| tests skipped  | FALSE  |
| system-tests  | master  |
| gosdk  | NOT OVERRIDDEN  |
| zboxcli  | staging  |
| zwalletcli  | staging  |
| 0chain  | staging  |
| blobber  | staging  |
| authorizer  | staging  |
| 0box  | staging  |
| 0block  | staging  |
| 0dns  | staging  |
| explorer  | master  |
| 0proxy  | staging  |
| 0search  | staging  |

PR #2090, opened 2023-02-07
Add config cache for smartcontracts

### LINKED ISSUE #1392: Load config from MPT on node starts and update on block finalization

din-mukhammed:
## Fixes
- #1392 
## Tests

- blobber:
- gosdk:
- system_test:
- zboxcli:
- zwalletcli:
- Other: ...

### DIFF
--- a/code/go/0chain.net/smartcontract/faucetsc/config.go
+++ b/code/go/0chain.net/smartcontract/faucetsc/config.go
@@ -1,6 +1,7 @@
 package faucetsc
 
 import (
+	"sync"
 	"time"
 
 	cstate "0chain.net/chaincore/chain/state"
@@ -54,6 +55,21 @@ type FaucetConfig struct {
 	Cost            map[string]int `json:"cost"`
 }
 
+type cache struct {
+	config *FaucetConfig
+	l      sync.RWMutex
+}
+
+var c = &cache{
+	l: sync.RWMutex{},
+}
+
+func (*cache) update(conf *FaucetConfig) {
+	c.l.Lock()
+	c.config = conf
+	c.l.Unlock()
+}
+
 // configurations from sc.yaml
 func getFaucetConfig() (conf *FaucetConfig, err error) {
 
@@ -82,19 +98,21 @@ func getFaucetConfig() (conf *FaucetConfig, err error) {
 }
 
 func InitConfig(balances cstate.CommonStateContextI) error {
-	gn := new(GlobalNode)
+	c.l.Lock()
+	defer c.l.Unlock()
+
+	var gn = &GlobalNode{ID: ADDRESS}
 	err := balances.GetTrieNode(globalNodeKey, gn)
-	if err != nil {
-		if err != util.ErrValueNotPresent {
-			return err
-		}
+	if err == util.ErrValueNotPresent {
 		gn.FaucetConfig, err = getFaucetConfig()
 		if err != nil {
 			return err
 		}
-		gn.ID = ADDRESS
 		_, err = balances.InsertTrieNode(globalNodeKey, gn)
-		return err
+		if err != nil {
+			return err
+		}
 	}
-	return nil
+	c.config = gn.FaucetConfig
+	return err
 }
--- a/code/go/0chain.net/smartcontract/faucetsc/handler.go
+++ b/code/go/0chain.net/smartcontract/faucetsc/handler.go
@@ -13,7 +13,6 @@ import (
 	"0chain.net/core/common"
 	"0chain.net/smartcontract"
 	"github.com/0chain/common/core/currency"
-	"github.com/0chain/common/core/util"
 )
 
 const (
@@ -63,16 +62,7 @@ func (frh *FaucetscRestHandler) getConfig(w http.ResponseWriter, r *http.Request
 		return
 	}
 
-	var faucetConfig *FaucetConfig
-	if gn.FaucetConfig == nil {
-		faucetConfig, err = getFaucetConfig()
-		if err != nil {
-			NoResourceOrErrInternal(w, r, err)
-			return
-		}
-	} else {
-		faucetConfig = gn.FaucetConfig
-	}
+	var faucetConfig = gn.FaucetConfig
 
 	pourAmount, err := faucetConfig.PourAmount.ToZCN()
 	if err != nil {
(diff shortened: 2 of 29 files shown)


code/go/0chain.net/smartcontract/faucetsc/config.go
```diff
@@ -82,19 +98,21 @@ func getFaucetConfig() (conf *FaucetConfig, err error) {
 }
 
 func InitConfig(balances cstate.CommonStateContextI) error {
-	gn := new(GlobalNode)
+	c.l.Lock()
+	defer c.l.Unlock()
```
peterlimg, line 102: We should avoid using the mutex out of the `cache` methods.

code/go/0chain.net/smartcontract/faucetsc/handler.go
```diff
@@ -194,17 +184,17 @@ func (frh *FaucetscRestHandler) getPersonalPeriodicLimit(w http.ResponseWriter,
...
-	err := sctx.GetTrieNode(gn.GetKey(), &gn)
-	if err != nil {
-		if err != util.ErrValueNotPresent {
-			return gn, err
-		}
-		gn.FaucetConfig, err = getFaucetConfig()
+func getGlobalNode(sctx state.QueryStateContextI) (node *GlobalNode, err error) {
+	c.l.RLock()
+	if c.config == nil {
+		c.l.RUnlock()
```
peterlimg, line 190: Same as above, should never use lock outside of the cache.

code/go/0chain.net/smartcontract/faucetsc/sc.go
```diff
@@ -152,6 +152,7 @@ func (fc *FaucetSmartContract) updateSettings(
 	if err != nil {
 		return "", common.NewError("update_settings", "saving global node: "+err.Error())
 	}
+	c.update(gn.FaucetConfig)
```
peterlimg, line 155: We need to update the cache when the block is finalized rather than immediately when SC is executed. Imagine two different txns update the config in different block, but only one of the block will get finalized.

code/go/0chain.net/smartcontract/minersc/sc.go
```diff
@@ -44,6 +44,23 @@ var (
 	}
 )
 
+type cache struct {
+	gnode *GlobalNode
+	l     sync.RWMutex
+	err   error
```
peterlimg, line 50: We don't need to save err in the cache.

peterlimg (changes requested):
The SC config cache is basically a minor feature of MPT cache, and changes should not be updated to the cache immediately when SC is exeucted, instead we need to update it on block finalization. Also, we need this cache because we want to reduce the MPT access when executing SCs,  but I don't see the cache is used for that in this PR. It's a none trivial feature actually, we will also need to consider whether we should use the cached config or the config in MPT when the config updating txn is not finalized yet. Therefore, I suggest to hold on this issue and PR for a while, I'll take over this issue later maybe.

din-mukhammed:
> The SC config cache is basically a minor feature of MPT cache, and changes should not be updated to the cache immediately when SC is exeucted, instead we need to update it on block finalization. Also, we need this cache because we want to reduce the MPT access when executing SCs, but I don't see the cache is used for that in this PR. It's a none trivial feature actually, we will also need to consider whether we should use the cached config or the config in MPT when the config updating txn is not finalized yet. Therefore, I suggest to hold on this issue and PR for a while, I'll take over this issue later maybe.

This PR updates the config cache in response to MPT config updates in SC packages through the `InsertTrieNode` function, rather than doing so when block is finalized, as suggested by your comments [here](https://github.com/0chain/0chain/pull/1692#discussion_r1038663385). It caches the SC configs in package variables during config initialization and updates them when MPT config changes are made through `InsertTrieNode`. This eliminates the need for `GetTrieNode` calls when using the `getConfig` functions.

dabasov:
@peterlimg @din-mukhammed I don't think it is possible to have any cache rather then in state-context itself. the problem is that every snapshot of MPT (every block version of it) should have it's own cache, the same as every block has it's own MPT. 
1. Package variable obviously doesn't work, since change config transaction can happen in 2 different blocks concurrently processed, it is not possible to overwrite one version by another.
2. Refreshing cache on finalization won't work too, since it will create a window of stale config, it can happen that client updates config and wants execute another transaction immediately, it will give 2 different results depending on block when it was executed (next to this or after finalization).

The only good solution is to keep cache close to MPT (I started it once but haven't finished) for now I suggest to keep configs cached in stateContext, we can make it as a singleton with lazy loading and initialize on the first access from the MPT, if it is updated then refresh this cache. Your thoughts?

din-mukhammed:
> @peterlimg @din-mukhammed I don't think it is possible to have any cache rather then in state-context itself. the problem is that every snapshot of MPT (every block version of it) should have it's own cache, the same as every block has it's own MPT.
> 
>     1. Package variable obviously doesn't work, since change config transaction can happen in 2 different blocks concurrently processed, it is not possible to overwrite one version by another.
> 
>     2. Refreshing cache on finalization won't work too, since it will create a window of stale config, it can happen that client updates config and wants execute another transaction immediately, it will give 2 different results depending on block when it was executed (next to this or after finalization).
> 
> 
> The only good solution is to keep cache close to MPT (I started it once but haven't finished) for now I suggest to keep configs cached in stateContext, we can make it as a singleton with lazy loading and initialize on the first access from the MPT, if it is updated then refresh this cache. Your thoughts?

I believe implementing it in this manner would be a good approach. I would like to proceed with this plan, pending approval from @peterlimg.

peterlimg:
@dabasov yes, I agree that each block need to have it's own cache, and yes, it would not work on updating immediately or on finalizing blocks as you explained. 

But for now, @din-mukhammed I'd prefer to use the config from MPT directly to minimize possibilities of the `state hash mismatch` error, as any chance the cache was not updated or used correctly, we will see different state. So please hold on this issue, and maybe do it after mainnet, when we can have enough time to handle the potential issues.

PR #2101, opened 2023-02-11
2085 unit tests to check mpt consistency in scs

### LINKED ISSUE #2085: Unit tests to check MPT consistency in SCs
- All map iterating code in Smart contracts should be thread-safe for read write and write.

M0706:
## Fixes

Added locks to prevent to avoid random map accessing

## Tests

- blobber:
- gosdk:
- system_test:
- zboxcli:
- zwalletcli:
- Other: ...

### DIFF
--- a/code/go/0chain.net/chaincore/block/mpk.go
+++ b/code/go/0chain.net/chaincore/block/mpk.go
@@ -2,6 +2,7 @@ package block
 
 import (
 	"encoding/json"
+	"sync"
 
 	"0chain.net/chaincore/threshold/bls"
 	"0chain.net/core/encryption"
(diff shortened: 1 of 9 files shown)


peterlimg (changes requested):
The lock is not the problem, you could think all transactions are executed in single thread. Currently, I would only require test code to check if running the same smart contract with the same input would result in the same MPT state root. For example: [TestVerifyChallengeRunMultipleTimes](https://github.com/0chain/0chain/blob/b23843eb6e57429c2a78c223e0aec13a7764e34d/code/go/0chain.net/smartcontract/storagesc/challenge_test.go#L561-L599). 


There's no need to change the smart contracts until we find new places that caused the `state hash mismatch` error. And we need tests like above for all SCs.

peterlimg:
Close this PR as it does not fix the issue as explained above, contact me for the details if need to continue.

PR #2158, opened 2023-02-24
Add Blobber and validator, kill and shutdown endpoints

### LINKED ISSUE #1020: Delete provider plus blacklist

### LINKED ISSUE #1040: Providers

### LINKED ISSUE #1044: Provider health check

### LINKED ISSUE #1045: Provider shut down

### LINKED ISSUE #1046: Kill provder

Sriep:
## Changes
Closes https://github.com/0chain/0chain/issues/1040, Closes https://github.com/0chain/0chain/issues/1044, Closes https://github.com/0chain/0chain/issues/1045, Closes https://github.com/0chain/0chain/issues/1046, Closes https://github.com/0chain/0chain/issues/1020
1. New `health_check_period` setting.
2. Providers can now be deactivated. While deactivated providers will not be available for further commitmetns.
3. Two endpoints for validators and blobbers.
* `Shut down` Permianlty deactivates a provider. It can still receive rewards for existing commitments. Only the provider can deactivate itself.
* `Kill` Permianlty deactivates a provider with prejudice. It will no longer be able to receive rewards. Only the 0chain's owner can deactivate a provider. Stakeholders lose a percentage of their stake. New `sc.yaml` settings for this fraction.
4. Shut down and killed blobbers are no longer available for new allocation and are removed from `/blobbers` endpoint.
5. Killed blobbers are removed from partitions used for challenges and block rewards.
6. Killed and shutdown validators are removed from the partition for challenge validation.
7. New fields to indicate killed and shut down added to providers and stake pools. 
8. Distribute rewards not carried out for killed providers.
9. get blobbers for new allocations. Ignores killed or shutdown blobbers

This only handles the 0chian changes. Resolving killed blobbers that are part of an allocation will need to be dealt with elsewhere.
## Tests

- blobber:
- gosdk:https://github.com/0chain/gosdk/pull/797
- system_test: https://github.com/0chain/system_test/pull/576
- zboxcli: https://github.com/0chain/zboxcli/pull/270
- zwalletcli:
- Other: ...


service-0chain:
| config  | value |
| tests skipped  | FALSE  |
| system-tests  | master  |
| gosdk  | kill_shutdown  |
| zboxcli  | provider_kill  |
| zwalletcli  | staging  |
| 0chain  | blobber_kill_shutdown  |
| blobber  | staging  |
| authorizer  | staging  |
| 0box  | staging  |
| 0block  | staging  |
| 0dns  | staging  |
| explorer  | staging  |
| 0proxy  | staging  |
| 0search  | staging  |

service-0chain:
| config  | value |
| tests skipped  | FALSE  |
| system-tests  | master  |
| gosdk  | NOT OVERRIDDEN  |
| zboxcli  | staging  |
| zwalletcli  | staging  |
| 0chain  | blobber_kill_shutdown  |
| blobber  | staging  |
| authorizer  | staging  |
| 0box  | staging  |
| 0block  | staging  |
| 0dns  | staging  |
| explorer  | master  |
| 0proxy  | staging  |
| 0search  | staging  |

code/go/0chain.net/smartcontract/dbs/event/provider.go
```diff
@@ -27,6 +32,8 @@ type Provider struct {
 	Rewards         ProviderRewards  `json:"rewards" gorm:"foreignKey:ProviderID"`
 	Downtime        uint64           `json:"downtime"`
 	LastHealthCheck common.Timestamp `json:"last_health_check"`
+	IsKilled        bool             `json:"is_killed"`
```
dabasov, line 35: is it better to have single column with Status there? I mean can provider be killed and not shutdown at the same time?

code/go/0chain.net/smartcontract/dbs/event/provider.go
```diff
@@ -117,3 +124,47 @@ func (edb *EventDb) ReplicateProviderAggregates(round int64, limit int, offset i
 	}
 	return nil
 }
+
+func providerToTableName(pType spenum.Provider) string {
+	return pType.String() + "s"
+}
+
+func splitProviders(
```
dabasov: usually it is called simply mapProviders()

code/go/0chain.net/smartcontract/dbs/types.go
```diff
@@ -7,30 +7,39 @@ import (
...
+			Id:      id,
+			Updates: make(map[string]interface{}),
+		},
+		Type: typ,
 	}
+
 }
 
-type StakePoolId struct {
+type Provider struct {
```
dabasov: will it collude with Provider entity in event package?

code/go/0chain.net/smartcontract/provider/kill.go
```diff
@@ -0,0 +1,81 @@
...
+	if p.IsShutDown() {
+		return fmt.Errorf("already shutdown")
+	}
+	if p.IsKilled() {
+		return fmt.Errorf("already killed")
+	}
+	p.Kill()
+
+	sp.Kill()
+	if err := sp.SlashFraction(
```
dabasov: sp.SlashFraction can be called inside sp.Kill()

code/go/0chain.net/smartcontract/stakepool/stakepool.go
```diff
@@ -302,6 +314,31 @@ func (sp *StakePool) Empty(sscID, poolID, clientID string, balances cstate.State
...
+) error {
+	if fraction == 0.0 {
+		return nil
+	}
+	for _, dp := range sp.Pools {
+		dpSlash, err := currency.Float64ToCoin(float64(dp.Balance) * fraction)
+		if err != nil {
+			return err
+		}
+		dp.Balance, err = currency.MinusCoin(dp.Balance, dpSlash)
```
dabasov: what if it is negative? provider won't be killed, right? Some delegate pools can be small enough, you should slash proportionally to stake amount at least

code/go/0chain.net/smartcontract/provider/provider.go
```diff
@@ -1,17 +1,70 @@
...
+	"fmt"
+
+	"0chain.net/core/common"
 	"0chain.net/core/datastore"
 	"0chain.net/smartcontract/stakepool/spenum"
 )
 
 //go:generate msgp -io=false -tests=false -v
 
+type Abstract interface {
```
dabasov: AbstractProvider?

code/go/0chain.net/smartcontract/provider/provider.go
```diff
@@ -1,17 +1,70 @@
...
 }
 
-type Provider struct {
-	ID           string          `json:"id" validate:"hexadecimal,len=64"`
-	ProviderType spenum.Provider `json:"provider_type"`
+func (p *Provider) Id() string {
+	return p.ID
+}
+
+func (p *Provider) IsActive(now, healthCheckPeriod common.Timestamp) (bool, string) {
```
dabasov, line 39: can we have status instead with values Active, Inactive, Shutdown, Killed values?

code/go/0chain.net/smartcontract/stakepool/spenum/enums.go
```diff
@@ -9,7 +9,8 @@ func init() {
 type Provider int
 
 const (
-	Miner Provider = iota + 1
+	Invalid Provider = iota
```
dabasov: we don't need it, simply when get from map check whether status is found or not

code/go/0chain.net/smartcontract/storagesc/kill.go
```diff
@@ -0,0 +1,126 @@
...
+	tx *transaction.Transaction,
+	input []byte,
+	balances cstate.StateContextI,
+) (string, error) {
+	conf, err := getConfig(balances)
+	if err != nil {
+		return "", common.NewError("can't get config", err.Error())
+	}
+
+	var blobber = newBlobber("")
```
dabasov: why not `&Blobber{}`

code/go/0chain.net/smartcontract/storagesc/kill.go
```diff
@@ -0,0 +1,126 @@
...
+	tx *transaction.Transaction,
+	input []byte,
+	balances cstate.StateContextI,
+) (string, error) {
+	conf, err := getConfig(balances)
+	if err != nil {
+		return "", common.NewError("can't get config", err.Error())
+	}
+
+	var validator = newValidator("")
```
dabasov: same here

dabasov (changes requested):
also update sc.yaml used for benchmark

code/go/0chain.net/smartcontract/dbs/event/provider.go
Sriep: [I am waiting for feedback on that.](https://0chain.slack.com/archives/C03788G8N2D/p1677245719377479). I was keeping it simple until I was sure what way we want to do it.

The initial suggestion was that both states be immutable. My thought is that you might want to kill a shutdown provider. Killing is punitive with loss of stake. "If you believe your provider will be killed, you should shut it down preemptively to protect your stake." To avoid this you might allow killing a shutdown provider, put these together, and you would have a provider that is both killed and shut down.

code/go/0chain.net/smartcontract/dbs/types.go
Sriep: I don't see how? `dbs.Provider` in a different package from the other Provider objects, so as far as golang code is concerned. It is not stored directly in a database but instead used to transfer messages between smart contracts and the event database. Plus the old name was just wrong.

code/go/0chain.net/smartcontract/stakepool/stakepool.go
Sriep: If the result is negative then an error will be returned, which is what we want. Simplified the code a bit.

code/go/0chain.net/smartcontract/provider/provider.go
Sriep: What dose `Active` and `Inactive` mean exactly?

That is what I tried the first time. On reflection, I decieded a single `IsActive` method was simpler and needed less ongoing maintenance.

code/go/0chain.net/smartcontract/storagesc/kill.go
Sriep: Why not not? Initially, it was done this way because I embedded `*Provider` into StorageNode, this was changed to embed `Provider` so either way now works. I left the code as it was so that if later StorageNode needs default initialisation it can all be done in one place.

code/go/0chain.net/smartcontract/stakepool/spenum/enums.go
Sriep: It was added to give [ToProviderType](https://github.com/0chain/0chain/blob/000ff47a73fe23f85be4318ed2aee47bc0e6c5a2/code/go/0chain.net/smartcontract/stakepool/spenum/enums.go#L39) something to return if called with an invalid string.

service-0chain:
| config  | value |
| tests skipped  | FALSE  |
| system-tests  | kill_shudown_config  |
| gosdk  | NOT OVERRIDDEN  |
| zboxcli  | staging  |
| zwalletcli  | staging  |
| 0chain  | blobber_kill_shutdown  |
| blobber  | staging  |
| authorizer  | staging  |
| 0box  | staging  |
| 0block  | staging  |
| 0dns  | staging  |
| explorer  | master  |
| 0proxy  | staging  |
| 0search  | staging  |

service-0chain:
| config  | value |
| tests skipped  | FALSE  |
| system-tests  | kill_shudown_config  |
| gosdk  | NOT OVERRIDDEN  |
| zboxcli  | staging  |
| zwalletcli  | staging  |
| 0chain  | blobber_kill_shutdown  |
| blobber  | staging  |
| authorizer  | staging  |
| 0box  | staging  |
| 0block  | staging  |
| 0dns  | staging  |
| explorer  | master  |
| 0proxy  | staging  |
| 0search  | staging  |

Sriep:
https://github.com/0chain/0chain/actions/runs/4342756056/jobs/7583967491 green system test with https://github.com/0chain/system_test/pull/576

code/go/0chain.net/smartcontract/dbs/types.go
dabasov: Just logically, we have too much Providers in different packages. Can we call id `ID`, as it is composite id by nature

code/go/0chain.net/smartcontract/stakepool/stakepool.go
dabasov: but we don't want to break the killing process, we want to force kill and get all the money possible from the stakepool if not enough

code/go/0chain.net/smartcontract/storagesc/kill.go
dabasov: it's a natural way for golang to use default initialization `&Blobber{}` when possible, it's more obvious what happens when you read code, imagine you are reading code and see `newBlobber("")` do you know what will happen inside this func?

code/go/0chain.net/smartcontract/dbs/types.go
Sriep: Not quite sure what you mean, but I renamed Provider to ProviderID. 

The field names were chosen to match the json tags, ProviderId for provider_id. Although it might make sense to change ProviderId to ID as the struct name implies the provider part. Anyway changed to 
```go
type ProviderID struct {
	ID   string          `json:"provider_id"`
	Type spenum.Provider `json:"provider_type"`
}
```

service-0chain:
| config  | value |
| tests skipped  | ABILITY TO SKIP TESTS IS DISABLED. SKIP_TESTS OPTION IGNORED.  |
| system-tests  | kill_shudown_config  |
| gosdk  | NOT OVERRIDDEN  |
| zboxcli  | staging  |
| zwalletcli  | staging  |
| 0chain  | blobber_kill_shutdown  |
| blobber  | staging  |
| authorizer  | staging  |
| 0box  | staging  |
| 0block  | staging  |
| 0dns  | staging  |
| explorer  | master  |
| 0proxy  | staging  |
| 0search  | staging  |

service-0chain:
| config  | value |
| tests skipped  | ABILITY TO SKIP TESTS IS DISABLED. SKIP_TESTS OPTION IGNORED.  |
| system-tests  | kill_shudown_config  |
| gosdk  | NOT OVERRIDDEN  |
| zboxcli  | staging  |
| zwalletcli  | staging  |
| 0chain  | blobber_kill_shutdown  |
| blobber  | staging  |
| authorizer  | staging  |
| 0box  | staging  |
| 0block  | staging  |
| 0dns  | staging  |
| explorer  | master  |
| 0proxy  | staging  |
| 0search  | staging  |

dabasov (approved):
lgtm

service-0chain:
| config  | value |
| tests skipped  | ABILITY TO SKIP TESTS IS DISABLED. SKIP_TESTS OPTION IGNORED.  |
| system-tests  | kill_shudown_config  |
| gosdk  | staging  |
| zboxcli  | staging  |
| zwalletcli  | staging  |
| 0chain  | blobber_kill_shutdown  |
| blobber  | staging  |
| authorizer  | staging  |
| 0box  | staging  |
| 0block  | staging  |
| 0dns  | staging  |
| explorer  | staging  |
| 0proxy  | staging  |
| 0search  | staging  |

service-0chain:
| config  | value |
| tests skipped  | TRUE  |
| system-tests  | master  |
| gosdk  | NOT OVERRIDDEN  |
| zboxcli  | kill_shudown_config  |
| zwalletcli  | staging  |
| 0chain  | blobber_kill_shutdown  |
| blobber  | staging  |
| authorizer  | staging  |
| 0box  | staging  |
| 0block  | staging  |
| 0dns  | staging  |
| explorer  | master  |
| 0proxy  | staging  |
| 0search  | staging  |

code/go/0chain.net/smartcontract/stakepool/stakepool.go
```diff
@@ -302,6 +322,38 @@ func (sp *StakePool) Empty(sscID, poolID, clientID string, balances cstate.State
...
+		return fmt.Errorf("kill slash %v should be in the interval [0,1]", killSlashFraction)
+	}
+	reduction := 1 - killSlashFraction
+	if reduction < 0 {
+		reduction = 0
+	}
+	if reduction > 1 {
+		reduction = 1
+	}
+	for _, dp := range sp.Pools {
```
peterlimg, line 346: We'd better process pools in order to avoid potential state mismatch. We can use `sp.OrderedPoolIds` to get ids first, get pool by id and process it.

Sriep: I am uneasy with the suggestion that unit64 arithmetic is not commutative. The only error message `ErrUint64AddOverflow` from `AddCoin` is a constant and does not contain the values causing the error. 

Ordering the pools is a slight overhead. Do we need to do this?

peterlimg: `currency.MultFloat64()` could return two kinds of error, the `ErrNegativeValue` and `ErrUint64OverflowsFloat64`. So if any chance miner1 got `ErrNegativeValue` first and `miner2 got the other one first, then we will see different MPT state. We MUST execute code in order and avoid random MAP iterating. Also see issue https://github.com/0chain/0chain/issues/2085#issuecomment-1421697224. 

We will also exame all the SCs to see if there're other places that do this random MAP accessing, so before fixing exist ones, let's not add new one.

Sriep: Ok, but these problems are not related to this PR, but should be handled in a separate issue.

PR #2172, opened 2023-02-27
Calculate client_aggregates on change

manasnisar:
## Tests

- blobber:
- gosdk:
- system_test:
- zboxcli:
- zwalletcli:
- Other: ...


dabasov:
unfortunately, it won't help much, since we still will make a lot of select every time we recalculate user_aggregates. we need to recalculate exactly when changes happen, so we need to catch events on every change in `process()` the same as we do for snapshots, and when particular event is emitted, we should recalculate user_aggregate. 

we also can try to aggregate different events for a particular user (can require some tricky coding), or we can try to keep map of user_id:user_aggregate in memory and update it on every event (load snapshot it only at first time and keep in memory), then persist on processing end (when no more events left)

PR #2175, opened 2023-02-27
Filter Validators 

Hitenjain14:
## Changes
Filter validators according to LastHealthCheck . 

## Tests

- blobber:
- gosdk:
- system_test:
- zboxcli:
- zwalletcli:
- Other: ...


dabasov (approved):
lgtm

dabasov (approved):
lgtm

code/go/0chain.net/smartcontract/storagesc/challenge.go
```diff
@@ -974,9 +974,24 @@ func (sc *StorageSmartContract) populateGenerateChallenge(
...
+	now := txn.CreationDate
+	filterValidator := filterHealthyValidators(now)
+
 	for i := 0; i < needValidNum; i++ {
 		randValidator := randValidators[perm[i]]
 		if randValidator.Id != blobberID {
+			validator, err := getValidator(randValidator.Id, balances)
+
+			if err != nil {
+				continue
```
peterlimg, line 975: We need to check if the err is `node not found` error, and we need to return immediately if it is to avoid `mismatched state` error that will cause network stuck. So we can do

```go
if cstate.ErrInvalidState(err) {
    return err
}
```

code/go/0chain.net/smartcontract/storagesc/challenge.go
```diff
@@ -974,9 +974,24 @@ func (sc *StorageSmartContract) populateGenerateChallenge(
...
+			}
+
+			kick, err := filterValidator(validator)
```
peterlimg, line 978: Check if the returned `err` is necessary, I find there's no error returned in `filterHealthyValidators` actualy.

peterlimg (changes requested):

code/go/0chain.net/smartcontract/storagesc/challenge.go
Hitenjain14: No there is no error returned , I was following the convention used for `filterHealthyBlobbers` .

peterlimg: okay, but please check error and return if any. Doing this is to make sure we won't get mismatched state in the future when the filter function involves MPT operation and return error.

code/go/0chain.net/smartcontract/storagesc/challenge.go
peterlimg: We don't need to include the `mismatched state` msg to the error, it is the one of the result that we will have if we don't return err here, the error itself is not `mismatched state`.

peterlimg (changes requested):

peterlimg (approved):
LGTM

Hitenjain14:
Should not be merged until we make sure validators are sending health check txns .

Hitenjain14:
This is the error we are getting  - 
    `Bad transaction cost	{"error": "no cost given for validator_health_check", "txn_hash": "569161647bbb2ccbfe53affe3cab90d96eacf3d5576b96e62f46fd89c3383a14"}` , I have asked @m-s-a-c to add the value and update the helm chart .

shahnawaz-creator:
added

PR #2187, opened 2023-03-03
blobber slash from delegate balance

### LINKED ISSUE #2122: Blobber slash should be taken from stake pool not rewards.

Hitenjain14:
## Fixes
Blobber slash is now deducted from [DelegatePool.Balance](https://github.com/0chain/0chain/blob/staging/code/go/0chain.net/smartcontract/dbs/event/delegate_pool.go#L24) and added to [DelegatePool.TotalPenalty](https://github.com/0chain/0chain/blob/staging/code/go/0chain.net/smartcontract/dbs/event/delegate_pool.go#L27).
#2122 

## Changes
Added new tag for `TagStakePoolPenalty`

- blobber:
- gosdk:
- system_test:
- zboxcli:
- zwalletcli:
- Other: ...


code/go/0chain.net/smartcontract/storagesc/stakepool.go
```diff
@@ -259,8 +259,8 @@ func (sp *stakePool) slash(
 		}
 		edbSlash.DelegatePenalties[id] = dpSlash
 	}
-	// todo we should slash from stake pools not rewards. 0chain issue 1495
-	if err := edbSlash.Emit(event.TagStakePoolReward, balances); err != nil {
+	//Added New Tag for StakePoolPenalty
+	if err := edbSlash.Emit(event.TagStakePoolPenalty, balances); err != nil {
```
dabasov, line 263: is stake pool reduced for this amount? I don't see the reduction. At first we should reduce the stakepool and all the delegate pools proportionally in the MPT, and only after it update the events_db. looks like with current implementation events_db will be changed, but MPT won't and balances will be different in different storages

dabasov (changes requested):

code/go/0chain.net/smartcontract/storagesc/stakepool.go
dabasov: here is the code 
```go
		if balance, err := currency.MinusCoin(dp.Balance, dpSlash); err != nil {
			return 0, err
		} else {
			dp.Balance = balance
		}
```

dabasov (approved):
lgtm

Kishan-Dhakan:
@Hitenjain14 can we add system-tests for this?

peterlimg (approved):
LGTM

PR #2201, opened 2023-03-08
Optimize cancel_allocation SC

### LINKED ISSUE #572: Cancel allocation request and BlobberChallenge object - improve performance
https://github.com/0chain/0chain/issues/533 should improve cancelAllocationRequest benchmark performance. This should be enough for `finalize_allocation`, but `cancel_allocation` had another problem due to its need to handle open challenges.

The second problem is in [canceledPassRates](https://github.com/0chain/0chain/blob/staging/code/go/0chain.net/smartcontract/storagesc/allocation.go#L1145). In particular unmarshalling blobber challenges with 100 challenges seems to be take around 5ms per blobber  [BlobberChallenge.Decode](https://github.com/0chain/0chain/blob/staging/code/go/0chain.net/smartcontract/storagesc/models.go#L117)
```go
func (sn *BlobberChallenge) Decode(input []byte) error {
	err := json.Unmarshal(input, sn)
	if err != nil {
		return err
	}
```

peterlimg:
## Fixes
- closes #572
## Changes
- Optimize cancel_allocation SC from around `14ms` to about `3.8ms`. Alloc blobbers' stake pools save process is still slow, so the more blobbers an alloc has, the slower the cancel_allocation SC would be. The benchmark test seems only has 4 blobbers each alloc. So it could be slower if an allocation has 12 blobbers. One way to improve it is do not save rewards to all blobbers' stake pools at once. Instead, we can save the rewards into an `alloc_blobber_rewards` partition, and distribute the rewards to one blobber each round in `payFees` txn.  In this way, no matter how many blobbers each alloc has, the cancel_allocation SC execution time would be the same all the time. 
- Removed unnecessary blobber storage node loading and saving
- Fixed a mistake on saving stake pools. Rewards didn't saved correctly, there were coins distributing to pools after saving. 

## Tests

- blobber:
- gosdk:
- system_test:
- zboxcli:
- zwalletcli:
- Other: ...

### DIFF
--- a/code/go/0chain.net/smartcontract/storagesc/allocation.go
+++ b/code/go/0chain.net/smartcontract/storagesc/allocation.go
@@ -1660,12 +1660,12 @@ func (sc *StorageSmartContract) finishAllocation(
 	}
 	fmt.Println("distribute rewards", time.Since(st))
 
-	st = time.Now()
-	var blobbers []*StorageNode
-	if blobbers, err = sc.getAllocationBlobbers(alloc, balances); err != nil {
-		return fmt.Errorf("could not get alloc blobbers: %v", err)
-	}
-	fmt.Println("get allocation blobbers", time.Since(st))
+	//st = time.Now()
+	//var blobbers []*StorageNode
+	//if blobbers, err = sc.getAllocationBlobbers(alloc, balances); err != nil {
+	//	return fmt.Errorf("could not get alloc blobbers: %v", err)
+	//}
+	//fmt.Println("get allocation blobbers", time.Since(st))
 
 	var cp *challengePool
 	if cp, err = sc.getChallengePool(alloc.ID, balances); err != nil {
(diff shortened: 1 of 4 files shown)


code/go/0chain.net/smartcontract/storagesc/allocation.go
```diff
@@ -1456,8 +1456,10 @@ func (sc *StorageSmartContract) cancelAllocationRequest(
 	if err = req.decode(input); err != nil {
 		return "", common.NewError("alloc_cancel_failed", err.Error())
 	}
+	ts := time.Now()
 	var alloc *StorageAllocation
 	alloc, err = sc.getAllocation(req.AllocationID, balances)
+	fmt.Println("get allocation", time.Since(ts))
```
dabasov: we better clean up. fmt

dabasov (approved):
lgtm, clean up timing logic please

service-0chain:
| config  | value |
| tests skipped  | TRUE  |
| system-tests  | master  |
| gosdk  | NOT OVERRIDDEN  |
| zboxcli  | staging  |
| zwalletcli  | staging  |
| 0chain  | optmize/cancel-alloc  |
| blobber  | staging  |
| authorizer  | staging  |
| 0box  | staging  |
| 0block  | staging  |
| 0dns  | staging  |
| explorer  | master  |
| 0proxy  | staging  |
| 0search  | staging  |

PR #2226, opened 2023-03-19
Extend benchmark tests with MPT hash root mismatch check

### LINKED ISSUE #2085: Unit tests to check MPT consistency in SCs

din-mukhammed:
## Changes

- Benchmark tests have been extended with a MPT state root mismatch check. This additional check helps detect inconsistencies in the MPT state, ensuring the same function produces the same MPT state when run multiple times. Additionally, a run counter has been added which displays the number of iterations for each function being tested.

This check has been tested by introducing code that purposefully causes a hash root mismatch.

## Tests

- blobber:
- gosdk:
- system_test:
- zboxcli:
- zwalletcli:
- Other: ...

### DIFF
--- a/.github/workflows/conductor-authorizer-tests.yml
+++ b/.github/workflows/conductor-authorizer-tests.yml
@@ -8,7 +8,7 @@ on:
 jobs:
   CONDUCTOR_AUTHORIZER_TESTS:
     name: CONDUCTOR-AUTHORIZER-TESTS
-    runs-on: [self-hosted, arc-runner]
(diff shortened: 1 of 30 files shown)


din-mukhammed:
@dabasov @peterlimg no hash root mismatches were identified during bench tests. Considering that default `benchTime` is set to 1 second and each run takes around 1ms, we should get approximately 500-1000 iterations per test, which is quite enough to catch any hash root mismatches, right?

dabasov (approved):
lgtm

PR #2235, opened 2023-03-22
Extend validators filtering for challenge creation

### LINKED ISSUE #2216: Challenge could be generated with no validators
We do not validate the selected validator num when create the challenge, therefore if there's no healthy validator, we will create challenge with no validators. 
https://github.com/0chain/0chain/blob/6cede0dfabd0d692958513024a77ede5a32628fe/code/go/0chain.net/smartcontract/storagesc/challenge.go#L965-L1006

din-mukhammed:
## Fixes
- closes #2216 
## Tests

- blobber:
- gosdk:
- system_test:
- zboxcli:
- zwalletcli:
- Other: ...


peterlimg (changes requested):
We need to check if the `needValidaNum` is greater than `minValidators`  before iteratiing to get validators in line 991. Otherwise, we would just waste resources on loading validators and filter them. Also, we need to iterating all `randValidators` till the selected number is >= `minValidators`, or if the needValidNum is equal to minValidators, any validator got ignored during the iterationg will fail the final `if len(selectedValidators) >= minValidators` checking.

peterlimg:
The `neededValidNum` should be `>= minValidators`, otherwise if `len(randValidators) > minValidators`, then the `neededValidNum` should be `minValidators`. So persuade code could be:
```go
if len(shards) + 1 > minVaildators {
    minValidators = len(shards) + 1
}

if len(randValidators) < minValidators {
    return error
}

for _, v := range randValidators {
    // load and filter validators
    
}

if len(selectedValidators) < minValidators {
    return error
}
```

code/go/0chain.net/smartcontract/storagesc/challenge.go
```diff
@@ -972,57 +973,82 @@ func (sc *StorageSmartContract) populateGenerateChallenge(
...
+		perm                = r.Perm(len(randValidators))
+		remainingValidators = len(randValidators)
 	)
 
 	now := txn.CreationDate
 	filterValidator := filterHealthyValidators(now)
 
-	for i := 0; i < needValidNum; i++ {
+	for i := 0; i < len(randValidators) && len(selectedValidators) < minValidators; i++ {
+		if remainingValidators < minValidators {
```
dabasov: imagine we have 
```go
minValidators = 50
remainingValidators = 50 //partitionSize
```
we still can fill the list of validators, also we do not decrease the minValidators on every successful validator, which doesn;'t look correct, I'd suggest to remove this check

dabasov (changes requested):

code/go/0chain.net/smartcontract/storagesc/challenge.go
```diff
@@ -972,57 +973,82 @@ func (sc *StorageSmartContract) populateGenerateChallenge(
 		return nil, errors.New("invalid blobber for allocation")
 	}
 
+	if alloc.DataShards+1 > minValidators {
```
dabasov: looks like we should get fixed number of validators
```yaml
    # number of validators per challenge
    validators_per_challenge: 2
```

dabasov (dismissed):

code/go/0chain.net/smartcontract/storagesc/challenge.go
din-mukhammed: I think it's not a fixed number of validators, but rather a lower bound i.e. [minimum number of validators](https://github.com/0chain/0chain/blob/c51117b529af747139ed591dacc4d3f08bde1aa0/code/go/0chain.net/smartcontract/storagesc/challenge.go#L1107)

code/go/0chain.net/smartcontract/storagesc/challenge.go
din-mukhammed: the purpose of `if remainingValidators < minValidators {` check is to accomplish early fail to avoid wasting resources on iterating through the list in case that remaining number of validators to choose from is already not enough to fulfill the minValidators requirement. (as @peterlimg requested above)

dabasov: I don't understand the logic, how can you address the situation in my comment?

code/go/0chain.net/smartcontract/storagesc/challenge.go
dabasov: this doesn't make sense
```
minValidators = alloc.DataShards + 1
```
we only need to get the minimum validators for any challenge, it's due to @guruhubb

din-mukhammed: Got it, I will make the changes to only get fixed number of validators from validators list, as set by `validators _per_challenge`

code/go/0chain.net/smartcontract/storagesc/challenge.go
din-mukhammed: > I don't understand the logic, how can you address the situation in my comment?

_As per current implementation it would work this way:_  if the `minValidators = 50` and `remainingValidators = 50`, loop will iterate through all the validators in the list checking if they meet the criteria, and if any validator gets kicked (filtered out) it will stop challenge creation, because the remainingValidators will be decremented and become 49 which will fail `remainingValidators < minValidators` check on next iteration.  

I will change the code to only get certain number of validators as set by `validators_per_challenge` in config

peterlimg (approved):
LGTM

peterlimg:
@din-mukhammed please check with @Kishan-Dhakan about the system tests failures.

service-0chain:
| config  | value |
| tests skipped  | FALSE  |
| system-tests  | reward_test_timeout  |
| gosdk  | staging  |
| zboxcli  | staging  |
| zwalletcli  | staging  |
| 0chain  | 2216-challenge-could-be-generated-with-no-validators  |
| blobber  | staging  |
| authorizer  | staging  |
| 0box  | staging  |
| 0block  | staging  |
| 0dns  | staging  |
| explorer  | master  |
| 0proxy  | staging  |
| 0search  | staging  |

code/go/0chain.net/smartcontract/storagesc/challenge.go
```diff
@@ -972,57 +974,82 @@ func (sc *StorageSmartContract) populateGenerateChallenge(
...
+
+		// sp, err := sc.getStakePool(spenum.Validator, validator.ID, balances)
+		// if err != nil {
+		// 	return nil, fmt.Errorf("can't get validator %s stake pool: %v", randValidator.Id, err)
+		// }
+		// stake, err := sp.stake()
+		// if err != nil {
+		// 	return nil, err
+		// }
+		// if stake < minStake {
```
peterlimg: @din-mukhammed any reason for removing the min stake checking.

peterlimg: Oh, it was not needed originally, if so, please remove the commented code and remove the `minStake` param, thanks.

peterlimg (changes requested):

code/go/0chain.net/smartcontract/storagesc/challenge.go
din-mukhammed: @peterlimg I was just testing to see if this min stake check was the cause of failing system tests. The check was requested from this [issue](https://github.com/0chain/0chain/issues/587), can you please confirm if it's the right way to filter out the validators with stake lower than minimum during `populateGenerateChallenge`? Anyway I can create a separate PR for that issue and unblock this one.

peterlimg: @din-mukhammed yes, let's do it in another PR.

peterlimg (approved):
LGTM

code/go/0chain.net/smartcontract/storagesc/challenge.go
peterlimg: @din-mukhammed let's select fixed number (validators_per_challenge) validators.

peterlimg (approved):
lgtm

PR #2269, opened 2023-04-04
Feat/user snapshots

### LINKED ISSUE #2267: User Snapshots
- Simply, old user aggregates are deleted. This is good to keep sharders db not bloated with historical data.
- On creating a new aggregate for the same, we depend on the latest value that's in the user_aggregates table. Also we create new aggregates ONLY based on change events, not in a periodic manner.
- When a user keeps unchanged for a long time, it's data is simply deleted from the db, so no old record to update when the new change occurs, thus having the new aggregate record with 0's (defaults) everywhere except for the changed fields.
- The solution is: simply keep the latest record in a new table, `user_snapshot` and depend on that to calculate the new aggregates.
- Cost: One more update to this table on change.

devyetii:
## Changes
- Add user snapshots table
- Use it in calculating user aggregates
## Tests

- blobber:
- gosdk:
- system_test:
- zboxcli:
- zwalletcli:
- Other: ...

### DIFF
--- a/code/go/0chain.net/smartcontract/dbs/event/user_aggregate.go
+++ b/code/go/0chain.net/smartcontract/dbs/event/user_aggregate.go
@@ -229,6 +229,7 @@ func (edb *EventDb) updateUserAggregates(e *blockEvents) error {
 	}
 
 	for _, aggr := range updatedAggrs {
+		curAggr := aggr
 		snap, ok := snapsMap[aggr.UserID]
 		if !ok {
 			snapsMap[aggr.UserID] = &UserSnapshot{
@@ -242,7 +243,7 @@ func (edb *EventDb) updateUserAggregates(e *blockEvents) error {
 			}
 			continue
 		}
-		merge(snap, &aggr)
+		merge(snap, &curAggr)
 		snap.UpdatedAt = time.Now()
 	}
 
--- a/code/go/0chain.net/smartcontract/dbs/event/user_aggregate.go
+++ b/code/go/0chain.net/smartcontract/dbs/event/user_aggregate.go
@@ -245,7 +245,7 @@ func (edb *EventDb) updateUserAggregates(e *blockEvents) error {
 		snapsMap[aggr.UserID] = snap
 	}
 
-	var newAggregates map[string]*UserAggregate
+	newAggregates := make(map[string]*UserAggregate, len(snapsMap))
 	for _, snap := range snapsMap {
 		newAggregates[snap.UserID] = &UserAggregate{
 			Round:           snap.Round,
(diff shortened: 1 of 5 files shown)


code/go/0chain.net/smartcontract/dbs/event/user_aggregate.go
```diff
@@ -206,41 +206,77 @@ func (edb *EventDb) updateUserAggregates(e *blockEvents) error {
 		}
 	}
 
-	ids := make(map[string]interface{})
-	for i := range updatedAggrs {
-		updatedAggrs[i].Round = e.round
-		ids[updatedAggrs[i].UserID] = struct{}{}
+	idSet := make(map[string]bool)
```
dabasov: interface should be preferable https://dave.cheney.net/2014/03/25/the-empty-struct

code/go/0chain.net/smartcontract/dbs/event/user_aggregate.go
```diff
@@ -206,41 +206,77 @@ func (edb *EventDb) updateUserAggregates(e *blockEvents) error {
...
+				TotalStake:      aggr.TotalStake,
+				ReadPoolTotal:   aggr.ReadPoolTotal,
+				WritePoolTotal:  aggr.WritePoolTotal,
+			}
 			continue
 		}
-		merge(a, &u)
+		merge(&snap, &aggr)
+		snap.UpdatedAt = time.Now()
+		snapsMap[aggr.UserID] = snap
```
dabasov: this assignment is not needed

code/go/0chain.net/smartcontract/dbs/event/user_snapshot.go
```diff
@@ -0,0 +1,32 @@
...
+	TotalStake      int64  `json:"total_stake"`
+	ReadPoolTotal   int64  `json:"read_pool_total"`
+	WritePoolTotal  int64  `json:"write_pool_total"`
+	PayedFees       int64  `json:"payed_fees"`
+	CreatedAt       time.Time
+	UpdatedAt       time.Time
+}
+
+func (edb *EventDb) GetUserSnapshotsByIds(ids []string) (snapshots []UserSnapshot, err error) {
+	err = edb.Store.Get().Where("user_id in (?)", ids).Find(&snapshots).Error
```
dabasov: use temp table for it, it's preferable then `in ?` since every different set of parameters generate different query and different plan as well, it is all cached in db and wastes resources

dabasov (changes requested):

code/go/0chain.net/smartcontract/dbs/event/user_aggregate.go
devyetii: Very insightful 💡Will apply

code/go/0chain.net/smartcontract/dbs/event/user_aggregate.go
devyetii: Needed it since `snapsMap` is a map of values not ptrs. Changed it to map of ptrs.

dabasov (approved):
lgtm

Issue #1270, opened 2022-05-12
Check if there are tokens burned after each transaction


peterlimg:
We need to check if transactions have tokens burned unexpectedly due to any reason, need to check code to confirm how this constraint checking could be implemented and applied.

Related issue #1255

TODOs:
- [x] For `send` transaction, verify that the sum of the original balances of `from` and `to` is equal to the sum of their updated balances after executing the transaction, including the fee.
- [x] Do balances checking for [DistributeRewards](https://github.com/0chain/0chain/blob/90c1cfc57e7c6fcf95aec36cbaf7b6096f68118b/code/go/0chain.net/smartcontract/stakepool/stakepool.go#L506) method, the input `value` that is going to be distributed must be the same as the sum of all tokens that are added to the pools.
- [x] For `mint` SC, assert that client won't get tokens beyond `max_mint`. Also, check the balances. 
- [x] Check the `burn` SC does not burn tokens overflow.

The `mint` and `burn` are checked in `transfer` the same as how `send` is processed.

dabasov:
we can use benchmark tests to make sure that no tokens were really burnt, the idea is to track all the 
 mints of the run and assert that `sum_balances_before = sum_balances_after - mints`

PR #2284, opened 2023-04-10
Assert client balances after transfer and mint

peterlimg:
## Fixes
- Partial fixes for #1270

## Changes
- Added state context cach for clients, so when client state is loaded/inserted to MPT, the client state will be set to the cahce. Adding this is to avoid duplicate MPT reading when checking the clients balances.
- The zcnsc.Burn transaction would be processed in the `transfer` case

## Tests

- blobber:
- gosdk:
- system_test:
- zboxcli:
- zwalletcli:
- Other: ...

### DIFF
--- a/code/go/0chain.net/chaincore/chain/state.go
+++ b/code/go/0chain.net/chaincore/chain/state.go
@@ -418,7 +418,7 @@ func (c *Chain) updateState(ctx context.Context, b *block.Block, bState util.Mer
 
 	ue := make(map[string]*event.User)
 	for _, transfer := range sctx.GetTransfers() {
-		tEvents, err := c.transferAmount(sctx, transfer.ClientID, transfer.ToClientID, transfer.Amount)
+		tEvents, err := c.transferAmountWithAssert(sctx, transfer.ClientID, transfer.ToClientID, transfer.Amount)
 		if err != nil {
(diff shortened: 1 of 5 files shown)


dabasov (commented):
do you think that checking for transfer and mint solely is enough? I thought we want to check transaction in common, I mean the series of transfers and mints

peterlimg:
> do you think that checking for transfer and mint solely is enough? I thought we want to check transaction in common, I mean the series of transfers and mints

Not sure what the series of transfers and mints mean. If you mean added up all the balances of before and after, and check if there are tokens other than minted amount added. Then we may need to consider the coin overflow when adding up the tokens (though most of the time we won't see this). Suppose we have 2 transfers: `transfer_A_to_B` and `transfer_A_to_C`, all from the same client `A`. Suppose A has `(2/3)*token_cap`, adding total balances before will have balance of `(4/3) * token_cap`, which is overflow the total tokens we support, and it is also possible to overflow the `uint64`. The same could happen as `mints`.

dabasov:
Ah, let me try explain better, every transaction generates a set of transfers and mints in general, changing a couple of balances each, I asked why dn't we assert states changed during the transaction before and after? we can do it in banchmark tests I think, similar to how we check MPT state consistency

dabasov (approved):

PR #2331, opened 2023-04-26
Fix txn saving panic

peterlimg:
## Fixes

- Sharders panic after several times of failure on saving trasactions to cassandra. It could happen when cassandra ran out of resources, or when the transactions data size was too big. This PR tries to fix the second case by reducing the txn saving batch size to 1, so txns will be saved one by one. If it still fails, then we need to do further investigation.

## Tests

- blobber:
- gosdk:
- system_test:
- zboxcli:
- zwalletcli:
- Other: ...


dabasov (approved):
lgtm

peterlimg:
Seems this PR does not fix the panic, we'll fix the panic via #2333 as a workaround, and do further investigation in a new PR.

PR #2353, opened 2023-04-29
Fix: nil pointer error during authorizers retrieval request

### LINKED ISSUE #2352: Fix: nil pointer error during authorizers retrieval request

YarikRevich:


peterlimg (approved):
LGTM

code/go/0chain.net/smartcontract/zcnsc/handler.go
```diff
@@ -72,11 +71,6 @@ func (zrh *ZcnRestHandler) getAuthorizerNodes(w http.ResponseWriter, r *http.Req
 		return
 	}
 
-	if len(authorizers) == 0 {
```
dabasov, line 75: why do you want to remove this check?

code/go/0chain.net/smartcontract/zcnsc/handler.go
```diff
@@ -57,10 +56,10 @@ func (zrh *ZcnRestHandler) getAuthorizerNodes(w http.ResponseWriter, r *http.Req
 		return
 	}
 
-	var (
-		authorizers []event.Authorizer
-		err         error
-	)
+	var err error
+
+	authorizers := make([]event.Authorizer, 0)
```
dabasov, line 61: how should help? you are guaranteed to receive error is no records are found

dabasov (changes requested):

code/go/0chain.net/smartcontract/zcnsc/handler.go
YarikRevich: It's more natural to receive empty array instead of error, if no authorizers are registered. The logic, which handles not enough number of authorizers has already been implemented in [gosdk](https://github.com/0chain/gosdk/blob/f0d08895c0f1a0abd22f884a8d2fa767c2acbf25/zcnbridge/authorizers_query.go#L53-L55)

code/go/0chain.net/smartcontract/zcnsc/handler.go
YarikRevich: Please, remember, that it's a public API. Approach of returning error if no records found is acceptable in ORMs, but not public APIs.

dabasov: I didn't get you, check the code, in line 66 and 64 you initialize this slice or process the error, you simply do nothing with this initialization.

dabasov (changes requested):

code/go/0chain.net/smartcontract/zcnsc/handler.go
YarikRevich: That initialization is needed, because of the logic of default JSON encoder. If your slice is empty and not initialized with "make" it will be interpreted as "null" in JSON.

peterlimg: @dabasov this will create an instance of authorizer slice that has empty items, so that http response will return 200 with empty list, while if we return nil, the common.Response will return `StatusNoContent`,  which I think the client ends does not work with it.

dabasov: which issue do we fix here? I thought nil reference, but there should be no nil reference

dabasov: actually nil reference was fixed by me, when we passed nil error in 76, so not sure what is the reason of this PR

dabasov (approved):
lgtm

PR #2425, opened 2023-05-19
fixed random blobber allocs out of range

dabasov:
## Tests

- blobber:
- gosdk:
- system_test:
- zboxcli:
- zwalletcli:
- Other: ...

### DIFF
--- a/code/go/0chain.net/smartcontract/storagesc/challenge.go
+++ b/code/go/0chain.net/smartcontract/storagesc/challenge.go
@@ -931,7 +931,7 @@ func (sc *StorageSmartContract) populateGenerateChallenge(
 
 	for i := 0; i < findValidAllocRetries; i++ {
 		// get a random allocation
-		allocID := randBlobberAllocs[randPerm[i]].ID
+		allocID := randBlobberAllocs[randPerm[i%blobberAllocPartitionLength]].ID
 
 		// get the storage allocation from MPT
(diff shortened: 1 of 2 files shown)


code/go/0chain.net/smartcontract/storagesc/challenge.go
```diff
@@ -931,7 +931,7 @@ func (sc *StorageSmartContract) populateGenerateChallenge(
 
 	for i := 0; i < findValidAllocRetries; i++ {
 		// get a random allocation
-		allocID := randBlobberAllocs[randPerm[i]].ID
+		allocID := randBlobberAllocs[randPerm[i%blobberAllocPartitionLength]].ID
```
peterlimg, line 934: It's also fixed in fix/stuck-debug branch, but in a different way. We'd better limit the findValidAllocRetries to be less or equal than the randBlobberAllocs, we don't need to do another round of finding, imagine the randBlobberAllocs is 3, the findValidaAllocRetries is 5, then we will do extra 2 unnecessary tries.

peterlimg (changes requested):

code/go/0chain.net/smartcontract/storagesc/challenge.go
dabasov: it's a hotfix, since miners are crashing, you can replace it with `fix/stuck-debug` when it will be merged

peterlimg (approved):
LGTM

peterlimg:
System tests failures not related.

PR #2444, opened 2023-05-24
Rollback event db on block finalize error or panic

### LINKED ISSUE #2435: Rollback event db changes if any error happens when finalizing a block
Event db changes could be saved to db successfully before the block is marked as finalized, error could happen before the block is finalized, such as transactions failed to be saved. The saved event db changes will cause duplication issue once the sharder get restarted and try to finalize the block again. See details in [discussion](https://0chain.slack.com/archives/C03788G8N2D/p1684680367680079) and related errors below:

```
panic: ERROR: duplicate key value violates unique constraint "idx_cchallenge_id" (SQLSTATE 23505)
helm-sharder-02-5567f84bcb-jvqc2
Defaulted container "helm-sharder-02" out of: helm-sharder-02, helm-promtail-02
panic: db store transaction failed. Error: failed to save transactions
Error: Process completed with exit code 1.
```

peterlimg:
## Fixes
- closes #2435

## Changes
- Post events commit till finalize block state persisted successfully. So any error or panic happens during the process, the events changes will all be dropped.
- Move the LFB round storing out of goroutine and do it as the last step in UpdateFinalizedBlock method. Because load LFB process on sharder starting starts from getting latest stored round from state db. So if any error or panic happens during the UpdateFinalizedBlock() method, we should not store the round to state db. 

- [x] Test to panic after saving state db successfully, LFB and events were saved, restarting will start from next round
- [x] Test to panic during the state db saving, LFB and events should all be ignored, and restarting should finalize it again

## Tests

- blobber:
- gosdk:
- system_test:
- zboxcli:
- zwalletcli:
- Other: ...

### DIFF
--- a/code/go/0chain.net/chaincore/block/block_eventdb.go
+++ b/code/go/0chain.net/chaincore/block/block_eventdb.go
@@ -47,8 +47,8 @@ func CreateBlockEvent(block *Block) (error, event.Event) {
 	}
 }
 
-func CreateFinalizeBlockEvent(block *Block) (error, event.Event) {
-	return nil, event.Event{
+func CreateFinalizeBlockEvent(block *Block) event.Event {
+	return event.Event{
 		BlockNumber: block.Round,
 		TxHash:      "",
 		Type:        event.TypeChain,
--- a/code/go/0chain.net/chaincore/chain/entity.go
+++ b/code/go/0chain.net/chaincore/chain/entity.go
@@ -89,7 +89,7 @@ type BlockStateHandler interface {
 	// nil if it don't have this ability.
 	SaveMagicBlock() MagicBlockSaveFunc
 	UpdatePendingBlock(ctx context.Context, b *block.Block, txns []datastore.Entity)
-	UpdateFinalizedBlock(ctx context.Context, b *block.Block) func() error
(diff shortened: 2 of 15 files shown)


dabasov (changes requested):
you can't simply return rollback function in defer, since if you commit inside process, rollback won't work. you can return full control over transaction outside and decide there what to do, depending on error in this method. this logic can be fuzzy though

```go
if err := tx.Commit(); err != nil {
		logging.Logger.Error("error committing block events",
			zap.Int64("block", es.round),
			zap.Error(err))
		return nil, err
	}
```

peterlimg:
> you can't simply return rollback function in defer, since if you commit inside process, rollback won't work. you can return full control over transaction outside and decide there what to do, depending on error in this method. this logic can be fuzzy though
> 
> ```go
> if err := tx.Commit(); err != nil {
> 		logging.Logger.Error("error committing block events",
> 			zap.Int64("block", es.round),
> 			zap.Error(err))
> 		return nil, err
> 	}
> ```

Ah, yes you're right, the rollback won't work if we committed, I'll check and fix

dabasov (approved):
lgtm

PR #2447, opened 2023-05-25
Fix/remove cassandra

### LINKED ISSUE #2409: Investigate cassandra failures causing sharder crash
Example run: https://github.com/0chain/system_test/actions/runs/4967778312/jobs/8889985209?pr=659
Related logs can be found in run logs or GitHub artifacts

devyetii:
## Fixes
Removing Cassandra everywhere from the sharder

## Changes
- Removed `persistencestore` class and all references to it.
- Removed `txn_confirmation` db as it's no more needed. Confirmation were never persisted, they were always generated on demand from txn_summary and block data.
- Changed txn_summary and magic_block_mab dbs to use rocksdb instead
- Added key spacing to txn_summary, it now stores both txn to round mapping in (TRANSACTION) keyspace and round to txn count mapping in (ROUND) keyspace.
- Refactored both of the fns (GetHighestMagicBlockMap, getTxnCountForRound) to use rocksdb
- Some unit tests
Yes

## Tests

- blobber:
- gosdk:
- system_test:
- zboxcli:
- zwalletcli:
- Other: ...


code/go/0chain.net/chaincore/block/magic_block_map_entity.go
```diff
@@ -113,8 +115,20 @@ func (mb *MagicBlockMap) Decode(input []byte) error {
 func SetupMagicBlockMapEntity(store datastore.Store) {
 	magicBlockMapEntityMetadata = datastore.MetadataProvider()
 	magicBlockMapEntityMetadata.Name = "magic_block_map"
+	magicBlockMapEntityMetadata.DB = "magicblockmapdb"
```
peterlimg, line 118: let's name it consistently, so `magic_block_map_db`.

code/go/0chain.net/chaincore/transaction/round_txns_count_entity.go
```diff
@@ -0,0 +1,56 @@
...
+	"0chain.net/core/datastore"
+	"0chain.net/core/encryption"
+)
+
+type RoundTxnsCount struct {
+	datastore.HashIDField // Keyspaced hash of the round number - used as key
+	TxnsCount  int   `json:"txns_count"`
+}
+
+const RoundKeySpace = "ROUND"
```
peterlimg: Let's keep it private unless it will be used by other packages.

Also please use `round_txn_count` instead of `ROUND` to avoid potential key conflicts, which can happen if we want to add another index for round in the future.

code/go/0chain.net/chaincore/transaction/round_txns_count_entity.go
```diff
@@ -0,0 +1,56 @@
...
+
+// SetRoundKey - set the entity hash to the keyspaced hash of the round number
+func BuildSummaryRoundKey(roundNumber int64) datastore.Key {
+	return datastore.ToKey(
+		fmt.Sprintf(
+			"%s:%s",
+			RoundKeySpace,
+			encryption.Hash(strconv.FormatInt(roundNumber, 10)),
```
peterlimg: Let's use use`Hash(fmt.Sprintf("%s:%d", RoundKeySpace, roundNumber))` instead of hash the round number.

code/go/0chain.net/chaincore/transaction/transaction_summary_entity.go
```diff
@@ -2,16 +2,21 @@ package transaction
...
 )
 
 /*TransactionSummary - the summary of the transaction */
 type TransactionSummary struct {
-	datastore.HashIDField
+	datastore.HashIDField // Keyspaced transaction hash - used as key
 	Round int64 `json:"round"`
 }
 
+const TransactionKeyspace = "TRANSACTION"
```
peterlimg: Keep it private the same as above, and use `transaction_round` to give more info about the key, also let's avoid using uppercase in key string.

code/go/0chain.net/chaincore/transaction/transaction_summary_entity.go
```diff
@@ -25,6 +30,17 @@ func (t *TransactionSummary) GetEntityMetadata() datastore.EntityMetadata {
...
 
+// SetTransactionKey - set the entity hash to the keyspaced hash of the transaction hash
+func BuildSummaryTransactionKey(hash string) datastore.Key {
+	return datastore.ToKey(
+		fmt.Sprintf(
+			"%s:%s",
+			TransactionKeyspace,
+			hash,
+		),
+	)
```
peterlimg, line 44: Same as above, use `Hash(content)` as key to have the consistent key length.

code/go/0chain.net/chaincore/transaction/transaction_summary_entity.go
```diff
@@ -59,8 +75,23 @@ func (t *TransactionSummary) Delete(ctx context.Context) error {
 func SetupTxnSummaryEntity(store datastore.Store) {
 	transactionSummaryEntityMetadata = datastore.MetadataProvider()
 	transactionSummaryEntityMetadata.Name = "txn_summary"
+	transactionSummaryEntityMetadata.DB = "txnsummarydb"
```
peterlimg, line 81: `txn_summary_db`

code/go/0chain.net/sharder/block.go
```diff
@@ -165,33 +168,35 @@ func (sc *Chain) GetMagicBlockMap(ctx context.Context, magicBlockNumber string)
...
-	var mctx = persistencestore.WithEntityConnection(ctx, mbmemd)
-	defer persistencestore.Close(mctx)
+	var mctx = ememorystore.WithEntityConnection(ctx, mbmemd)
+	defer ememorystore.Close(mctx)
 
-	const query = `SELECT MAX(id) FROM zerochain.magic_block_map;`
+	con := ememorystore.GetEntityCon(mctx, mbmemd)
+	if con == nil {
+		return nil, common.NewErrorf("get_highest_mbm",
+			"getting rocksdb connection: %v", err)
```
peterlimg: what's the `err`?  we can say `could not get connection of magic block rocksdb`

code/go/0chain.net/sharder/transaction.go
```diff
@@ -159,45 +157,55 @@ func (sc *Chain) StoreTransactions(b *block.Block) error {
...
 
-var txnSummaryMV = false
-var roundToHashMVTable = "round_to_hash"
+	rtcKey := transaction.BuildSummaryRoundKey(roundNumber)
+	var rtc transaction.RoundTxnsCount
+	err := txnSummaryMetadata.GetStore().Read(tctx, rtcKey, &rtc)
+	if err != nil {
+		rtc.SetKey(rtcKey)
+	}
+	rtc.TxnsCount += len(sTxns)
```
peterlimg: We should use `Merge` operation to do `read->modify->write` instead of do `get` and `put`. Check https://github.com/facebook/rocksdb/wiki/Merge-Operator and perhaps do it later after all works well.

peterlimg (dismissed):

peterlimg:
I did a sample for how to use the gorocksdb Merge operation, feel free to check it https://gist.github.com/peterlimg/97aee8241aeae3bcc7ee00a74c2a5855

code/go/0chain.net/chaincore/block/magic_block_map_entity.go
devyetii: I've compared it with all the other entities that use rocksdb, like round_summary, which is named "roundsummarydb". I think we need to change them all if so.

code/go/0chain.net/chaincore/transaction/round_txns_count_entity.go
devyetii: So the key will be just the hash, yes ?

code/go/0chain.net/chaincore/transaction/transaction_summary_entity.go
devyetii: Same here.

code/go/0chain.net/sharder/transaction.go
devyetii: Yeah sure, merge operation is what should be used here. However, it's not supported by the wrapper we're using. I guess we need to change it if not any other issues will emerge

devyetii: Will double check it and if it's available I'll certainly use it.

code/go/0chain.net/chaincore/block/magic_block_map_entity.go
dabasov: leave them all consistent, not sure we want to rename everything now

code/go/0chain.net/chaincore/transaction/round_txns_count_entity.go
dabasov: the hash of concatenated string, only single hash, since hashing function is not cheap

devyetii:
@peterlimg I managed to get the example you sent working but there's an issue there, which is that I need to know if the key is found or not. Looks like the merger doesn't return an error for this nor initialize the key with 0 or null, which means I'll still need to access this value before merging it to see if it's in the store or not.

service-0chain:
| config  | value |
| tests skipped  | false  |
| system-tests  | master  |
| gosdk  | staging  |
| zboxcli  | staging  |
| zwalletcli  | staging  |
| 0chain  | fix/remove-cassandra  |
| blobber  | staging  |
| authorizer  | staging  |
| 0box  | staging  |
| 0dns  | staging  |

peterlimg:
@devyetii thanks for the checking, we can do merge first, and if it returns error of `Failed to get value: Corruption: Error: Could not perform merge.`, then we do `Put()`. In this way, we can avoid extra `get` action once it's initialized.

devyetii:
@peterlimg Thanks for the comment. I've found a way that can help us do this. We can use the merge operator to initialize with an initial value if the existingValue is not found i.e. `string(existingValue) == ""`. I tested it and looks fine.

dabasov:
green run https://github.com/0chain/0chain/actions/runs/5149814818

PR #2456, opened 2023-05-26
Future transaction limit

peterlimg:
## Fixes
- Limit the future transaction nonce to accept limited number of future transactions only
- Fix panic on fs_store when reading block with empty hash
## Changes
- Remove past transactions.
TODO:
~-  [ ] Should limit the same nonce transaction from the same client as well, user can attack by sending large number of future transactions with the valid nonce but all have the same value.~

## Tests

- blobber:
- gosdk:
- system_test:
- zboxcli:
- zwalletcli:
- Other: ...


code/go/0chain.net/chaincore/chain/handler.go
```diff
@@ -1405,6 +1396,13 @@ func PutTransaction(ctx context.Context, entity datastore.Entity) (interface{},
 		return nil, errors.New("invalid transaction nonce")
 	}
 
+	if nonce+10 < txn.Nonce {
```
dabasov: is it config.TxnFutureNonce?

code/go/0chain.net/chaincore/chain/handler.go
```diff
@@ -1415,14 +1413,14 @@ func PutTransaction(ctx context.Context, entity datastore.Entity) (interface{},
...
 				// to avoid txn rejected due to miner's own fault
-				txnRsp, err := transaction.PutTransaction(ctx, txn)
-				if err != nil {
-					logging.Logger.Error("failed to save transaction",
-						zap.Error(err),
-						zap.Any("txn", txn))
-					return nil, common.NewErrInternal("failed to save transaction")
-				}
-				return txnRsp, nil
+				//txnRsp, err := transaction.PutTransaction(ctx, txn)
```
dabasov: do we really need to comment this?

code/go/0chain.net/miner/protocol_block.go
```diff
@@ -732,27 +732,40 @@ func txnProcessorHandlerFunc(mc *Chain, b *block.Block) txnProcessorHandler {
...
+		nonce, err := mc.validateTransaction(b, bState, txn, waitC)
 		switch err {
 		case PastTransaction:
 			tii.pastTxns = append(tii.pastTxns, txn)
-			if debugTxn {
-				logging.Logger.Info("generate block (debug transaction) error, transaction hash old nonce",
-					zap.String("txn", txn.Hash), zap.Int32("idx", tii.idx),
-					zap.Any("now", common.Now()), zap.Int64("nonce", txn.Nonce))
-			}
+			//if debugTxn {
```
dabasov: let's uncomment

code/go/0chain.net/miner/protocol_block.go
```diff
@@ -776,12 +789,12 @@ func txnProcessorHandlerFunc(mc *Chain, b *block.Block) txnProcessorHandler {
 
 		events, err := mc.UpdateState(ctx, b, bState, txn, waitC)
 		if err != nil {
-			if debugTxn {
-				logging.Logger.Error("generate block (debug transaction) update state",
-					zap.String("txn", txn.Hash), zap.Int32("idx", tii.idx),
-					zap.String("txn_object", datastore.ToJSON(txn).String()),
-					zap.Error(err))
-			}
+			//if debugTxn {
```
dabasov: uncomment maybe

dabasov (changes requested):

code/go/0chain.net/miner/protocol_block.go
```diff
@@ -1022,6 +1067,54 @@ func (mc *Chain) generateBlock(ctx context.Context, b *block.Block,
...
+				}
+			}
+
+			logging.Logger.Debug("remove future txns",
+				zap.Int("count", len(deleteTxns)),
+				zap.Strings("txns", txnHashes),
+				zap.Int64("future transaction limit", futureNonceAllowed))
+		}
+
+		if len(iterInfo.pastTxns) > 0 {
```
dabasov, line 1099: we can't remove past transactions here. tx_pool is shared among all the blocks, not only the current one, in some situations this block state will be reverted, but transaction in the pool will be lost forever. 

we can clean up transaction based on finalized block state, but not speculative

dabasov (changes requested):

code/go/0chain.net/miner/protocol_block.go
peterlimg: This is where the block generator packing the transactions, if the generator see them as past transactions, no matter how many times it re-run the block generation process, they will always be treated as past transactions unless the transaction selection algorithm is different in different tries. Also, if the transactions with the same nonce get packed, then all remaining transactions will be considered as past transactions, they wouldn't have the chance to be removed until timeout, which then users can attack the pool by submitting large number of transactions with the same nonce. They will all stuck in the the pool till timeout. 

While comparing to open the door to the attack, I think delete those past transactions when detected may not a bad choice, as they will be removed anyway after timeout, the only drawback is we can't have them back when need to revert the block, but that's a rare case, and we can still pack the valid transactions if the block is not finalized. 

Re 'clean up transactions on finalized block state', 
yes, but we would only clean up transaction packed in the finalized block, we can't afford to go through the txn pool again here to get past transactions. 

@guruhubb your thought?

dabasov: @peterlimg well, it's doubtful, in general it should be cheap to run through the in-memory pool and clean up transactions in a worker, we do not need to clean up them just now and doing it in background can be good. just new worker that get's lfb and cleanups all the past transactions

dabasov (approved):
lgtm, please check my previous comment, it should be better than what we have now

PR #2460, opened 2023-05-29
Fully order all event database pagination queries

### LINKED ISSUE #2455: Event database queries using offset should have no order ambiguity
We need the order stable when we use offset to page through an event database table. For example, at the moment, we often only order by round; this allows for ambiguity for objects indexed to the same round.

We should check all event database queries using offset to ensure no ambiguity in the order in which data is returned. Adding extra order by columns as necessary.

Sriep:
## Fixes
When using pagination we need the order results are returned in to be fully defined. Otherwise, order could be different between different offsets leading to confused results.

The Pr adds order by clauses to pagination queries to fully define the order. In general, we try to use data defined by smart contracts so that they remain consistent between sharders. `RewardProvider` and `RewardDelegate` were problematic so we used `id` which is defined by the sharder.
## Tests

- blobber:
- gosdk:
- system_test: https://github.com/0chain/system_test/pull/660
- zboxcli:
- zwalletcli:
- Other: ...

### DIFF
--- a/code/go/0chain.net/smartcontract/dbs/event/allocation.go
+++ b/code/go/0chain.net/smartcontract/dbs/event/allocation.go
@@ -66,11 +66,19 @@ func (edb *EventDb) GetClientsAllocation(clientID string, limit common.Paginatio
 
 	err := edb.Store.Get().
 		Preload("Terms").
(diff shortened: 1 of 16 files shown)


dabasov (approved):
lgtm

Issue #2439, opened 2023-05-23
Negative values in write pool


Hitenjain14:
```
allocation_fetch_error: Error fetching the allocation.internal_error: can't get allocation: error retrieving allocation: 212ac818f47f8a095c637c10c9068729f785f384ed4b97da9ee2fa46bc6b0ddf, error: sql: Scan error on column index 32, name "write_pool": converting driver.Value type int64 ("-333644590") to a uint64: invalid syntax
```

Sriep:
This error comes from [GetAllocation](https://github.com/0chain/0chain/blob/staging/code/go/0chain.net/smartcontract/dbs/event/allocation.go#L54).

Hence 
```
error: sql: Scan error on column index 32, name "write_pool": converting driver.Value type int64 ("-333644590") to a uint64: invalid syntax
```
is an SQL error trying to a `select *` on the allocations table.

dabasov:
should be something related to uint64 -> int64 conversion

[Transaction value int64 overflow error (postgres bigint doesn't allow unsigned int)](https://github.com/0chain/0chain/issues/2431)

peterlimg:
closed via #2484

PR #2462, opened 2023-05-30
Max token supply and validate build-in txns amount

### LINKED ISSUE #2397: Should validate amount of synthetic transaction
We need to limit the number of mandatory transactions (payFees, generate_challenge, blobber_block_rewards) per block.
At the moment w check that those transactions are created for a particular round, we need to check also that there is not more than 1 transaction in the block. Most like we will need to add this logic to block verification

This way we can avoid spamming these tx_types and it should allow us to make these transactions exempt, since it can be costly for miners to pay fees for several transactions

### LINKED ISSUE #2431: Transaction value int64 overflow error (postgres bigint doesn't allow unsigned int)
Sharders are crashing with the error :

Defaulted container "helm-sharder-01" out of: helm-sharder-01, helm-promtail-01
panic: failed to encode args[59]: unable to encode 0xffffffffc50fbdbe into binary format for int8 (OID 20): 18446744072720727486 is greater than maximum value for int64

TestCreateAllocation-Create_allocation_for_locking_cost_less_than_minimum_cost_should_not_work is the cause of this error because it is trying to lock negative tokens to create allocation. But sharders should not panic in such scenarios and we should fix it.

https://0chain.slack.com/archives/C03788G8N2D/p1684651726750149

### LINKED ISSUE #2438: Differentiate error response in allocations whether allocation is expired or cancelled
When allocation operations fails because the allocation is expired or cancelled, in both cases we are getting `Can't do something on expired allocation`.

We need to differentiate and return cancelled in the places it is returning expired. Relevant system tests should also be fixed.

peterlimg:
## Fixes
- closes #2431
- closes #2397
- closes #2438

## Tests

- blobber:
- gosdk:
- system_test:
- zboxcli:
- zwalletcli:
- Other: ...

### DIFF
--- a/code/go/0chain.net/chaincore/chain/handler.go
+++ b/code/go/0chain.net/chaincore/chain/handler.go
@@ -1360,8 +1360,17 @@ func PutTransaction(ctx context.Context, entity datastore.Entity) (interface{}, 
 		return nil, fmt.Errorf("put_transaction: invalid request %T", entity)
 	}
 
-	if txn.CreationDate < common.Now()-common.Timestamp(transaction.TXN_TIME_TOLERANCE) {
-		return nil, fmt.Errorf("put_transaction: time out of sync with server time")
+	err := txn.Validate(ctx)
+	if err != nil {
+		logging.Logger.Error("put transaction error", zap.String("txn", txn.Hash), zap.Error(err))
+		return nil, err
+	}
+
+	if txn.Value > config.MaxTokenSupply {
+		logging.Logger.Error("put transaction error - value exceeds max token supply",
+			zap.Uint64("value", uint64(txn.Value)),
+			zap.Uint64("max_token_supply", config.MaxTokenSupply))
+		return nil, fmt.Errorf("transaction value exceeds max token supply")
 	}
 
 	sc := GetServerChain()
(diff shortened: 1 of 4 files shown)


dabasov:
we need to add hard limit in currency.Coin with 40m ZCN, MAX value then will be 40*10^16, which is close to max.int64, it will prevent any overflow conversions to max.UINT64 which happens a lot in the code

peterlimg:
I thought about to add it to currency package too, but consider it's a util package, I decided not to add it there, also we don't have token calculation overflow issue recently, If ever it happens, the token mint/burn unexpectedly checking should report.

The overflow we can see usually happens in wrong calculation led negative int64 value, which seems often happen in event db processing, otherwise block state computing process should detected the int64 overflow. So for now, other than add the max token supply checking in  `/tranaction/put` endpoint, we need to use `currency` package to do token calculation in event db to stop int64 overflow first, the current currency package should be enough to detect those negative value caused overflow, and we can add the max token overflow to currency package later when needed (I mean when it happens).

dabasov:
@peterlimg please check this
[Negative values in write pool](https://github.com/0chain/0chain/issues/2439)

I think the problem is the same, when uint64 is converted to int64 leading bit is 1 and value becomes negative, having a hard cap in a currency type which is ZCN type can be reasonable, 40m*10^10 is the real max_value for this type

dabasov (approved):
lgtm

PR #2469, opened 2023-05-31
Marked all remaining challenges as passed for expired allocations

### LINKED ISSUE #2430: Mark all expired challenges after allocation expiry as successful in the Stats
Currently in the stats, we are seeing all the challenges that are not processed because of the allocation expiry as open challenges. It will be unfair for blobber. 

All challenges that are not expired by time should be in the successful challenges.

Jayashsatolia403:
## Tests

- blobber:
- gosdk:
- system_test:
- zboxcli:
- zwalletcli:
- Other: ...


dabasov (changes requested):
you can't access any external resources from smart contracts, event_db is an external resource, only MPT is accessible

Jayashsatolia403:
Updated

code/go/0chain.net/smartcontract/storagesc/allocation.go
```diff
@@ -1717,6 +1717,12 @@ func (sc *StorageSmartContract) finishAllocation(
 
 	alloc.Finalized = true
 
+	challenges, err := sc.getAllocationChallenges(alloc.ID, balances)
+	if err != nil {
+		return err
+	}
+
+	emitUpdateChallengesForExpiredAllocations(alloc.ID, balances, challenges)
```
peterlimg: We need to update the open challenge stats in MPT too, such as the [Stats](https://github.com/0chain/0chain/blob/6c19a04429c3cdeb9dfb08a7c150df4fc06ef6c7/code/go/0chain.net/smartcontract/storagesc/models.go#L630) in Allocation, and the [BlobberRewardStat](https://github.com/0chain/0chain/blob/0b6013e510c26e04f39b80ce61cdaf0a60b4b6a1/code/go/0chain.net/smartcontract/storagesc/challenge.go#L664) in ongoing partitions if any, etc.

peterlimg (changes requested):

code/go/0chain.net/smartcontract/storagesc/allocation.go
Jayashsatolia403: I don't think we need to update it for the BlobberRewardStats. They are used for block rewards and these numbers will make an issue there. 

We only need this change in the eventsdb so that blobber ratings doesn't get down because of this.

Jayashsatolia403: In the StorageAllocationStats in MPT, I can't think of a problem but will it really make some useful difference?

peterlimg: The data in event db and MPT should not be different, event db will provide richer info but should be aligned with the MPT data. And it should be applied on all cases in processing event db.

Jayashsatolia403: @dabasov Should I update it according to MPT? 

it should be applied on all cases in processing event db. But why is it mandatory for expired allocations?

dabasov: it is the correct way to update stats, look like there is also an update event
```go
	emitUpdateChallenge(cab.challenge, true, balances, cab.alloc.Stats, cab.blobAlloc.Stats)
```

Jayashsatolia403: Updated. Please review again. Thanks!!

code/go/0chain.net/smartcontract/storagesc/allocation.go
```diff
@@ -1717,6 +1722,23 @@ func (sc *StorageSmartContract) finishAllocation(
 
 	alloc.Finalized = true
 
+	for _, challenge := range challenges.OpenChallenges {
+		ba, ok := alloc.BlobberAllocsMap[challenge.BlobberID]
+
+		if ok {
+			ba.Stats.FailedChallenges++
```
peterlimg, line 1729: Shouldn't we increase the SuccessfulChallenges? or am I missed something

PR #2484, opened 2023-06-04
Fix negative write  pool

peterlimg:
## Fixes
This issue should have been fixed via #2462 actually, it happens in similar way that tests submit lock transactions with negative transaction value, while before #2462, we don't validate the value and will put it to txn pool. In the txn processing code, we ignored the token converting error and emit event to update the allocation.write_pool, which will be fixed by this PR. Anyway, because the #2462, the transactions with invalid value will be ignored, that's why we don't see this error recently. 

## Tests

- blobber:
- gosdk:
- system_test:
- zboxcli:
- zwalletcli:
- Other: ...

### DIFF
--- a/code/go/0chain.net/smartcontract/storagesc/writepool.go
+++ b/code/go/0chain.net/smartcontract/storagesc/writepool.go
@@ -2,6 +2,7 @@ package storagesc
 
 import (
 	"encoding/json"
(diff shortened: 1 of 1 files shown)


dabasov (approved):
lgtm

PR #2501, opened 2023-06-08
Fix challenge numbers for expired allocations

Jayashsatolia403:
## Tests

- blobber:
- gosdk:
- system_test:
- zboxcli:
- zwalletcli:
- Other: ...


code/go/0chain.net/smartcontract/storagesc/allocation.go
```diff
@@ -1715,8 +1723,21 @@ func (sc *StorageSmartContract) finishAllocation(
 		Amount:       i,
 	})
 
-	alloc.Finalized = true
+	if openChallengesAvailable {
```
dabasov: you can iterate empty list the same way, do you really need `openChallengesAvailable`?

Jayashsatolia403: Let me remove it.

dabasov (approved):
lgtm

code/go/0chain.net/smartcontract/storagesc/challenge.go
```diff
@@ -97,7 +97,7 @@ func (sc *StorageSmartContract) blobberReward(alloc *StorageAllocation, latestCo
 			zap.Int64("allocation expiry", int64(alloc.Expiration)),
 			zap.Int64("challenge time", int64(challengeCompletedTime)))
 
-		return nil
+		return fmt.Errorf("blobber reward failed: allocation is finalized")
```
peterlimg: Same as I commented in another PR, see [here](https://github.com/0chain/0chain/pull/2511/files#r1225972543)

code/go/0chain.net/smartcontract/storagesc/allocation.go
```diff
@@ -1538,6 +1538,11 @@ func (sc *StorageSmartContract) finishAllocation(
 	before := make([]currency.Coin, len(sps))
 	deductionFromWritePool := currency.Coin(0)
 
+	challenges, err := sc.getAllocationChallenges(alloc.ID, balances)
+	if err != nil {
+		logging.Logger.Info("finishAllocation: getAllocationChallenges", zap.Error(err))
```
peterlimg: Please do not ignore error returned from MPT.

peterlimg (changes requested):

code/go/0chain.net/smartcontract/storagesc/allocation.go
Jayashsatolia403: In this case, we are getting value not present error for the sc.ID. I am ignoring it because it doesn't matter much if there are no open challenges. 

Not sure, how to deal with this. Can you please guide me on it?

peterlimg: Yes, it's okay to ignore the `value not present` error, but it can also be `node not found` error,  which is returned when the miner's state is not fully synced. It's a common error that happens a lot in miners and randomly (especially after restarting). If we see `node not found` error we should return immediately. So the code could be:

```go
if err != nil {
    return fmt.Errorf("could not get allocation challenges: %v", err)
  } 
  log the value not present error if wanted
}

```

code/go/0chain.net/smartcontract/storagesc/challenge.go
```diff
@@ -494,10 +494,6 @@ func (sc *StorageSmartContract) verifyChallenge(t *transaction.Transaction,
 			"can't get related allocation: %v", err)
 	}
 
-	if alloc.Finalized {
-		return "", common.NewError(errCode, "allocation is finalized")
```
peterlimg: So we'll still process the challenge even the alloc is finalized?

Jayashsatolia403: No, I added this change and it accidently removed when I was merging to staging.

Jayashsatolia403: Readded it

peterlimg (approved):
LGTM

PR #2511, opened 2023-06-10
WIP : Fix finalize allocation blobber rewards

Jayashsatolia403:
## Tests

- blobber:
- gosdk:
- system_test:
- zboxcli:
- zwalletcli:
- Other: ...


code/go/0chain.net/smartcontract/storagesc/challenge.go
```diff
@@ -97,7 +97,7 @@ func (sc *StorageSmartContract) blobberReward(alloc *StorageAllocation, latestCo
 			zap.Int64("allocation expiry", int64(alloc.Expiration)),
 			zap.Int64("challenge time", int64(challengeCompletedTime)))
 
-		return nil
+		return fmt.Errorf("blobber reward failed: allocation is finalized")
```
peterlimg, line 107: If we need to return error for finalized allocation when verifyChallenge, then we should check it early to avoid unnecessary MPT access. Actually we can do the `if alloc.Finalized` checking [here](https://github.com/0chain/0chain/blob/be6f23cfefedea1565885c6ca0633deec01a35ac/code/go/0chain.net/smartcontract/storagesc/challenge.go#L486), in that way, all the MPT access afterwards can be avoid.

code/go/0chain.net/smartcontract/storagesc/challenge.go
```diff
@@ -653,15 +681,22 @@ func (sc *StorageSmartContract) challengePassed(
 				DataRead:          dataRead,
 			})
 		if err != nil {
-			return "", common.NewError("verify_challenge",
-				"can't add to ongoing partition list "+err.Error())
+			logging.Logger.Info("verify_challenge can't add to ongoing partition list " + err.Error())
```
peterlimg, line 701: We shall never miss the error checking and return if any, as it will hide the `node not found` error and then cause state hash mismatch error.

peterlimg (changes requested):

PR #2513, opened 2023-06-11
fixed transaction management

dabasov:
## Tests

- blobber:
- gosdk:
- system_test:
- zboxcli:
- zwalletcli:
- Other: ...


code/go/0chain.net/chaincore/chain/protocol_block.go
```diff
@@ -490,6 +469,27 @@ func (c *Chain) finalizeBlock(ctx context.Context, fb *block.Block, bsh BlockSta
 		}
 	}
 
+	wg.Run("finalize block - delete dead blocks", fb.Round, func() error {
```
peterlimg, line 472: We'd better use `wg.Run` together with `wg.Wait()`, or we can simply run the function in a goroutine if we don't need to wait for it.

peterlimg (changes requested):

code/go/0chain.net/smartcontract/dbs/event/process.go
```diff
@@ -86,16 +95,24 @@ func (edb *EventDb) ProcessEvents(
 				zap.String("block", block),
 				zap.Int("block size", blockSize))
 		}
+
+		if !commit {
+			err := tx.Rollback()
+			if err != nil {
+				return nil, err
+			}
```
peterlimg, line 103: We need to return `nil, error` here. If `commit` is false, that means the event was failed to be processed, and we need to avoid this `tx` to be used later after rollback.

peterlimg (changes requested):

PR #2662, opened 2023-07-31
WorkAggregates: return err

Sriep:
## Fixes
Returns an error that was being lost. This will cause some extra panics when aggregates fail, and the error is now being [returned rather than forgotten.](https://github.com/0chain/0chain/blob/staging/code/go/0chain.net/smartcontract/dbs/event/process.go#L255). Hopefully this will highlight some aggregate errors that were previously being missed.
``` go
func (edb *EventDb) addEventsWorker(ctx context.Context) {
	var gs *Snapshot
	p := int64(-1)
	edb.managePartitions(0)

	for {
		es := <-edb.eventsChannel

		s, err := edb.work(ctx, gs, es, &p)
		if err != nil {
			if config.Development() { //panic in case of development
				log.Panic(err)
			}
		}
		if s != nil {
			gs = s
		}
	}
}
```
## Tests

- blobber:
- gosdk:
- system_test:
- zboxcli:
- zwalletcli:
- Other: ...

### DIFF
--- a/code/go/0chain.net/smartcontract/dbs/event/process.go
+++ b/code/go/0chain.net/smartcontract/dbs/event/process.go
@@ -375,12 +375,14 @@ func (edb *EventDb) WorkAggregates(
 				zap.Int("block size", blockEvents.blockSize),
 				zap.Error(err),
 			)
+			return nil, err
 		}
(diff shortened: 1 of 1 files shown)


dabasov (approved):
lgtm

Issue #2559, opened 2023-06-27
Add further Conductor tests for blobber/validator


lpoli:
We need to add few conductor tests for blobber and validators.

- [x]  Case One:
1. Generate only single challenge
2. Make minority validator adversarial
3. Challenge should still pass.

- [x] Case Two: 
1. Upload a file and commit it.
2. Blobber shall commit it to blockchain
3. Client deletes file but writemarker is not committed
4. Blockchain will challenge blobber.
5. Blobber should still pass the challenge.

- [x] Case Three: // Maybe we need to move this to system-test
1. Create allocation with 2-data, 2-parity
2. Upload file in all blobbers
3. Rename file in one blobber only
4. Updating file should rollback minority blobber i.e. rename and at the end should sync up with each other.

- [x] Case Four:
1. Create allocation with 2-data, 2-parity
2. Upload file in all blobbers
3. Rename file in 3 blobbers
4. Update file in 2 + excluded blobber in step 3. Assuming a blobber is down.
5. Should consensus meet?
    -  If in step 4 it was able to repair excluded blobber then it should work.
    -  Otherwise it won't. Client should make rollback request
	      -  Client makes rollback request to two blobbers
	      -  Only one blobber accepts and another blobber rejects as it is adversarial for now.
	      -  Allocation is not in sync among blobber.
	      -  Repair should fix allocation state as there are 3 blobber which contain same file content. In this step
		    all blobber will be honest.

devyetii:
Related discussion: https://0chain.slack.com/archives/C02AV6MKT36/p1690611888903859

devyetii:
Done, all merged and added a workflow here: https://github.com/0chain/0chain/pull/2800

PR #2675, opened 2023-08-04
Add Case 3 and case 4

lpoli:
## Fixes
This PR is in draft state. It needs to add last two cases of https://github.com/0chain/0chain/issues/2559.
Currently case 3 is add but is yet to be tested. Case 4 is similar to case-3.
## Tests

- blobber: https://github.com/0chain/blobber/pull/1198
- gosdk:
- system_test:
- zboxcli:
- zwalletcli:
- Other: ...


code/go/0chain.net/smartcontract/storagesc/challenge_integration_tests.go
```diff
@@ -0,0 +1,128 @@
...
+	m := map[string]interface{}{
+		"blobber_id": cab.blobAlloc.BlobberID,
+		"status":     0,
+	}
+
+	if err == nil {
+		m["status"] = 1
+	}
+
+	crpc.Client().SendChallengeStatus(m)
```
dabasov: it will be called on every block, not only on those that are in the chain

dabasov (changes requested):

code/go/0chain.net/smartcontract/storagesc/challenge_integration_tests.go
devyetii: resolved

PR #2777, opened 2023-09-16
Fix /openchallenges ordering

Jayashsatolia403:
## Tests

- blobber:
- gosdk:
- system_test:
- zboxcli:
- zwalletcli:
- Other: ...

### DIFF
--- a/code/go/0chain.net/smartcontract/dbs/event/challenge.go
+++ b/code/go/0chain.net/smartcontract/dbs/event/challenge.go
@@ -141,9 +141,6 @@ func (edb *EventDb) GetOpenChallengesForBlobber(blobberID string, from int64, li
 		Order(clause.OrderByColumn{
 			Column: clause.Column{Name: "round_created_at"},
 			Desc:   true,
-		}).
-		Order(clause.OrderByColumn{
-			Column: clause.Column{Name: "challenge_id"},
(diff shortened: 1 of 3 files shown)


code/go/0chain.net/smartcontract/dbs/event/challenge.go
```diff
@@ -141,9 +141,6 @@ func (edb *EventDb) GetOpenChallengesForBlobber(blobberID string, from int64, li
 		Order(clause.OrderByColumn{
 			Column: clause.Column{Name: "round_created_at"},
 			Desc:   true,
-		}).
```
dabasov, line 144: you cannot remove ordering by id, pagination will be inconsistent then

dabasov (changes requested):

code/go/0chain.net/smartcontract/dbs/event/challenge.go
Jayashsatolia403: Actually round_created_at is already unique so it is redundant now. We had this earlier when we were using created_at  which had multiple entries on same created_at  because it was in second and there are multiple challenges per second.
So I think it is safe to remove.

dabasov (approved):
lgtm

PR #2782, opened 2023-09-16
Fix: check min mint amount no to be lower than max fee

### LINKED ISSUE #2781: Fix: check min mint amount no to be lower than max fee

YarikRevich:


dabasov (approved):
lgtm

PR #2837, opened 2023-10-07
Fix MPT state growth

### LINKED ISSUE #2807: Investigate MPT state in rocksdb grows fast issue
MPT data size grows fast in rocksdb, reached to about 25GB in 20days. Need to investigate what are the core issues. 

The MPT state pruning seems is suppressed when sync blocks from remote for sharders. Only few logs of `Do prune client state worker` while it should be logged every 30 seconds.

The prune will work after the block syncing process finish, but since the prune rarely happened during the syncing process, if the sharder is a fresh one, which start the sync from 0, the MPT state db could get pretty big before it finish the sync and clean up all dead nodes.

peterlimg:
The MPT state size become stable 

## Fixes
- #2807
## Changes
- Fix missed dead nodes when prunning state
- Reduce the delete obstable files period from 6 hours to 10 minutes. 
- Added debug endpoint `_diagnostics/est_num_keys` to returns the estimated total key num of state column family, dead_nodes column
- Do client state prune more frequently, previously we do every 30 seconds, and now changed to 7 seconds. The longer the period, the more dead nodes would be in the state DB. And if the detete process was slow, we may see the nodes number in MPT state grows with the `/_diagnostics/est_num_keys` 

## Tests

- common: https://github.com/0chain/common/pull/17
- blobber:
- gosdk:
- system_test:
- zboxcli:
- zwalletcli:
- Other: ...


dabasov (approved):
lgtm

PR #2925, opened 2023-11-04
Feature/conductor rollback tokenomics

### LINKED ISSUE #901: Add conductor tests to cover rollback allocation tokens flow

Jayashsatolia403:
## Tests

- blobber:
- gosdk:
- system_test:
- zboxcli:
- zwalletcli:
- Other: ...


code/go/0chain.net/conductor/services/allocations.go
```diff
@@ -0,0 +1,80 @@
...
+}
+
+func NewAllocationService(baseUrl string) *AllocationService {
+	return &AllocationService{
+		allocation: &types.Allocation{},
+		baseUrl:    baseUrl,
+	}
+}
+
+func (s *AllocationService) CompareAllocationsValue() (bool, error) {
```
devyetii: We need to make this function more general, where it accepts a list of fields to compare to. You can cast the allocation to a map[string]interface{} which will help you access those fields by their string key

code/go/0chain.net/conductor/services/allocations.go
```diff
@@ -0,0 +1,80 @@
...
+	// Read allocationID from file
```
devyetii: Please move this to a separate private (unexported) function within AllocationService and use it here.

docker.local/config/conductor.blobber-5.yaml
```diff
@@ -6,11 +6,45 @@ enable:
 sets:
   - name: "Minority Cases"
     tests:
+      - "Rollback tokenomics"
       - "Allocation state should be same after rollback"
       - "Allocation state should be same after repair"
 
 # test cases
 tests:
+  - name: "Rollback tokenomics"
```
devyetii: The test look incomplete to me

devyetii (changes requested):

docker.local/config/conductor.blobber-5.yaml
devyetii: Where is the part where you sync the value before it changes? Also the rollback will not run since the number of blobbers caught the change is equal DataShards. My resource: https://app.gitbook.com/o/-MWVD-8BJlEUSju1kMMq/s/3ep8kikKEUh61iaksVb3/

Jayashsatolia403: I am maintaining list of allocations objects in CS store so fetching it directly from there.

devyetii (approved):
LGTM

PR #3081, opened 2024-01-04
Add claimable rewards field to user aggregates

devyetii:
## Tests

- blobber:
- gosdk:
- system_test:
- zboxcli:
- zwalletcli:
- Other: ...


code/go/0chain.net/smartcontract/dbs/event/user_aggregate.go
```diff
@@ -162,6 +164,51 @@ var handlers = map[EventTag]func(e Event) (updatedAggrs []UserAggregate){
...
+			logging.Logger.Error("user aggregate",
+				zap.Any("event", event.Data), zap.Error(ErrInvalidEventData))
+			return
+		}
+
+		if len(*spus) == 0 {
+			return nil
+		}
+
+		rewards, err := aggregateProviderRewards(*spus)
```
dabasov, line 179: why do we need it here, we already call this func in `func (edb *EventDb) addStat(event Event) (err error) {
`

devyetii: This functions takes the event data and builds a reward map of it, including the `delegateRewards` map which will be used in the following code to calculate claimable rewards delta for each user. It doesn't add anything to the db or do anything else, just transforming events data to a form easier to deal with. Calling it in addStat will not have any effect outside addStat, as well as calling it here.

dabasov (approved):
lgtm

PR #3106, opened 2024-01-18
Add total_allocations field to the snapshots table

devyetii:
## Tests

- blobber:
- gosdk:
- system_test:
- zboxcli:
- zwalletcli:
- Other: ...

### DIFF
--- a/code/go/0chain.net/smartcontract/dbs/goose/migrations/1705569023_add_total_allocations_field_to_global_snapshot.sql
+++ b/code/go/0chain.net/smartcontract/dbs/goose/migrations/1705569023_add_total_allocations_field_to_global_snapshot.sql
@@ -1,6 +1,9 @@
 -- +goose Up
 -- +goose StatementBegin
 ALTER TABLE snapshots ADD COLUMN total_allocations BIGINT NOT NULL DEFAULT 0;
+
+UPDATE snapshots SET total_allocations = (SELECT COUNT(*) FROM allocations);
+
 -- +goose StatementEnd
 
 -- +goose Down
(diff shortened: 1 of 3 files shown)


code/go/0chain.net/smartcontract/dbs/goose/migrations/1705569023_add_total_allocations_field_to_global_snapshot.sql
```diff
@@ -0,0 +1,12 @@
+-- +goose Up
+-- +goose StatementBegin
+ALTER TABLE snapshots ADD COLUMN total_allocations BIGINT NOT NULL DEFAULT 0;
+
+UPDATE snapshots SET total_allocations = (SELECT COUNT(*) FROM allocations);
```
dabasov: update only latest snapshot, updating all snapshots can take a lot of time

dabasov (changes requested):
lgtm

code/go/0chain.net/smartcontract/dbs/goose/migrations/1705569023_add_total_allocations_field_to_global_snapshot.sql
devyetii: Ok sure. I was trying a different query to set the total_allocations of each snapshot entry to the number of allocations in the allocations table whose time is less than created_at of the snapshot record. To optimize, I was using the rank() function of the allocation entry - allocations being ordered by created_at - instead of counting,  but was trying to optimize the comparison step, which finds the nearest allocation record less than the snapshot record, in terms of creation date. This problem is similar to the graph queries in 0box, will take no less than O(nlogn) to generate total_allocations field for all n snapshot records.

dabasov (approved):
lgtm

PR #3156, opened 2024-02-20
Initial Kafka setup

Roshan-Mehta:
## Tests

- blobber:
- gosdk:
- system_test:
- zboxcli:
- zwalletcli:
- Other: ...


code/go/0chain.net/chaincore/chain/worker.go
```diff
@@ -267,6 +267,10 @@ func (c *Chain) FinalizedBlockWorker(ctx context.Context, bsh BlockStateHandler)
 						zap.Duration("duration", time.Since(ts)))
 				}()
 
+				if c.GetEventDb() != nil && c.GetEventDb().Store != nil {
+					go c.GetEventDb().PublishUnpublishedEvents()
+				}
```
peterlimg: The block is not finalized yet here, btw, why do we need to publish the unpublished events here, isn't all events are pushed to kafa when process the events when finalizing the block? i.e the `addEvents()` below. 
Also I see they are pushed to the same topic but the data are different, the `PublishUnpublisedEvents()` push the json of `event`, while the `addEvents` publish the `event.data` together with the round number. I guess this may cause issues when consumer use the data.

devyetii: > The block is not finalized yet here

Can you please mention where the block is totally finalized. We can add like a worker that publishes the unpublished events.

> why do we need to publish the unpublished events here, isn't all events are pushed to kafa when process the events when finalizing the block? i.e the addEvents() below.

No, events are published only by calling `PublishUnpublishedEvents`, `addEvents` should just adds them to the DB. The DB here acts as a buffer for the events. This approach has 3 advantages:
- Decouple events publishing from event generation. Event publishing shouldn't make event generation fail if it fails.
- Events publishing failover. This provides retry for the events that wasn't pushed successfully. Thus, no assumption of a 24/7 available message broker.
- Faster block finalization at the sharder, since event storing (i.e. `addEvents`) is part of the block finalization, so storing the events to the db is definitely faster than pushing them to Kafka directly.

> Also I see they are pushed to the same topic but the data are different, the PublishUnpublisedEvents() push the json of event, while the addEvents publish the event.data together with the round number. I guess this may cause issues when consumer use the data.

They're the same. `addEvents` adds the event to the DB and `PublishUnpublishedEvents` loads it from there.

After reviewing the latest code changes, I'll add some context here. `PublishUnpublishedEvents` was the only point of contact with Kafka, @Roshan-Mehta then added the kafka publishing logic to the `addEvents`.
What I suggest now is to remove the publishing logic from addEvents and move `PublishUnpublishedEvents` to a worker on the sharder for complete separation and add limit on the events that we can store in the DB, with proper alerting if the events failing to be published exceeded a specific threshold.
This is just a suggestion, The final decision if for you guys at the end of the day, I just presented the whole context as it was when I was working on it.

code/go/0chain.net/smartcontract/dbs/queueProvider/kafka.go
```diff
@@ -0,0 +1,127 @@
...
+		}
+	}
+	err := writer.WriteMessages(toutCtx,
+		kafka.Message{
+			Value: message,
+		},
+	)
+	if err != nil {
+		logging.Logger.Error("Publish: failed to write message on kafka", zap.String("topic", topic), zap.Any("message", message), zap.Error(err))
+		err := k.ReconnectWriter(topic)
```
goanchit, line 109: @dabasov FYI

dabasov (approved):
lgtm

PR #3202, opened 2024-03-19
Hotfix challenges

### LINKED ISSUE #3212: Add extensive coverage for testing all partitions related unit tests

Jayashsatolia403:
## Tests

- blobber:
- gosdk:
- system_test:
- zboxcli:
- zwalletcli:
- Other: ...

### DIFF
--- a/code/go/0chain.net/smartcontract/partitions/partitions.go
+++ b/code/go/0chain.net/smartcontract/partitions/partitions.go
@@ -521,6 +521,8 @@ func (p *Partitions) GetRandomItems(balances state.StateContextI, r *rand.Rand, 
 
 	var (
 		index int
+		its   []item
+		its2  []item
 	)
 
 	if p.Last.Loc > 0 {
@@ -535,11 +537,7 @@ func (p *Partitions) GetRandomItems(balances state.StateContextI, r *rand.Rand, 
 					return err
 				}
 
-				its, err := part.itemRange(0, part.length())
-				if err != nil {
-					return err
-				}
-				err = setPartitionItems(its, vs)
+				its2, err = part.itemRange(0, part.length())
 				if err != nil {
 					return err
 				}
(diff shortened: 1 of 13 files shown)


code/go/0chain.net/smartcontract/partitions/partitions.go
```diff
@@ -514,26 +514,65 @@ func (p *Partitions) removeItem(
 	return p.loadLastFromPrev(state)
 }
 
-func (p *Partitions) GetRandomItems(state state.StateContextI, r *rand.Rand, vs interface{}) error {
+func (p *Partitions) GetRandomItems(balances state.StateContextI, r *rand.Rand, vs interface{}, prev, partitionSize int) (int, error) {
```
peterlimg: We don't need to pass the `partitionSize`, the `Partitions` has `PartitionSize` field.

code/go/0chain.net/smartcontract/partitions/partitions.go
```diff
@@ -514,26 +514,65 @@ func (p *Partitions) removeItem(
...
+			part, err = p.getPartition(balances, index)
+			if err != nil {
+				return err
+			}
+			return nil
+		}, func() error {
+			for {
+				index = r.Intn(p.Last.Loc + 1)
+
+				if index != prev {
```
peterlimg: The new added `prev` params looks a bit weird for me.
We don't have to get rand index again and check if it's the index of previous one. Usually only the last partition could have number less than half of the partition size. So what we can do are:
1. check if it's the last partition, and if it is, then we check if it's less than half of the Size, if yes, then we get its previous partition if exists and get get the last `N` from the prev partition, `N = parititionsSize - items_num_in_last_part`. Then return the `append(prev_part_item, last_item)`. 
2. If the last partition is the only partition, then we just return all the items we have
3. If the selected part is not the last part, return the items in the selected part. 

With the changes above, we don't need to change the signature of `GetRandomItems()`, and the logic would be simpler.

dabasov (approved):
lgtm

PR #3308, opened 2024-06-13
Fix/validator url and partitions

Jayashsatolia403:
## Tests

- blobber:
- gosdk:
- system_test:
- zboxcli:
- zwalletcli:
- Other: ...

### DIFF
--- a/code/go/0chain.net/smartcontract/storagesc/challenge.go
+++ b/code/go/0chain.net/smartcontract/storagesc/challenge.go
@@ -1104,6 +1104,7 @@ func (sc *StorageSmartContract) populateGenerateChallenge(
 					return common.NewError("add_challenge",
 						err.Error())
 				}
+				return nil
 			}, func() error {
 				if cstate.ErrValueNotPresent(err) {
 					err = validators.Remove(balances, validator.Id())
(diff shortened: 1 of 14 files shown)


code/go/0chain.net/smartcontract/storagesc/challenge.go
```diff
@@ -1103,6 +1103,21 @@ func (sc *StorageSmartContract) populateGenerateChallenge(
 				return nil, common.NewError("add_challenge",
 					err.Error())
 			}
+
+			actErr = cstate.WithActivation(balances, "demeter", func() error {
+				return nil
+			}, func() error {
+				err = validators.Remove(balances, validator.Id())
```
dabasov: not sure we need to delete here

Jayashsatolia403: If validator is not in MPT (which can be the casse when validator is killed), we should remove it from the partition as well right? 

I added this because in kill validator we were not removing the validator from partition. I have fixed it now.

dabasov: can you check that the error is "NotPresent" error then?

Jayashsatolia403: Fixed here : https://github.com/0chain/0chain/pull/3308/commits/fba7b9d8e9cc3042201a239fa642f9b22399b5b8.

dabasov (approved):
lgtm

PR #3324, opened 2024-07-10
Demo network stuck fixes and experiments

peterlimg:
## Fixes
- Fixed the frequently round timeout and restarts caused slow network
## Changes
There are many reasons caused the slow network:
1. Miners on demo3 could not compute any real time blocks, most of them failed due to `node not found` error and the missing node syncing could not done in time.
2. High CPU usage caused by the full MPT state syncing, mainly because we need to iterate through the whole db to find all missing nodes, which could take hours to finish on demo networks.
3. High CPU usage caused by sharder on the demo3. 

This PR fixed the first 2 issues by 
a) Do strict block finalization like sharder: finalize block one after another, do not allow gaps.
b) Start miner from LFB load from local state instead fetching the latest LFB from remote and start from it. 
c) Add blocks sync worker like sharder to sync up blocks from LFB when detected stuck. 

After applying the fixes above, the first 2 issues should be fixed, the network now can move smoother than before, we can see much less `Round timeout cound`. 

However, due to the sharder on demo3 server is using too much CPU resources, also it is struggling on synching up, the 2 miners on the server sometime could not process blocks in time and could be behind the latest round of the network, but they will sync up soon and participant the real time block executing. 

Another benefit is that we should rarely see `node not found` error since this PR, because the miners will also have full MPT state.

## Tests

- blobber:
- gosdk:
- system_test:
- zboxcli:
- zwalletcli:
- Other: ...


code/go/0chain.net/miner/chain.go
```diff
@@ -248,6 +250,52 @@ func (mc *Chain) SetLatestFinalizedBlock(ctx context.Context, b *block.Block) {
...
+	if err != nil {
+		return fmt.Errorf("load_lfb - could not load lfb from state DB, err: %v", err)
+	}
+
+	logging.Logger.Debug("load_lfb - load from stateDB",
+		zap.Int64("round", lfbr.Round),
+		zap.String("block", lfbr.Hash))
+
+	// fetch from sharders
+	b, err := mc.GetNotarizedBlockFromSharders(ctx, lfbr.Hash, lfbr.Round)
```
dabasov, line 267: do we always want to load this block? it can be stored locally

peterlimg: Miners do not store blocks, and this `LoadLatestBlocksFromStore` will only be called once when start the miner, so should be fine.

dabasov (approved):
lgtm

service-0chain:
| config  | value |
| tests skipped  | false  |
| system-tests  |   |
| gosdk  | NOT OVERRIDDEN  |
| zboxcli  |   |
| zwalletcli  |   |
| 0chain  | staging  |
| blobber  |   |
| authorizer  |   |
| 0box  |   |
| 0dns  |   |

PR #3329, opened 2024-07-17
Fix rocksdb disk full

peterlimg:
## Fixes
- Fix rocksdb disk full issue. The more we sync state node from remote, the more nodes will be left in the rocksdb.  The process to clean up rocksdb dead nodes is like: mark deleted nodes that are collected when computing block state as dead nodes when finalize the block. We have another dead nodes clean worker that will remove those nodes from rocksdb peridocially. The problem is if we don't run the ComputeState for the block(where the deleted nodes are collected), instead we sync state changes from remote, which only sync new nodes. Then, when we finalize this block, we will see no delete nodes, therefore they will be left in the rocksdb forever. 

## Changes
Instead of synching the new added state nodes only, we sync the deleted nodes too, and when we apply the synced changes, we will add those deleted nodes to the client state. When finalize the block, those deleted nodes will be marked as dead nodes, then be clean up by the dead node worker. 

## Tests

- Common: - https://github.com/0chain/common/pull/24


dabasov (approved):
lgtm

PR #3370, opened 2024-08-16
Fix demo sharder stuck

peterlimg:
## Fixes
- Fix sharder stuck when current round only has a block that is not gona to be finalized, means the next block that the block worker try to process can't connect to the current round block. Finalization will stop due to the missed block in current round(the one that all others agreed and will finalize)
## Changes
- Fetch the missed block when detect it
- Set the processed block round as current round only when it's one round ahead of the previous one. This is to avoid skipping round setting, which could lead to potential missing blocks.

## Tests

- blobber:
- gosdk:
- system_test:
- zboxcli:
- zwalletcli:
- Other: ...


dabasov (approved):
lgtm

PR #3373, opened 2024-08-25
Fix miner stuck 

peterlimg:
## Fixes
- There's a bug in local bocks storage. When add a block that has already in the local store, we will return the exist one and apply changes on it, but due to the bug, we are applying changes on the new block instance, so even we computed the block state successfully, the state in the local store would not be updated, therefore, when another thread fetchs the block, it will see its state as not computed yet, and the client state will be nil.
## Tests

- blobber:
- gosdk:
- system_test:
- zboxcli:
- zwalletcli:
- Other: ...


dabasov (approved):
lgtm

